PR #138, opened 2013-10-13
Add MySQL enum support

aarbrown:
This PR allows a user to create enum columns using MySQL

Usage is as follows:

```
// Normal usage - pass a PHP array in for the 'limit' option
$this->addColumn('column_name', 'enum', array('limit' => array('value1', 'value2'))

// or create a blank enum (not sure why you would do this, but it is supported since MySQL supports it)
$this->addColumn('column_name', 'enum', array('limit' => array())

// creating an enum column and not specifying a limit will throw a PDOException
$this->addColumn('column_name', 'enum')
```

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -659,7 +659,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
                 return array('name' => 'tinyint', 'limit' => 1);
                 break;
             case 'enum':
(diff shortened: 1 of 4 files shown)


shadowhand:
This PR is broken in several ways:

1. it adds `enum` as an accepted column type, but does not declare a matching `PHINX_TYPE_ENUM` for it
2. it abuses `limit` to inject the enum constraints
3. it has many unrelated whitespace changes
4. it is out of date

As such, I am closing this. Please create a new PR if the functionality is still desired.

PR #239, opened 2014-04-29
Feature/config schema postgres

rsdevigo:
This makes it possible to define a schema for the log table is created.


shadowhand:
Looks like it was merged, but the PR never closed. Thanks!

PR #256, opened 2014-05-25
Adds an optional bind param Adapter::execute()

cyrusboadway:
This is a _proposal_, with none of the necessary additional test coverage (though the implementation does pass current tests).

The migration `::execute()` method does a simple `PDO::exec()` call for the implemented adapters. Unfortunately, that means the query's parameter escaping has to be done manually (`sprinf` might build a query, but that doesn't give some people a satisfactory way to write a custom migration).

Now, the following is valid:

```php
class SomeMigration extends AbstractMigration{
    public function up(){
        $params = array(         // array keys are the bind names
            'strColVal' => 'foo',    // default as string
            'intColVal' => array(    // but you can specify a type if you pass an array with 'value' and optionally 'type'
                'value' => 6,
                'type'  => static::PHINX_PARAM_INT,
            ),
        );
        $sql = 'INSERT INTO test(intCol, strCol) VALUE(:intColVal, :strColVal)';
        $this->execute($sql, $params); 
    }
}
```

__Change list:__

- Changes `::execute()` return value to `bool` from `int`. It previously returned the # rows affected, but was not internally used anywhere (output of call never assigned). PDO prepared statement returns boolean for success, so it was a convenience change.
- Adds a `array $params = null` property to the `MigrationInterface` and `AdapterInterface`.
- Adds a method to convert generalized bind PARAM types (defined in constants) to adapter specific PARAM types (i.e. to `\PDO::PARAM_*`).
- Updates adapters: PDO adapter in particular to run `PDO::exec()` if no properties are assigned (previous behaviour, and therefore supports complete backwards compatibility), and runs the query as prepared statements with the `binds` are defined.


shadowhand:
@cyrusboadway any progress on updating this?

cyrusboadway:
@shadowhand I'll take a swing at it over the weekend. Should have something late Sunday.

robmorgan:
I'm closing this in favour of #486.

PR #259, opened 2014-06-06
A new command LOG to display last SQL queries

solo-framework:
I suggest a new command to display SQL queries which occured since the last migration. It can be usefull to track changes in database during development.

**$phinx log** - shows all SQL commands from the last migration moment for now
**$phinx log --u** shows unique SQL queries only
**$phinx log --d** Clear MySQL binary log

Workflow is:
 * Clear binary log if you need with **$phinx log --d** (I usually clean this one because it more convenient for me otherwise I have a deal with a large data)
 * Do same changes in database: alter tables, modify data, etc.
 * At the moment when you have decided to create a new migration, you can get all your SQL queries in order to easier writing the migration. Just run **$phinx log** and see the result.

Ensure you MySQL user has proper permission to RESET MASTER command.
To get details see http://dev.mysql.com/doc/refman/5.0/en/binary-log.html


robmorgan:
I'm closing this until I see more interest from the community.

PR #268, opened 2014-07-10
MySQL createDatabase with charset/collation from config

speroshiropetto:
When first using `createDatabase()`, I had `charset: utf8mb4` in `phinx.yml` and made the assumption `charset` was passed along beyond the connection dns #134. So when setting the collation `$this->createDatabase('newDb', array('collation' => 'utf8mb4_unicode_ci'));`, it naturally produced an syntax error without being accompanied by `'charset' => 'utf8mb4'`.

    [PDOException]
    SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_unicode_ci'
    is not valid for CHARACTER SET 'utf8'

Changes that I would propose...

Under MySQL, `createDatabase()` would look to both the options array passed as well as the `phinx.yml` configuration file for `charset` & `collation`. If neither exist, it defaults to `utf8` and the default collation of the character set.

* Looks to `phinx.yml` configuration to set charset & collation if supplied
during `createDatabase()`.
* Passed attributes in array will override configuration parameters.
* Leaving both off will default
* Added `collation: utf8_general_ci` to default `phinx.yml` and Docs


robmorgan:
This can be closed as the MysqlAdapter now supports a `collation` option.

PR #275, opened 2014-07-24
Allow to use one manager to module based migrations

Komanc:
I have few modules. Each of them should have own migrations. When I create manager instance for each module it tooks too long time. I needed change schema table on the fly per each module.

```php
    public function install()
    {
        $config = $this->getPhinxConfig();
        $environment = $config->getEnvironment('module_migrate');
        $migrationTable = $environment['default_migration_table'];
        $this->manager->setConfig($config);
        $this->manager->switchSchemaTableName('module_migrate', $migrationTable);

        // migrate to the latest version
        $this->manager->migrate('module_migrate');
        $this->manager->reset();
    }
```


robmorgan:
closing in favour of #488.

PR #277, opened 2014-07-27
Change the way to get migration template on Create and Abstract commands

anyeloamt:
Currently I need to use a custom migration template for the create command. With this change I can override this method to return my custom template.


shadowhand:
Sorry it took so long to get this merged!

PR #300, opened 2014-09-04
Added support for char datatype

dingels35:


JCMais:
Just a heads up, the docs have no mention to the ``char`` type.

pburkind:
Still not mentioned in the docs

PR #317, opened 2014-09-23
Do not force sqlite database extension .sqlite3

amorce:
sqlite database file extension must be choose by the developper not by phinx


shadowhand:
Replaced by #392.

PR #330, opened 2014-10-11
Add support for TINY/MEDIUM/LONGTEXT columns in MySQL

opsydev:
The column type is now selected based on the limit specified. This is done the Rails way. The limit must be specified in bytes. The limits are selected based on documentation here
http://dev.mysql.com/doc/refman/5.0/en/string-type-overview.html

If this is okay I can add support for TINYINT as well.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -43,6 +43,11 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
 
     protected $signedColumnTypes = array('integer' => true, 'biginteger' => true, 'float' => true, 'decimal' => true);
 
+    const TEXT_SMALL = 255;
+    const TEXT_REGULAR = 65535;
+    const TEXT_MEDIUM = 16777215;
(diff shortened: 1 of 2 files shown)


adamski:
I'm getting ```An invalid column type "longtext" was specified for column``` on a MySQL table migration. I'm looking through the code and I see that ```getColumnTypes()``` is not implemented for the MySQL Adapter. Is there a reason for this? 
My code looks like:
```php
->addColumn('message', 'longtext', array('null'=>true))
```

twistedR:
@adamski To keep compatibility with other DB systems there is no column 'longtext' you specify the limit and for MySQL it becomes long text. So your code should be

```
->addColumn('message', 'longtext', array('null'=>true, 'limit' => MysqlAdapter::TEXT_LONG))

**Comment by adamski** [comment]
OK great, thank you

**Comment by adamski** [comment]
The code that worked was 
```php
->addColumn('message', 'text', array('limit' => MysqlAdapter::TEXT_LONG, 'null' => true))
```
and I had to add the namespace to the top of the file: 
```php
use Phinx\Db\Adapter\MysqlAdapter;
```

PR #334, opened 2014-10-16
support table comment

yoshikitanaka:


odan:
Hi!

I'm using `MySQL` and I tried this here without success. The `users` table comment value is still empty.

```php
$table = $this->table("users", ['comment' => "my table comment"]);
$table->save();
```

Does phinx support MySQL table comments?

lorenzo:
Yes, it does, but I think I’m on creation. Open a new ticket if it is not working for you and add details about what you are doing

robmorgan:
Hi @odan MySQL table comments are only supported by Phinx on table creation. Refer to the PR here: https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/MysqlAdapter.php#L259. However it looks like MySQL supports this after the table has been created, so its probably just a case of someone implementing it: https://dev.mysql.com/doc/refman/5.5/en/alter-table.html. I assume your table already exists?

odan:
@robmorgan 

> I assume your table already exists?

Yes, my table already exists and I tried to add/change a table comment.
As suggested by @lorenzo, I create a new ticket. Thank you.

robmorgan:
@odan even better... why don't you submit a pull request and try and contribute something?

odan:
@robmorgan Okay, I'll do that :-)

PR #349, opened 2014-11-10
Support PostgreSql Array type

tuannh99:
PostgreSql's Array type is there (http://www.postgresql.org/docs/9.3/static/arrays.html) since version **9.0**.
Now we support it in Phinx!


shadowhand:
Was manually merged into 0.4.x branch. Thanks!

PR #355, opened 2014-11-15
Make Util return timestamps in UTC, fixes #354

### LINKED ISSUE #354: Migration serial numbers should be in UTC
On my OS X box, I get this annoying warning:

```
Warning: date(): It is not safe to rely on the system's timezone settings. You are *required* to use the 
date.timezone setting or the date_default_timezone_set() function. In case you used any of those 
methods and you are still getting this warning, you most likely misspelled the timezone identifier. We 
selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in 
./vendor/robmorgan/phinx/src/Phinx/Migration/Util.php on line 45
```

It might sound irrelevant, but it highlights the fact that Phinx does not use UTC internally for its serials/migration timestamps. 

This can in fact cause a lot of problems for developers working on the same source, in different timezones: if Developer-A adds a migration before Developer-B, but B is in a much earlier TZ, then the later migration might be actually sorted lower than the earlier one.

shadowhand:
By using UTC timestamps, we can avoid a situation where a developer A in timezone 1
creates a migration A:1, and then developer B in an "earlier" timezone 2 creates a
migration B:2 that appears to be made before A:1.
- Creates a separate `Util::getCurrentTimestamp` method
- Adds a constant for `Util::DATE_FORMAT`
- `Util::mapClassNameToFileName` uses `Util::getCurrentTimestamp`

### DIFF
--- a/src/Phinx/Migration/Util.php
+++ b/src/Phinx/Migration/Util.php
@@ -45,6 +45,7 @@ class Util
         $dt = new \DateTime('now', new \DateTimeZone('UTC'));
         return $dt->format(static::DATE_FORMAT);
     }
+
     /**
      * Turn migration names like 'CreateUserTable' into file names like
      * '12345678901234_create_user_table.php' or 'LimitResourceNamesTo30Chars' into


JCMais:
@shadowhand Found a regression for that on https://github.com/robmorgan/phinx/blob/0.4.x-dev/src/Phinx/Migration/Manager/Environment.php#L125 -> ```date('Y-m-d H:i:s', time())```

Issue #306, opened 2014-09-10
Config flag for prepending timestamp to migration classes to allow duplicate names


anlutro:
In cases where an application goes through a lot of changes there can be a lot of back and forth in terms of adding/removing tables, columns, indexes, foreign keys etc. In such cases, not being able to have two migration files named the same can be annoying. I have been in positions where I've had migrations called

- create_something_table / class CreateSomethingTable
- drop_something_table / class DropSomethingTable
- create_something_table_again / class CreateSomethingTableAgain
- drop_something_table_again / class CreateSomethingTableAgain
- create_something_table_thirdtime / class CreateSomethingTableThirdTime

In my eyes, if you know you're going to be in such a position, having at least the possibility to prepend a timestamp to the class name itself in order to simplify the name of the migration itself is useful.

I'd suggest a root config key called `prepend_timestamp` or something, which is a boolean.

robmorgan:
I'm closing this for now until I see further support from the community.

PR #359, opened 2014-11-19
auto timestamp migration class names if auto_timestamp_class is set in configuration

zacek:
This is implementation of the issue #306. I'm missing this feature too, it is even a blocker for me.

I've created a new configuration option named "auto_timestamp_class". When this is set to "yes", "true" or 1, the class name will be suffixed by the timestamp, which avoids conflict in names. If the configuration value is missing or "no", "false" or "0" it works as before (thus it is backward compatible)

I've chosen postfix instead of prefix because PHP class names must start with a letter or underscore.

I've also included a small fix to upper case the first letter of the migration name automatically, i.e. it is now equivalent
phinx create Abc
and
phinx create abc

This pull request is replacing #328

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -251,7 +251,21 @@ class Config implements \ArrayAccess
 
         return $path;
     }
-    
+
+    /**
+     * Get the flag tp auto timestamp migration class names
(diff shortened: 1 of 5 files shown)


robmorgan:
I'm closing this for now until I see further support from the community.

PR #392, opened 2014-12-06
Make the sqlite3 extension optional. Closes #317

josegonzalez:
Refs cakephp/migrations#7

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -69,7 +69,10 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
             if (isset($options['memory'])) {
                 $dsn = 'sqlite::memory:';
             } else {
-                $dsn = 'sqlite:' . $options['name'] . '.sqlite3';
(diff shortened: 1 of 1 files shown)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -69,7 +69,10 @@ public function connect()
             if (isset($options['memory'])) {
                 $dsn = 'sqlite::memory:';
             } else {
-                $dsn = 'sqlite:' . $options['name'] . '.sqlite3';
+                $dsn = 'sqlite:' . $options['name'];
+                if (file_exists($options['name'] . '.sqlite3')) {
```
shadowhand: Technically, this should use `is_file`, because `file_exists` will return `true` for a directory as well as a file.

josegonzalez:
Any reason why you manually committed my change as opposed to merging the PR?

shadowhand:
It was targeting the old 0.3.x branch. I merged it, but it looks like it failed to preserve authorship. Sorry about that.

josegonzalez:
No worries.

PR #421, opened 2014-12-28
Call createSchemaTable on setConnection, fixes #420

### LINKED ISSUE #420: Phinx log table should be created for existing connections
#379 added support for providing an existing connection to Phinx, but did not update the code to automatically to call `createSchemaTable` when setting an existing connection.

See #409 as an example of this side effect.

shadowhand:
- modify `PostgresAdapter::createSchemaTable` to reuse `PdoAdapter::createSchemaTable`
- make `PdoAdapter::createSchemaTable` compatible with Postgres (remove unnecessary `limit` option)

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -82,11 +82,6 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
             }
 
             $this->setConnection($db);
-
-            // Create the schema table if it doesn't already exist
-            if (!$this->hasSchemaTable()) {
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -375,7 +381,7 @@ public function createSchemaTable()
             );
 
             $table = new Table($this->getSchemaTableName(), $options, $this);
-            $table->addColumn('version', 'biginteger', array('limit' => 14))
```
robmorgan: maybe there wasn't a default set in the beginning? or I added it unnecessarily.

PR #423, opened 2014-12-29
When using an alternative base class as defined by migration_base_class ...

RQuadling:
...in the config file that is of a different namespace to Phinx\Migration, the default template does not allow for substitution of the 'use' clause.

I'm making feature requests on master, but because that is on the same commit as 0.4.x-dev, I'm not sure if this will merge cleanly.


src/Phinx/Config/Config.php
```diff
@@ -239,15 +239,18 @@ public function getMigrationPath()
...
     {
         if (!isset($this->values['migration_base_class'])) {
-            return 'AbstractMigration';
+            $className = 'Phinx\Migration\AbstractMigration';
+        } else {
+            $className = $this->values['migration_base_class'];
         }
 
-        return $this->values['migration_base_class'];
+        return $dropNamespace ? basename(str_replace('\\', DIRECTORY_SEPARATOR, $className)) : $className;
```
shadowhand: While using `basename` here does work, I don't think it is entirely appropriate. I think using a non-filesystem approach would be better:

```php
if ($dropNamespace) {
    $className = substr(strrchr('\\', $className), 1);
}
```

src/Phinx/Console/Command/Create.php
```diff
@@ -50,6 +58,13 @@ protected function configure()
                 PHP_EOL,
                 PHP_EOL
             ));
+
+        // An alternative template.
+        $this->addOption('template', 't', InputOption::VALUE_REQUIRED, 'Use an alternative template');
+
+        // A classname to be used to gain access to the template content as well as the ability to
+        // have a callback once the migration file has been created.
+        $this->addOption('class', 'l', InputOption::VALUE_REQUIRED, 'Use a class implementing "' . self::CREATION_INTERFACE . '" to generate the template');
```
shadowhand: The addition of these two options should be done in a separate pull request. The ability to define a base class that exists outside of `Phinx` namespace is a separate issue from having a `CreationInterface`. Please make this PR deal with the namespace issue only. Thanks!

src/Phinx/Config/Config.php
RQuadling: I develop on Windows and realised that it was inconsistent for non Windows. Will use string manipulation.

src/Phinx/Console/Command/Create.php
RQuadling: I don't know how GitHub is handling things, but I've got 2 feature branches open - one for each issue. But 1 repo for all the work.

I use SourceTree with git-flow.

Not sure what I should be doing to keep these separate. Seems Github deals with repos and not branches when it comes to pull requests. So, does that mean I need multiple repos to have multiple pull requests? Please advise.

RQuadling:
Hmmm. I think I've completely buggered up my repo. Merged my develop branch back into a feature. So no wonder you are seeing inappropriate code in this pull request.

May be easier (and a lot cleaner) to start again with a new repo.

RQuadling:
Sorry for the noise. New PR https://github.com/robmorgan/phinx/pull/436

PR #424, opened 2014-12-29

RQuadling:
Feature branch, rather than my master.

Replacement for https://github.com/robmorgan/phinx/pull/407


src/Phinx/Console/Command/Create.php
```diff
@@ -112,6 +177,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ));
         }
 
+        // Do we need to do the post creation call to the creation class?
+        if (!empty($creationClassName)) {
```
RQuadling: One of our use cases for a callback is to deal with views, stored procedures, triggers, etc..

As there is no way to delta a view/sp/trigger/etc, you have to drop it and recreate it. That means multiple SQL statements. For this, our callback prepares a set of SQL files that are to be processed with one of our AbstractMigration implementations and a corresponding template (hence the reason for this entire pull request).

By having them as separate files, we get visual syntax highlighting as well as version control and frankly, SQL is easier to read/format outside of PHP.

Edge case? For those working with views/sps/triggers and wanting some sort of version control for when they change due to the business need, this seemed a good enough fit for us.

If there is any other way we can implement version controlled changes to the views/etc., then I'd certainly like to know about it!

lorenzo: Yeah, but if you have a separate class responsible of creating the template file, you could add your custom post-process logic there. No need to for that into phinx api.

RQuadling:
Sorry for the noise. New PR https://github.com/robmorgan/phinx/pull/437.

shadowhand:
Is this still relevant, or has it been replaced by #437?

RQuadling:
Sorry. 437 replaces this.

PR #431, opened 2015-01-06
OO design for Phinx within an application

alexbowers-tecmark:
This is a nicer design (in my opinion) for web based migration

### DIFF
--- a/app/web.php
+++ b/app/web.php
@@ -58,9 +58,6 @@
 
             $output = call_user_func([$wrap, 'getStatus'], $environment);
             $error  = $wrap->getExitCode() > 0;
-            if($error) {
-                return $error;
-            }
-            return $output;
+            return $error
         }
     }
--- a/app/web.php
+++ b/app/web.php
@@ -39,45 +39,28 @@
 // - (boolean) "debug" enable debugging?
 
 // Get the phinx console application and inject it into TextWrapper.
-$app = require __DIR__ . '/phinx.php';
-$wrap = new Phinx\Wrapper\TextWrapper($app);
+ class Phinx {
+        public static function migrate($environment = 'default') {
+            $app = require __DIR__ . '/phinx.php';
+            $wrap = new Phinx\Wrapper\TextWrapper($app);
 
-// Mapping of route names to commands.
-$routes = [
-    'status'   => 'getStatus',
-    'migrate'  => 'getMigrate',
-    'rollback' => 'getRollback',
-    ];
+            $output = call_user_func([$wrap, 'getMigrate'], $environment);
+            $error  = $wrap->getExitCode() > 0;
+            if($error) {
+                return $error;
+            }
+            return true;
+        }
 
-// Extract the requested command from the URL, default to "status".
-$command = trim(parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH), '/');
-if (!$command) {
-    $command = 'status';
-}
+        public static function status($environment = 'default') {
+            $app = require __DIR__ . '/phinx.php';
+            $wrap = new Phinx\Wrapper\TextWrapper($app);
 
-// Verify that the command exists, or list available commands.
-if (!isset($routes[$command])) {
-    $commands = implode(', ', array_keys($routes));
-    header('Content-Type: text/plain', true, 404);
-    die("Command not found! Valid commands are: {$commands}.");
-}
-
-// Get the environment and target version parameters.
-$env    = isset($_GET['e']) ? $_GET['e'] : null;
-$target = isset($_GET['t']) ? $_GET['t'] : null;
-
-// Check if debugging is enabled.
-$debug  = !empty($_GET['debug']) && filter_var($_GET['debug'], FILTER_VALIDATE_BOOLEAN);
-
-// Execute the command and determine if it was successful.
-$output = call_user_func([$wrap, $routes[$command]], $env, $target);
-$error  = $wrap->getExitCode() > 0;
-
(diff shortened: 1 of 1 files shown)


app/web.php
```diff
@@ -39,45 +39,28 @@
...
-// Mapping of route names to commands.
-$routes = [
-    'status'   => 'getStatus',
-    'migrate'  => 'getMigrate',
-    'rollback' => 'getRollback',
-    ];
+            $output = call_user_func([$wrap, 'getMigrate'], $environment);
+            $error  = $wrap->getExitCode() > 0;
+            if($error) {
+                return $error;
```
shadowhand: this is effectively:

```
if ($error) {
    return true;
}
return true;
```

alexbowers-tecmark: doesnt $error contain a string if it is true?

shadowhand:
What's the point of this? All your doing is replacing an example web app with a class that proxies some of `TextWrapper` methods.

app/web.php
```diff
@@ -39,45 +39,28 @@
...
+            $wrap = new Phinx\Wrapper\TextWrapper($app);
 
-// Mapping of route names to commands.
-$routes = [
-    'status'   => 'getStatus',
-    'migrate'  => 'getMigrate',
-    'rollback' => 'getRollback',
-    ];
+            $output = call_user_func([$wrap, 'getMigrate'], $environment);
+            $error  = $wrap->getExitCode() > 0;
```
shadowhand: No, error is a boolean, as per this line.

alexbowers-tecmark:
The point of it is that in your code you can simply call Phinx::migrate() now to perform a migration. It is just a wrapper around a wrapper, however it is clean to use. I will look into improvements if this sort of design is desired by the community.

app/web.php
alexbowers-tecmark: Ah my mistake. I'll fix that then.

shadowhand:
Calling static methods does not make a piece of code "OO". This offers no benefit over:

```php
$phinx = require 'vendor/app/phinx.php';
$wrap = new Phinx\Wrapper\TextWrapper($phinx);
$output = $wrap->getMigrate();
```

app/web.php
```diff
@@ -39,45 +39,25 @@
...
-    $args = implode(', ', [var_export($env, true), var_export($target, true)]);
-    echo "DEBUG: $command($args)" . PHP_EOL . PHP_EOL;
-}
-echo $output;
+ class Phinx {
+        public static function migrate($environment = 'default') {
+            $app = require __DIR__ . '/phinx.php';
+            $wrap = new Phinx\Wrapper\TextWrapper($app);
+
+            $output = call_user_func([$wrap, 'getMigrate'], $environment);
```
shadowhand: Why is `call_user_func` necessary here?

app/web.php
```diff
@@ -39,45 +39,25 @@
...
+            $error  = $wrap->getExitCode() > 0;
+            if($error) {
+                return $error;
```
shadowhand:
```php
if (true) {
  return true;
}
return true;
```

app/web.php
```diff
@@ -39,45 +39,25 @@
...
-header('Content-Type: text/plain', true, $error ? 500 : 200);
-if ($debug) {
-    // Show what command was executed based on request parameters.
-    $args = implode(', ', [var_export($env, true), var_export($target, true)]);
-    echo "DEBUG: $command($args)" . PHP_EOL . PHP_EOL;
-}
-echo $output;
+ class Phinx {
+        public static function migrate($environment = 'default') {
+            $app = require __DIR__ . '/phinx.php';
```
shadowhand: Every method call will create a new instance of Phinx?

alexbowers-tecmark:
This is just a quick POC to see whether people have any comments on this style. I plan to look into it more deeply tonight and actually make it cleaner and more usable.

app/web.php
alexbowers-tecmark: It isn't, this is very rough code. I suppose a pull request was a bad idea, I should really have done a RFC instead. My mistake

app/web.php
alexbowers-tecmark: In my final version of this which I hope to get some done tonight, I will be rewriting this to not require a new instance every time, and to instead use singletons

shadowhand:
Feel free to create a `Phinx\Wrapper\Singleton` class or something, but please do not outright replace `app/web.php` with something completely unrelated.

I'm not opposed to your approach, but the current implementation is very sloppy and not placed correctly. With some clean up, it could be a very useful contribution!

alexbowers-tecmark:
Okay. Will do. Apologies. I should have instead created an issue with a RFC. But i guess this had the same effect, again. Sorry. Will try to improve it tonight / some time this week

MartinDesigner:
Regarding only to design, look at this stuff:
imghome
So, as a supplement, however as a main point try this amazing free app. Createer.com is an online software for easily designing online ads, responsive landing pages and print materials such as flyers, posters, covers, invitations and more. Createer lets you easily share your designs on social media, send them via email, download as HTML, image or PDF. You can even design UI mockups for mobile as for standard display (html, js, css always generating without any coding required from you). Createer also includes 3,000 predesigned templates (browse here: https://beta2.createer.com/Templates?ghi) and over 250,000 images (illustrations, photos and icons) to work with in its user-friendly drag’n’drop enviroment. And finally this amazing stuff is here: https://beta2.createer.com/?ghi

robmorgan:
closing for now until I see more support from the community.

Issue #406, opened 2014-12-18
test style 


trilopin:
i have two ideas for testing:

**Replace assertions based in tested class methods**

``` php
public function testCreateTable()
{
        $table = new \Phinx\Db\Table('ntable', array(), $this->adapter);
        $table->addColumn('realname', 'string')
               ->addColumn('email', 'integer')
               ->save();
        $this->assertTrue($this->adapter->hasTable('ntable'));
        $this->assertTrue($this->adapter->hasColumn('ntable', 'id'));
        $this->assertTrue($this->adapter->hasColumn('ntable', 'realname'));
        $this->assertTrue($this->adapter->hasColumn('ntable', 'email'));
        $this->assertFalse($this->adapter->hasColumn('ntable', 'address'));
}
```

This test check CreateTable feature but use hasTable and hasColumn for assertions. 
If hasColumn method break, all tests with assert like those will fail (plus hasColumn itself) and test results wll be more complicated to read.

In the other hand, rewrite tests with pure sql for assertions is more complex to mantain...

**Make some unit tests**

Current tests are functional, execute code from start to end and assert results.
It will be nice have some unit tests with database mocking:
- this tests will be faster
- mocking real db, expecting generated sql -> it can be readed as sql generation documentation  

I say this as a proposal, not an attack of current tests.
I can help in both ideas (with mysql at least) 

what do you think?

Thanks

shadowhand:
@trilopin are you going to submit a pull request with these new tests? :)

trilopin:
@shadowhand just created it, i was waiting for some review but is better to do in pr.

PR #432, opened 2015-01-06
Unit tests for mysql adapter 

trilopin:
Add real unit tests for mysql adapter, refs #406 issue.

While coding this tests i saw #419, #418 and #417.

Some tricks as:
* implement a child class from real class to test protected/private methods
* make testeable and mockable PDO 
* reimplement method setConnection without typehinting.

This tricks are required for run with fake PDO and capture all message passing.

Some tests are not totally pure unit because have internal 'new' calls:
* testGetColumns has new Column
* testCreateTable* has new Column

The way for resolve this problem is refactoring adapter with dependency injection in parameters but I tried not to be too intrusive. 

Also, we have the folder structure... I didn't know where to put the unit test. 
I think ideal location is test/unit/Phinx/Db/Adapter/MysqlAdapterTest.php and moving all old tests to tests/functional but again is a big movement. For now, the tests is located in same folder than funcional with unit suffix.

If this code is finally merged, some tests from current mysqladapter test, must be deleted (duplicated in both: unit and functional).

Please, review and send your comments.

I will complete a few code not covered.

### DIFF
--- a/tests/Phinx/Db/Adapter/MysqlAdapterUnitTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterUnitTest.php
@@ -1,0 +1,1379 @@
+<?php
+
+namespace Test\Phinx\Db\Adapter;
+
+use Symfony\Component\Console\Output\NullOutput;
+use Phinx\Db\Adapter\MysqlAdapter;
+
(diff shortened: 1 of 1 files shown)


shadowhand:
Awesome, thanks!

Issue #433, opened 2015-01-07
Set Configuration path via TextWrapper


alexbowers:
I have been trying to find a way to set the configuration path (the phinx.yml location) via the TextWrapper but so far have come up empty. Any ideas how this can be done?

@shadowhand

shadowhand:
Currently not possible because you can't inject options directly into TextWrapper. Best thing to do would be to replace `$default_env` with `array $options` in the `__construct` to allow overloading options.

shadowhand:
Closed because question is answered, but certainly reference this issue if you make a PR.

alexbowers:
The problem I see is that environments and so on can be set after migration, but configuration has to be set before

IE. via command line it is

phinx -c "name" migrate -e default

Or is this a non-issue?

alexbowers:
@shadowhand I have created a pull request. Please could you look it over?

alexbowers:
Example Code Usage:
```php
<?php
	$app = new \Phinx\Console\PhinxApplication();
	$wrap = new \Phinx\Wrapper\TextWrapper($app, array(
                'config_path' => '../phinx.yml',
                'parser' => 'yaml'
        ));
	$test = $wrap->getMigrate();

	print_r($test);
```

It works without the second parameter. 
Either, all or none of the param values are required and can be passed. parser is an assumed value based on extension of the config_path, default config path is ./phinx.yml, Also, you can pass through the environment.

Issue #413, opened 2014-12-23
Dependency Injection / Inversion


shadowhand:
To allow Phinx to be fully extensible, it needs to apply [Dependency Inversion](http://en.wikipedia.org/wiki/Dependency_inversion_principle) to allow for dependency injection. The simplest way to approach DI compatibility would be to provide a "class map" that can be overloaded by configuration. This avoids the need to add a container dependency to Phinx.

Along with applying IoC, many methods that have hard-coded lists (eg `Column::setOptions`) should be split into multiple smaller methods to allow more precise modifications by extension. This helps satisfy the [LSP](http://en.wikipedia.org/wiki/Liskov_substitution_principle) part of [SOLID](http://en.wikipedia.org/wiki/SOLID_%28object-oriented_design%29).

Useful links on the subject:
- http://aspiringcraftsman.com/2008/12/28/examining-dependency-inversion/
- http://www.theserverside.com/news/1321158/A-beginners-guide-to-Dependency-Injection
- http://docs.structuremap.net/DependencyInjection.htm
- http://richardmiller.co.uk/2011/05/19/when-dependency-injection-goes-wrong/

dereuromark:
Anything here to be done still? Or can this be closed?

PR #434, opened 2015-01-07
Passing Config and Parser information through TextWrapper

alexbowers:


alexbowers:
#433 Linking for reference

alexbowers:
Example Code Usage:

<?php
    $app = new \Phinx\Console\PhinxApplication();
    $wrap = new \Phinx\Wrapper\TextWrapper($app, array(
                'config_path' => '../phinx.yml',
                'parser' => 'yaml'
        ));
    $test = $wrap->getMigrate();

    print_r($test);
It works without the second parameter. 
Either, all or none of the param values are required and can be passed. parser is an assumed value based on extension of the config_path, default config path is ./phinx.yml, Also, you can pass through the environment.

src/Phinx/Config/Config.php
```diff
@@ -49,6 +49,11 @@ class Config implements ConfigInterface
     protected $configFilePath;
 
     /**
+     * @var $phinx_version
+     */
+    public static $phinx_version = "0.4.1";
```
shadowhand: Why is this added here? We already have a `PHINX_VERSION` constant.

src/Phinx/Console/PhinxApplication.php
```diff
@@ -47,8 +48,12 @@ class PhinxApplication extends Application
      *
      * @param string $version The Application Version
      */
-    public function __construct($version)
+    public function __construct($version = null)
```
shadowhand: This is totally unrelated to the PR title/description. Please revert it.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -42,17 +42,20 @@ class TextWrapper
...
-     * @return void
+     * @param PhinxApplication $app
+     * @param array $options
      */
-    public function __construct(PhinxApplication $app, $default_env = 'development')
+    public function __construct(PhinxApplication $app, Array $options = array())
     {
         $this->app = $app;
-        $this->env = $default_env;
+        $this->env = (isset($options['environment']) ? $options['environment'] : 'development');
```
shadowhand: Why limit the access to options by assigning them in the constructor? Wouldn't it be better to store **all** the options and have a `getOption` method, like in `AdapterInterface`?

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -107,6 +110,45 @@ public function getRollback($env = null, $target = null)
         return $this->executeRun($command);
     }
 
+    /**
+     * Set the environment
+     *
+     * @param string $environment
+     */
+    public function setEnvironment($environment = null) {
```
shadowhand: What's the reason for these? Nothing seems to use them, and they can already be modified by passing the argument to the specific method.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -107,6 +110,45 @@ public function getRollback($env = null, $target = null)
...
+        if(is_null($environment)) {
```
shadowhand: I find it strange that you would define the arguments as optional with `= null`, then throw an exception if the **optional** argument is actually null.

src/Phinx/Config/Config.php
alexbowers: I added it for the PhinxApplication to use, since the PHINX_VERSION wasn't visible. I'll revert these changes as per your other message though.

shadowhand: It's injected by `app/phinx.php`.

src/Phinx/Wrapper/TextWrapper.php
alexbowers: This was done so that we could predict what is being passed through. I believe that these are the only possible values. Would it be better to accept any, and simply only use these three then?

shadowhand: It's better to accept an array, and then have a `getOption` method, like so:

```php
protected function getOption($name)
{
    if (!isset($this->options[$name])) {
        return null;
    }
    return $this->options[$name];
}
```

That way an option can be added in the future without having to modify code in two places.

src/Phinx/Wrapper/TextWrapper.php
alexbowers: This was so that we have an alternative option in setting the environment, based on user preference. They could either write it as this:
```php
<?php
$app = new PhinxApplication();
$wrap = new TextWrapper($app, array(
    ...
));
```

Or it could be done by:

```php
<?php
$app = new PhinxApplication();
$wrap = new TextWrapper($app);
$wrap->setEnvironment('Environment Name');
```

They would both do the same thing, but its just a different way for the user to write it.

src/Phinx/Wrapper/TextWrapper.php
alexbowers: Yeah, im not sure why I did that, I'll just make it a required argument.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -71,7 +74,7 @@ public function getExitCode()
      */
     public function getStatus($env = null)
     {
-        $command = ['status', '-e' => $env ?: $this->env];
```
shadowhand: Which then makes code like this easier:

```php
$command = ['status', '-e' => $env ?: $this->getOption('environment')];
```

src/Phinx/Config/Config.php
alexbowers: Would it not make sense to have it at a deeper level than there?

src/Phinx/Wrapper/TextWrapper.php
shadowhand: Also, default options can be defined like so:

```php
$options += array(
    'environment' => 'development',
);
```

src/Phinx/Config/Config.php
shadowhand: No, I don't think so. Dependency Inversion implies that the more dynamic or configurable something is, the **shallower** it should be. See #413 for additional detail on this topic.

alexbowers: Okay, I'll remove this. To me it seems messy to have to define PhinxApplication with a version number though, for what seems like purely to be used for the initial output line to report the version. That number can't be trusted much can it?

shadowhand: Nothing can be trusted, this is programming. ;) FWIW, that's just how [Symfony Console Application works](http://api.symfony.com/2.4/Symfony/Component/Console/Application.html#method___construct).

alexbowers:
@shadowhand Changes made. Is this better?

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -107,6 +113,30 @@ public function getRollback($env = null, $target = null)
...
+	 */
+    protected function getOption($key)
+    {
+        if(!isset($this->options[$key])) {
+	        return null;
+        }
+	    return $this->options[$key];
+	}
+
+	/**
```
shadowhand: Indentation is all screwy here.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -107,6 +113,30 @@ public function getRollback($env = null, $target = null)
...
+	 * Set option in options array
+	 *
+	 * @param $key
```
shadowhand: Missing type hint and return in this docblock.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -107,6 +113,30 @@ public function getRollback($env = null, $target = null)
...
+	 * @param $value
+	 */
+	public function setOption($key, $value) {
+		$this->options[$key] = $value;
+	}
```
shadowhand: Ideally, setters will `return $this`, for fluent usage.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -41,18 +41,23 @@
...
+     * @param PhinxApplication $app
+     * @param array $options
      */
-    public function __construct(PhinxApplication $app, $default_env = 'development')
+    public function __construct(PhinxApplication $app, Array $options = array())
     {
+        $options += array(
+            'environment' => 'development',
+            'config_path' => './phinx.yml',
+            'parser' => 'yaml'
```
shadowhand: No need to set parser, it is automatically determined by config file extension.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -41,18 +41,23 @@
...
-     * @return void
+     * @param PhinxApplication $app
+     * @param array $options
      */
-    public function __construct(PhinxApplication $app, $default_env = 'development')
+    public function __construct(PhinxApplication $app, Array $options = array())
     {
+        $options += array(
+            'environment' => 'development',
+            'config_path' => './phinx.yml',
```
shadowhand: It would be better if this option name matched the command line name "configuration".

alexbowers:
@shadowhand I think this is all the changes made. Any other thoughts?

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -40,88 +40,119 @@
...
-        $this->exit_code = $this->app->doRun(new ArrayInput($command), new StreamOutput($stream));
-
-        // Get the output of the command and close the stream, which will
-        // destroy the temporary file.
-        $result = stream_get_contents($stream, -1, 0);
-        fclose($stream);
-
-        return $result;
-    }
+	private $app;
```
shadowhand: You replaced all space indents with tab indents, which violates PSR style. Please fix.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -40,88 +40,119 @@
...
+	private $options;
+	private $exit_code;
+
+	/**
+	 * @param PhinxApplication $app
+	 * @param array $options
+	 */
+	public function __construct(PhinxApplication $app, Array $options = array())
```
shadowhand: `Array` type hint should be lowercase.

alexbowers: Elsewhere it is uppercase too 
```php
protected function executeRun(Array $command)
	{
		// Output will be written to a temporary stream, so that it can be
		// collected after running the command.
		$stream = fopen('php://temp', 'w+');

		// Execute the command, capturing the output in the temporary stream
		// and storing the exit code for debugging purposes.
		$this->exit_code = $this->app->doRun(new ArrayInput($command), new StreamOutput($stream));

		// Get the output of the command and close the stream, which will
		// destroy the temporary file.
		$result = stream_get_contents($stream, -1, 0);
		fclose($stream);

		return $result;
	}
```
(TextWrapper.php:141)

shadowhand: That's my mistake from when I wrote the original. Everywhere else in Phinx is lowercase. Doesn't really matter, just good to be consistent.

alexbowers: Want me to change yours too while I'm at it?

shadowhand: Yes please.

src/Phinx/Wrapper/TextWrapper.php
shadowhand: Tabs still need to be replaced.

alexbowers:
@shadowhand All of the spaces should be there now. I think it was an issue with me making an edit with Githubs edit feature since it was set to tab indent by default for me. This should be correct now since I have fixed it in phpstorm.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -1,127 +1,182 @@
...
-{
-    private $app;
-    private $env;
-    private $exit_code;
-
-    /**
-     * @param  Phinx\Console\PhinxApplication $app
-     * @param  String                         $default_env  environment fallback, defaults to "development"
-     * @return void
+    /* Phinx
```
shadowhand: Now all this is indented incorrectly... please review your own changes before pushing to Github.

alexbowers:
@shadowhand Sorry about the indenting. Fixed it now

PR #435, opened 2015-01-08
remove duplicate test + additional unit tests 

trilopin:
Refs #406, Duplicated tests related to PhinxType removed from functional test.
More testcases for int and text types and index definition.

Only connect method is not tested due to PDO static methods.


tests/Phinx/Db/Adapter/MysqlAdapterUnitTest.php
```diff
@@ -788,12 +806,33 @@ public function testGetSqlTypeExists()
         $this->assertEquals(array('name' => 'varchar', 'limit' => 255),
                             $this->adapter->getSqlType(MysqlAdapter::PHINX_TYPE_STRING));
         $this->assertEquals(array('name' => 'char', 'limit' => 255),
-                            $this->adapter->getSqlType(MysqlAdapter::PHINX_TYPE_CHAR));
+                            $this->adapter->getSqlType(MysqlAdapter::PHINX_TYPE_CHAR, 255));
+
+        //text combinations
+        $this->assertEquals(array('name' => 'tinytext'),
+                            $this->adapter->getSqlType(MysqlAdapter::PHINX_TYPE_TEXT, 512));
```
shadowhand: I think it would be more appropriate to use the defined constants `MysqlAdapter::TEXT_TINY` etc, rather than using hard numbers.

trilopin: you are right

I have updated test with constants when is equal or greater than limit.

shadowhand:
Thanks!

PR #436, opened 2015-01-08
Support any namespace\class for the migration base class.

RQuadling:
Allow migration base class to be any namespace\class.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -243,8 +243,6 @@ abstract class AbstractCommand extends Command
 
         $output->writeln('<info>using config parser</info> ' . $parser);
 
-        $output->writeln('<info>using migration base class</info> ' . $config->getMigrationBaseClassName(false));
-
         $this->setConfig($config);
     }
 
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -242,6 +242,9 @@ abstract class AbstractCommand extends Command
         }
 
         $output->writeln('<info>using config parser</info> ' . $parser);
+
+        $output->writeln('<info>using migration base class</info> ' . $config->getMigrationBaseClassName(false));
+
         $this->setConfig($config);
     }
 
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -105,7 +105,7 @@ class Create extends AbstractCommand
             '$className'     => $className,
             '$baseClassName' => $this->getConfig()->getMigrationBaseClassName(true),
         );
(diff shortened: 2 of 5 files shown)


src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -242,6 +242,9 @@ protected function loadConfig(InputInterface $input, OutputInterface $output)
         }
 
         $output->writeln('<info>using config parser</info> ' . $parser);
+
+        $output->writeln('<info>using migration base class</info> ' . $config->getMigrationBaseClassName(false));
```
shadowhand: This could be useful for the `create` command, but I don't think it's necessary for every command.

src/Phinx/Console/Command/Create.php
```diff
@@ -99,12 +99,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        // inject the base class name
-        $contents = str_replace('$baseClassName', $this->getConfig()->getMigrationBaseClassName(), $contents);
-        
+        // inject the class names appropriate to this migration
+        $classes = array(
+            '$useClassName'  => $this->getConfig()->getMigrationBaseClassName(false),
+            '$className'     => $className,
+            '$baseClassName' => $this->getConfig()->getMigrationBaseClassName(true),
+        );
+        $contents = str_replace(array_keys($classes), $classes, $contents);
```
shadowhand: This usage of `str_replace` is effectively the same as:

```php
$contents = strtr($contents, $classes);
```

RQuadling: Nice one. Not used that before.

src/Phinx/Console/Command/AbstractCommand.php
RQuadling: I personally would also be showing the template being used. With consideration of my other PR regarding alternative templates and template generating class, maybe moving them all to the ``create`` command would be a sensible place. Only showing the actual use (i.e. if overriding the template, then don't show the base template name). Happy with whatever you feel is appropriate. I just need the functionality. :-D

src/Phinx/Console/Command/Create.php
```diff
@@ -112,6 +114,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ));
         }
 
+        $output->writeln('<info>using migration base class</info> ' . $this->getConfig()->getMigrationBaseClassName(false));
```
shadowhand: Rather than doing a whole new `getMigrationBaseClassName` call here, why not use `$classes['$useClassName']`?

RQuadling: Because .... nope ... no reason ! ;-)

shadowhand:
Didn't realize this was targeting the wrong branch. In the future, please only target the `-dev` branches.

Corrected merge commit is a67a301ad8d8e85a2e0bcbabfa6ef2c03be6b91b.

PR #438, opened 2015-01-08
Changed behaviour of the MySQL adapter create method's default ID column...

maxbrokman:
... to an unsigned integer. ID columns should be unsigned as they can never be negative (auto incrementing).

### DIFF
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -157,7 +157,7 @@ class MysqlAdapterTest extends \PHPUnit_Framework_TestCase
 
         $table = new \Phinx\Db\Table('ntable', array(), $this->adapter);
         $table->addColumn('realname', 'string')
-              ->addColumn('tag_id', 'integer')
+              ->addColumn('tag_id', 'integer', array('signed' => false))
               ->addForeignKey('tag_id', 'ntable_tag', 'id', array('delete'=> 'NO_ACTION', 'update'=> 'NO_ACTION'))
               ->save();
 
@@ -710,7 +710,7 @@ class MysqlAdapterTest extends \PHPUnit_Framework_TestCase
         $refTable->addColumn('field1', 'string')->save();
 
         $table = new \Phinx\Db\Table('table', array(), $this->adapter);
-        $table->addColumn('ref_table_id', 'integer')->save();
+        $table->addColumn('ref_table_id', 'integer', array('signed' => false))->save();
 
         $fk = new \Phinx\Db\Table\ForeignKey();
         $fk->setReferencedTable($refTable)
@@ -727,7 +727,7 @@ class MysqlAdapterTest extends \PHPUnit_Framework_TestCase
         $refTable->addColumn('field1', 'string')->save();
 
         $table = new \Phinx\Db\Table('table', array(), $this->adapter);
-        $table->addColumn('ref_table_id', 'integer')->save();
+        $table->addColumn('ref_table_id', 'integer', array('signed' => false))->save();
 
         $fk = new \Phinx\Db\Table\ForeignKey();
         $fk->setReferencedTable($refTable)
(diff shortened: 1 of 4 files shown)


shadowhand:
What about line 219?

shadowhand:
Tests need to be updated to match new expectations.

maxbrokman:
Eek, missed that second table creation section. Also actually committed the tests this time.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -157,7 +157,7 @@ public function testCreateTableWithForeignKeys()
 
         $table = new \Phinx\Db\Table('ntable', array(), $this->adapter);
         $table->addColumn('realname', 'string')
-              ->addColumn('tag_id', 'integer')
+              ->addColumn('tag_id', 'integer', array('signed' => false))
```
shadowhand: Right, so this is exactly what worries me about this PR... by changing the default expectation for PK columns, we suddenly have to worry about every single FK reference too. From my standpoint, this PR drives MySQL adapter further from compatibility with other adapters, by making Phinx less consistent.

Thoughts?

maxbrokman: This is certainly a valid worry. I haven't used Phinx for any other SQL dialect but looking at e.g PostgresAdapter the behaviour there is the same in that it uses integer (signed) rather than serial (unsigned).

I am not familiar enough with Postgres to say what the common behaviour for auto-incrementing ids is, but in my experience with MySQL they are always unsigned. 

The PR is motivated by implementing Phinx in an existing large codebase / database. Applying foreign keys fails because the column types of the referencing / referenced columns do not match.  Our workaround with Phinx in production is to always specify our own primary key, and this is what brought me to write the PR.

I have to say that I had no considered portability until this point.

MySQL / Postgres could possibly be made compatible by applying a similar change and internally juggling the type of the column to Serial when it is set to unsigned? Looks like this might be the current behaviour for Identity columns in PostgresAdapter::getColumnSqlDefinition but would require some changes to the signature of getSqlType for other columns. 

I do not know enough about FKs in Postgres to comment on how they are used there, and know even less about SqlServer.

shadowhand: @robmorgan could use your input on this ^ I'm not really sure what the right answer here is.

maxbrokman: Just going off the docs, isn't 'signed' ignored outside of MySQL anyway? Certainly needs further input.

shadowhand: Yes, `signed` only affects MySQL.

fightbulc:
A general input on that topic since I am struggling all the time with this:

I should be able to dismiss the auto ID creation (as implemented in form of id => false) in order to create my own field. All what is missing at that point is a possibility to set ```auto_increment => true```

maxbrokman:
There is a work around for this, you can call make a new Column and call setIdentity (or similar) on it make it auto increment, and then pass this to addColumn

shadowhand:
@fightbulc see http://docs.phinx.org/en/latest/migrations.html#creating-a-table for syntax on specifying an alternate PK. As of Phinx 0.4.2 (next release) you can use `'identity' => true` option to force the column to be auto-increment, see 70fee0fd9764c2d6fe41d8c2c72291c9cdd6393b.

maxbrokman:
As I understand it the alternative PK syntax doesn't actually make the column you specify auto increment.

shadowhand:
@maxbrokman If you look at the commit in question, and the adapter code, you will see that's exactly what `identity` does.

maxbrokman:
Sorry I was referring to the current behaviour

fightbulc:
@shadowhand great. looking forward to 0.4.2.

robmorgan:
I'm closing this. Signed integer PKs in MySQL are sufficient for most people, if you need to override you can. This is also the behaviour in Rails + ActiveRecord.

PR #440, opened 2015-01-08
Documentation for Phinx via Code added

alexbowers:
Documentation Added for calling Phinx via code.

### DIFF
--- a/docs/code-usage.rst
+++ b/docs/code-usage.rst
@@ -1,0 +1,178 @@
+.. index::
+   single: Phinx via Code
+
+Calling Phinx via Code
+===================================
+
+Phinx is primarily a command line tool, however Phinx does support code-based calls.
+
+.. warning::
+
+    Not all terminal commands are available to be used within code yet.
+
+Initial Setup
+--------------
+
+.. note::
+
+    For Phinx via Code to work at all, you need to ensure that Phinx is accessible via autoloading,
+    the best way to manage this would be to ``require "vendor/autoload.php";`` the composer autoloader.
+
+Phinx will require instantiating, which can be achieved by the following.
+
+.. code-block:: php
+
+    $app = new \Phinx\Console\PhinxApplication($version);
+    $phinx = new \Phinx\Wrapper\TextWrapper($app);
+
+Where ``$version`` is the version number of Phinx (eg. '0.4.1').
+``$phinx`` now contains our Phinx object which will expose to us the commands.
+
+
+
+Custom Options
+---------------
+
+Phinx comes with some sensible options set for us, however these are not always appropriate and may need changing.
+
+There are two ways an option value can be set.
+
+The first way is instead of the ``$phinx`` declaration above, we can pass through a second parameter ``$options`` as an array of options.
+
+.. code-block:: php
+
+    $phinx = \Phinx\Wrapper\TextWrapper($app, array(
+        'option_name' => 'option_value',
+    ));
+
+To set multiple options and values, simply add more key value pairs to the ``$options`` array.
+
+The second way uses the original ``$phinx`` object and uses the ``setOption($option, $value)`` method.
+
+This method can be invoked multiple times, and either chained together or with each invokation done separately.
+
+.. code-block:: php
+
+    $phinx->setOption('option_name', 'option_value');
+
+The three options we can alter are:
+
+1. Environment name
+2. Configuration File Location
+3. Parser Information
+
+.. note::
+
+    There is no need to define these options if they are the same as the defaults.
+
+Configuring Environment Name
+-----------------------------
+
+Setting the Environment Name is done using the option name ``environment``.
+The value you assign should match the config file environment declared (eg. ``development``)
+
+.. code-block:: php
+
+    $phinx->setOption('environment', 'development');
+
+The default value if left blank is: ``development``.
+
+Configuring Configuration File Location
+----------------------------------------
+
+Setting the Configuration File Location is done using the option name ``configuration``.
+The value you assign should match the relative or absolute path to the phinx.yml (or alternative) file.
+
+.. code-block:: php
+
+    $phinx->setOption('configuration', '/var/www/phinx.yml');
+
+.. note::
+
+    The location may be relative to the webroot if you use an MVC framework.
+    Using ``__DIR__`` for relative paths or an absolute path is recommended.
+
+The default value if left blank is: ``./phinx.yml``.
+
+Configuring Parser Information
+-------------------------------
+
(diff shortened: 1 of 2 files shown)


docs/code-usage.rst
```diff
@@ -0,0 +1,178 @@
...
+    the best way to manage this would be to ``require "vendor/autoload.php";`` the composer autoloader.
+
+Phinx will require instantiating, which can be achieved by the following.
+
+.. code-block:: php
+
+    $app = new \Phinx\Console\PhinxApplication($version);
+    $phinx = new \Phinx\Wrapper\TextWrapper($app);
+
+Where ``$version`` is the version number of Phinx (eg. '0.4.1').
```
shadowhand: This works, but it might be better to use:

```php
$app = require 'vendor/robmorgan/phinx/app/phinx.php';
```

docs/code-usage.rst
```diff
@@ -0,0 +1,178 @@
...
+    If left blank, the value is assumed based on the Configuration File's file extension as per the list above.
+
+Migrating
+----------
+
+This command will run the same command as ``phinx migrate``.
+
+.. code-block:: php
+
+    $phinx->getMigrate($env, $target);
```
shadowhand: I think this example should include the return value:

```php
$output = $phinx->getMigrate(...);
```

docs/code-usage.rst
alexbowers: I have to disagree.

Many people will already have composer's autoloading defined, and thus requiring phinx.php will autoload PhinxApplication twice.

Also, I think this:
    $app= new PhinxApplication();
is clearer than
    $app= require 'vendor/robmorgan/phinx/app/phinx.php';
because its immediately clear whats going on.

alexbowers: I just saw your edit.

I agree that having an embedded one would be nice actually. Though, to me it would still be nicer if we had a Facade style class which was simply called Phinx which wrapped the TextWrapper and some of its methods

Having Phinx->migrate() is clearer than TextWrapper->getMigrate()

docs/code-usage.rst
alexbowers: Wouldn't the most common use case of Phinx via Code be without needing the return value for the migration. Since it is meaningless to the end-user (who is either running an install script, or something).

I can add it if you want, but the use case is more limited than without it.

shadowhand: The above comment and this comment make sense... Some modifications could be made that would clean all this up:

1. `$phinx = new Phinx\Phinx()` to get a fluent object
2. `$phinx->migrate($target = null)` and `$phinx->rollback($target = null)` that return a  boolean (based on exit code)
3. updated documentation

In many ways, this points back #413 and suggests that Phinx should be internally refactored to exist as a library, which also provides a console application that **uses** the library, rather than having the opposite situation (as it is currently).

alexbowers: To me, what makes sense is to have 1 and 2 as you specified. However, in a way that the message is also split down and segmented into an array of messages and timestamps etc, so that if the user wished to get the response, such as 'xxx migrated 0.2s' they can simply call a method like $phinx->getMessage(); and it will have the same message as $phinx->migrate() currently has, and with work done on the message too, we can provide it in many formats / split up.

I have to agree with the use of a library rather than a command.

alexbowers: Shall we commit it as is for now, and have an RFC for refactoring into a library?

alexbowers: @shadowhand Shall we open this up as a RFC?

shadowhand: I'd like to see the code and the documentation together in a single PR.

alexbowers: Hasn't the code already been merged in?

shadowhand: Not the fluent object we're talking about.

alexbowers: Ah sorry. I thought you were meant that was a future change. I'll start
work on that tonight

shadowhand: Have you made any progress on the fluent class, or should I start working on it?

alexbowers: I haven't had chance to start yet. If you can make it a branch I may have time to contribute a bit to it tonight, but i've started on a PoC for the company I work for.

shadowhand:
@alexbowers check out https://github.com/robmorgan/phinx/tree/i413-fluent-phinx

shadowhand:
That's only a basic starting point... I'd actually like to see a bigger refactor project that inverts Phinx more completely and makes the fluent interface the core part, and the console application a wrapper.

alexbowers:
I don't seem to be able to add comments to those files.

But a few thoughts from me so far:

All numbers refer to line numbers:

64: I prefer to have paths which have ../ ../ in them to be `dirname(dirname(__DIR__)). That way, when debugging, we get the actual path, rather than a convoluted way to get to it. 

97, 110: Not a particarly nice set of parameters there, passing null through first. Maybe TextWrapper could be refactored to remove them since we can't pass through values to it anymore?

alexbowers:
@shadowhand Is there anything else required for a merge?

Also, the documentation requires updating again. I'm a little busy at the minute, so if you wouldn't mind could you please update the docs with the new code examples and PR them into my branch?

If not, I'll probably get around to them tomorrow.

alexbowers:
@shadowhand Documentation has been updated.

alexbowers:
@shadowhand Any idea when this will be merged in along with Fluent.

I have an application which will require updating when this is merged; and having it done sooner rather than later would be nice.

shadowhand:
@alexbowers honestly, creating the Fluent interface doesn't make a lot of sense without the refactor, which is a large task and cause the code to diverge significantly. There are already multiple ways to work around this without the fluent interface in your application, or you can import the existing class I created into your project.

tl;dr: there are no plans to add the fluent interface soon, handle it in your own project for now.

alexbowers:
Okay. I was going to look into the refractor this evening. Is it not worth
me looking into it yet?

shadowhand:
@alexbowers I'm not opposed to you trying, but it requires significant understand of the codebase and how the 3 parts (manager, env, migrations) all work together.

alexbowers:
@shadowhand Any chance of getting this docs merged in too?

shadowhand:
@alexbowers honestly I would prefer not to document/encourage usage of the current interfaces until the underlying architecture is improved. Even the current names are pretty bad.

alexbowers:
Okay. can we draft up a RFC at some point if we're heading towards backwards incompatibility?

That will also allow some of us to fork off and create a new architecture

robmorgan:
I'm closing this. The current interfaces should not be documented as is.

PR #443, opened 2015-01-09
Update composer.json to use PSR-4

alexbowers:
PSR-0 is deprecated.
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,12 +28,12 @@
     },
     "autoload": {
         "psr-4": {
-            "Phinx": "src/Phinx"
+            "Phinx\\": "src/Phinx"
         }
     },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -27,13 +27,13 @@
         "squizlabs/php_codesniffer": "dev-phpcs-fixer"
     },
     "autoload": {
-        "psr-0": {
-            "Phinx": "src/"
+        "psr-4": {
+            "Phinx\\": "src/Phinx"
```
shadowhand: doesn't it need the trailing slash?

shadowhand: Hmm, tests passed, I guess not.

alexbowers: It does. I fixed it

PR #444, opened 2015-01-10

RQuadling:
Rebased my feature upon 0.4.x-dev 25eeeba. This replaces PR https://github.com/robmorgan/phinx/pull/437.

Really sorry for the noise. Still learning git/github.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -46,6 +46,11 @@ use Phinx\Db\Adapter\AdapterInterface;
 abstract class AbstractCommand extends Command
 {
     /**
+     * The location of the default migration template.
+     */
+    const DEFAULT_MIGRATION_TEMPLATE = '/../../Migration/Migration.template.php.dist';
+
+    /**
      * @var ConfigInterface
      */
     protected $config;
@@ -290,6 +295,6 @@ abstract class AbstractCommand extends Command
      */
     protected function getMigrationTemplate()
     {
-        return file_get_contents(dirname(__FILE__) . '/../../Migration/Migration.template.php.dist');
+        return file_get_contents(dirname(__FILE__) . self::DEFAULT_MIGRATION_TEMPLATE);
     }
 }
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -295,6 +295,6 @@ abstract class AbstractCommand extends Command
      */
     protected function getMigrationTemplate()
     {
-        return file_get_contents(dirname(__FILE__) . self::DEFAULT_MIGRATION_TEMPLATE);
+        return file_get_contents(__DIR__ . self::DEFAULT_MIGRATION_TEMPLATE);
     }
 }
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -3,7 +3,7 @@
  * Phinx
  *
  * (The MIT license)
- * Copyright (c) 2014 Rob Morgan
+ * Copyright (c) 2015 Rob Morgan
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated * documentation files (the "Software"), to
@@ -41,7 +41,7 @@ class Create extends AbstractCommand
     /**
      * The name of the interface that any external template creation class is required to implement.
      */
-    const CREATION_INTERFACE = '\Phinx\Migration\CreationInterface';
+    const CREATION_INTERFACE = 'Phinx\Migration\CreationInterface';
 
     /**
      * {@inheritdoc}
@@ -123,7 +123,7 @@ class Create extends AbstractCommand
         $fileName = Util::mapClassNameToFileName($className);
         $filePath = $path . DIRECTORY_SEPARATOR . $fileName;
 
-        if (file_exists($filePath)) {
+        if (is_file($filePath)) {
             throw new \InvalidArgumentException(sprintf(
                 'The file "%s" already exists',
                 $filePath
@@ -133,7 +133,7 @@ class Create extends AbstractCommand
         // Get the alternative template and static class options, but only allow one of them.
         $altTemplate = $input->getOption('template');
         $creationClassName = $input->getOption('class');
-        if (!empty($altTemplate) && !empty($creationClassName)) {
+        if ($altTemplate && $creationClassName) {
             throw new \InvalidArgumentException('Cannot use --template and --class');
         }
 
@@ -147,22 +147,18 @@ class Create extends AbstractCommand
 
         // Verify the static class exists and that it implements the required interface.
         if (!empty($creationClassName)) {
-            if (!class_exists($creationClassName, true)) {
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        'The class "%s" does not exist',
-                        $creationClassName
-                    )
-                );
+            if (!class_exists($creationClassName)) {
+                throw new \InvalidArgumentException(sprintf(
+                    'The class "%s" does not exist',
+                    $creationClassName
+                ));
             }
-            if (!is_subclass_of($creationClassName, self::CREATION_INTERFACE, true)) {
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        'The class "%s" does not implement the required interface "%s"',
-                        $creationClassName,
-                        self::CREATION_INTERFACE
-                    )
-                );
+            if (!is_subclass_of($creationClassName, self::CREATION_INTERFACE)) {
+                throw new \InvalidArgumentException(sprintf(
+                    'The class "%s" does not implement the required interface "%s"',
+                    $creationClassName,
+                    self::CREATION_INTERFACE
+                ));
             }
         }
 
@@ -195,16 +191,8 @@ class Create extends AbstractCommand
         }
 
         // Do we need to do the post creation call to the creation class?
-        if (!empty($creationClassName)) {
-            try {
-                $creationClass->postMigrationCreation($filePath, $className, $this->getConfig()->getMigrationBaseClassName());
-            } catch(\Exception $ex) {
-                throw new \RuntimeException(sprintf(
-                    'Problem calling %s->postMigrationCreation(), resulting in %s',
-                    $creationClassName,
-                    $ex->getMessage()
-                ));
-            }
+        if ($creationClassName) {
+            $creationClass->postMigrationCreation($filePath, $className, $this->getConfig()->getMigrationBaseClassName());
         }
 
         $output->writeln('<info>using migration base class</info> ' . $classes['$useClassName']);
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -28,15 +28,21 @@
  */
 namespace Phinx\Console\Command;
 
+use Phinx\Migration\CreationInterface;
 use Phinx\Migration\Util;
-use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 class Create extends AbstractCommand
 {
+    /**
+     * The name of the interface that any external template creation class is required to implement.
+     */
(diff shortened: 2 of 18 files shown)


src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -290,6 +295,6 @@ protected function verifyMigrationDirectory($path)
      */
     protected function getMigrationTemplate()
     {
-        return file_get_contents(dirname(__FILE__) . '/../../Migration/Migration.template.php.dist');
+        return file_get_contents(dirname(__FILE__) . self::DEFAULT_MIGRATION_TEMPLATE);
```
shadowhand: Always use `static` instead of `self`, to allow extending classes to easily overload.

src/Phinx/Console/Command/Create.php
```diff
@@ -38,6 +39,11 @@
 class Create extends AbstractCommand
 {
     /**
+     * The name of the interface that any external template creation class is required to implement.
+     */
+    const CREATION_INTERFACE = '\Phinx\Migration\CreationInterface';
```
shadowhand: No need for the leading `\`.

src/Phinx/Console/Command/Create.php
```diff
@@ -52,6 +58,13 @@ protected function configure()
                 PHP_EOL,
                 PHP_EOL
             ));
+
+        // An alternative template.
+        $this->addOption('template', 't', InputOption::VALUE_REQUIRED, 'Use an alternative template');
+
+        // A classname to be used to gain access to the template content as well as the ability to
+        // have a callback once the migration file has been created.
+        $this->addOption('class', 'l', InputOption::VALUE_REQUIRED, 'Use a class implementing "' . self::CREATION_INTERFACE . '" to generate the template');
```
shadowhand: This message should not use `CREATION_INTERFACE` constant, as it could be overloaded. Just use `"CreationInterface"`.

src/Phinx/Console/Command/Create.php
```diff
@@ -76,6 +89,10 @@ protected function getCreateMigrationDirectoryQuestion()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        /**
+         * @var CreationInterface $creationClass
```
shadowhand: What's this comment doing?

src/Phinx/Console/Command/Create.php
```diff
@@ -113,8 +130,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ));
         }
 
-        // load the migration template
-        $contents = $this->getMigrationTemplate();
+        // Get the alternative template and static class options, but only allow one of them.
+        $altTemplate = $input->getOption('template');
+        $creationClassName = $input->getOption('class');
+        if (!empty($altTemplate) && !empty($creationClassName)) {
```
shadowhand: The `empty` calls are unnecessary here, truthy calls are fine:

```php
if ($altTemplate && $creationClassName) { /* effectively the same as using !empty() */ }
```

src/Phinx/Console/Command/Create.php
```diff
@@ -113,8 +130,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            throw new \InvalidArgumentException('Cannot use --template and --class');
+        }
+
+        // Verify the alternative template file's existence.
+        if (!empty($altTemplate) && !file_exists($altTemplate)) {
```
shadowhand: Using `is_file` is recommended over `file_exists`, because the latter will also return `true` for directories.

src/Phinx/Console/Command/Create.php
```diff
@@ -113,8 +130,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            throw new \InvalidArgumentException(sprintf(
+                'The alternative template file "%s" does not exist',
+                $altTemplate
+            ));
+        }
+
+        // Verify the static class exists and that it implements the required interface.
+        if (!empty($creationClassName)) {
+            if (!class_exists($creationClassName, true)) {
```
shadowhand: The default for the `$autoload` parameter is already `true`, no need to explicitly provide it.

src/Phinx/Console/Command/Create.php
```diff
@@ -113,8 +130,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                throw new \InvalidArgumentException(
+                    sprintf(
+                        'The class "%s" does not exist',
+                        $creationClassName
+                    )
+                );
+            }
+            if (!is_subclass_of($creationClassName, self::CREATION_INTERFACE, true)) {
```
shadowhand: `static`

src/Phinx/Console/Command/Create.php
```diff
@@ -113,8 +130,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                'The alternative template file "%s" does not exist',
+                $altTemplate
+            ));
+        }
+
+        // Verify the static class exists and that it implements the required interface.
+        if (!empty($creationClassName)) {
+            if (!class_exists($creationClassName, true)) {
+                throw new \InvalidArgumentException(
+                    sprintf(
```
shadowhand: I would put `sprintf` on the same line as `throw`.

src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +194,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ));
         }
 
+        // Do we need to do the post creation call to the creation class?
+        if (!empty($creationClassName)) {
```
shadowhand: I would use `$creationClass` instead of just the name, as it will be operated with below.

src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +194,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            try {
+                $creationClass->postMigrationCreation($filePath, $className, $this->getConfig()->getMigrationBaseClassName());
+            } catch(\Exception $ex) {
```
shadowhand: PSR style violation here... should be `catch (` with a space between

src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +194,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                throw new \RuntimeException(sprintf(
+                    'Problem calling %s->postMigrationCreation(), resulting in %s',
```
shadowhand: Not sure if this is worth a try/catch block... it might actually be better to just allow the exception to bubble, so that debugging is more clear.

src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +194,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
         $output->writeln('<info>using migration base class</info> ' . $classes['$useClassName']);
+
+        if (!empty($altTemplate)) {
+            $output->writeln('<info>using alternative template</info> ' . realpath($altTemplate));
+        } elseif (!empty($creationClassName)) {
+            $output->writeln('<info>using template creation class</info> ' . $creationClassName);
+        } else {
+            $output->writeln('<info>using template</info> ' . realpath(dirname(__FILE__) . self::DEFAULT_MIGRATION_TEMPLATE));
```
shadowhand: `dirname(__FILE__)` is the same as `__DIR__`.

src/Phinx/Migration/CreationInterface.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Phinx
+ *
+ * (The MIT license)
+ * Copyright (c) 2014 Rob Morgan
```
shadowhand: It's 2015 :)

robmorgan: :+1:

src/Phinx/Console/Command/Create.php
RQuadling: Code completion type hinting. IDEs get a real kick out of this.

src/Phinx/Console/Command/AbstractCommand.php
RQuadling: Yep, but only for PHP V5.3+.

src/Phinx/Console/Command/Create.php
RQuadling: Just hardcoding the name then? ... ok.

RQuadling: Hmm.. http://php.net/manual/en/language.oop5.constants.php#114143 self:: vs static::, I assume that no matter what extension to the Create class exists, the value assigned to Create::CREATION_INTERFACE will always be used. Of course, hardcoding it is an option.

src/Phinx/Console/Command/Create.php
RQuadling: The template may be synthetic and not physical and that is why we don't use altTemplate. The CreationInterface IS about the creation of the template for the migration, not just the supplying of a potentially non-existent template filename.

src/Phinx/Console/Command/Create.php
RQuadling: Is suppose the issue is do we want to allow a Create subclass to be able to dictate the interface to be used, that interface would have to extend CreationInterface and the subclass would have to provide its own checks to make sure that the class was implementing the appropriate interface. In that case self is still the appropriate choice as as far as the Create class is concerned, it is going to call a method defined in CreationInterface. Hmm. I don't think I'm explaining myself very well. Basically I think self:: is fine, even if a subclass does implement its own value for the constant - it won't be overloaded.

RQuadling:
OK. I think once I've pushed the feature branch to github, rebasing it locally means I have to throw away the remote branch and replace it with the rebased one, but that seems to break the PR. Am I missing something?

src/Phinx/Console/Command/Create.php
shadowhand: Yes, but `@var` declarations don't go inside function blocks, they got before class variables.

src/Phinx/Console/Command/Create.php
```diff
@@ -106,15 +123,57 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if (!empty($altTemplate)) {
+            // Get the template from the alternative template filename.
+            $contents = file_get_contents($altTemplate);
+        } elseif (!empty($creationClassName)) {
+            // Get the template from the creation class
+            $creationClass = new $creationClassName();
+            $contents = $creationClass->getMigrationTemplate();
+        } else {
+            // load the migration template
+            $contents = $this->getMigrationTemplate();
```
shadowhand: I meant this call to `getMigrationTemplate`, which should really just return a path. If someone wants a synthetic template, instead of a file based on, they could use `CreationClass`.

src/Phinx/Console/Command/Create.php
shadowhand: See comment above.

shadowhand:
> have to throw away the remote branch and replace it with the rebased one

Not necessary to throw it away, use `git push -f` to force the update. Since you are working on a private branch, force push is totally acceptable workflow.

src/Phinx/Console/Command/Create.php
RQuadling: Which is pretty rubbish really. You should be able to docblock vars in any scope. PHPStorm is happy with this, but I'll remove it

src/Phinx/Console/Command/Create.php
RQuadling: Aha! That's the original code. Will need to adjust the AbstractCommand to supply a filename rather than the template. OK.

RQuadling: I'm unsure why the template name was moved to AbstractCommand as nothing other than Create will use it.

RQuadling: Actually, no need for the AbstractCommand:;getMigrationTemplate() any more it would seem?

RQuadling:
Can you reopen this PR then?

shadowhand:
@RQuadling it could have been reopened before your branch was updated, but now the option is disabled:

screen shot 2015-01-14 at 14 50 08

RQuadling:
Argh!!!

Another PR??

shadowhand:
Looks like. :(

RQuadling:
OK. Hopefully the last time.

If I want to rebase, I just manually to a ```git push -f``` to refresh the pre-existing branch, overwriting the history?

I use SourceTree and there's no UI for push -f. Lots of people complaining about it.

RQuadling:
Replaced by https://github.com/robmorgan/phinx/pull/449

shadowhand:
> If I want to rebase, I just manually to a ```git push -f``` to refresh the pre-existing branch

Correct.

PR #445, opened 2015-01-12
Docs: import the namespace where MysqlAdapter is

maxgalbu:
(otherwise one would need to dig through the code to find where MysqlAdapter is defined)


shadowhand:
Makes sense, thanks!

PR #449, opened 2015-01-15

RQuadling:
Hopefully, the final time I'm going to be making a pull request for this feature. So so sorry.

This PR replaces PR #444.

Rebased to 0.4.x-dev

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -41,7 +41,7 @@ class Create extends AbstractCommand
     /**
      * The name of the interface that any external template creation class is required to implement.
      */
-    const CREATION_INTERFACE = '\Phinx\Migration\CreationInterface';
+    const CREATION_INTERFACE = 'Phinx\Migration\CreationInterface';
 
     /**
      * {@inheritdoc}
@@ -123,7 +123,7 @@ class Create extends AbstractCommand
         $fileName = Util::mapClassNameToFileName($className);
         $filePath = $path . DIRECTORY_SEPARATOR . $fileName;
 
-        if (file_exists($filePath)) {
+        if (is_file($filePath)) {
             throw new \InvalidArgumentException(sprintf(
                 'The file "%s" already exists',
                 $filePath
@@ -133,7 +133,7 @@ class Create extends AbstractCommand
         // Get the alternative template and static class options, but only allow one of them.
         $altTemplate = $input->getOption('template');
         $creationClassName = $input->getOption('class');
-        if (!empty($altTemplate) && !empty($creationClassName)) {
+        if ($altTemplate && $creationClassName) {
             throw new \InvalidArgumentException('Cannot use --template and --class');
         }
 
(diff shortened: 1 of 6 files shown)


src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +183,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
         $output->writeln('<info>using migration base class</info> ' . $classes['$useClassName']);
+
+        if (!empty($altTemplate)) {
+            $output->writeln('<info>using alternative template</info> ' . realpath($altTemplate));
+        } elseif (!empty($creationClassName)) {
+            $output->writeln('<info>using template creation class</info> ' . $creationClassName);
+        } else {
+            $output->writeln('<info>using template</info> ' . realpath(dirname(__FILE__) . self::DEFAULT_MIGRATION_TEMPLATE));
```
shadowhand: I don't think this is necessary, it doesn't exist in the current code.

src/Phinx/Console/Command/Create.php
```diff
@@ -131,7 +183,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
+        // Do we need to do the post creation call to the creation class?
+        if ($creationClassName) {
+            $creationClass->postMigrationCreation($filePath, $className, $this->getConfig()->getMigrationBaseClassName());
+        }
+
         $output->writeln('<info>using migration base class</info> ' . $classes['$useClassName']);
+
+        if (!empty($altTemplate)) {
+            $output->writeln('<info>using alternative template</info> ' . realpath($altTemplate));
```
shadowhand: I wouldn't use `realpath` here, as it is not really necessary to show the full path. Reflecting the option the user entered should be enough.

src/Phinx/Migration/CreationInterface.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Phinx
+ *
+ * (The MIT license)
+ * Copyright (c) 2014 Rob Morgan
```
shadowhand: 2015 :)

src/Phinx/Console/Command/Create.php
```diff
@@ -106,15 +119,54 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
             ));
         }
 
-        // load the migration template
-        $contents = $this->getMigrationTemplate();
+        // Get the alternative template and static class options, but only allow one of them.
+        $altTemplate = $input->getOption('template');
+        $creationClassName = $input->getOption('class');
+        if ($altTemplate && $creationClassName) {
+            throw new \InvalidArgumentException('Cannot use --template and --class');
```
shadowhand: The wording here could be improved, it should clarify that the two cannot be used together:

> Cannot use --template and --class at the same time.

shadowhand:
Looking good! Most things requested from last time appear to be fixed. :)

src/Phinx/Console/Command/Create.php
RQuadling: Ha ha! Yes. " at the same time" really does make all the difference!

src/Phinx/Migration/CreationInterface.php
RQuadling: You need to do that in a LOT of places!!! There are 30 "2014 Rob Morgan"s in the source. :-)

src/Phinx/Console/Command/Create.php
RQuadling: I've added this as a way to report back exactly what was chosen. I would prefer to keep it.

src/Phinx/Migration/CreationInterface.php
shadowhand: New files get new date, other files get date updated as they are touched. No need to global search and replace. :)

RQuadling:
Have I got anything left to get this PR approved/merged?

RQuadling:
Thank you.

PR #451, opened 2015-01-15
ability to rollback an individual migration without migrations that have...

automatical:
... happened after it

There are some situations where migrations may happen out of order of they they are written, this change allows single migrations to be reverted on an individual basis.

### DIFF
--- a/src/Phinx/Console/Command/Rollback.php
+++ b/src/Phinx/Console/Command/Rollback.php
@@ -45,13 +45,17 @@ class Rollback extends AbstractCommand
 
         $this->setName('rollback')
              ->setDescription('Rollback the last or to a specific migration')
+             ->addOption('--individual', '-i', InputOption::VALUE_OPTIONAL, 'An individual migration to roll back')
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to rollback to')
              ->setHelp(
 <<<EOT
-The <info>rollback</info> command reverts the last migration, or optionally up to a specific version
+The <info>rollback</info> command reverts the last migration, or optionally up to a specific version.
+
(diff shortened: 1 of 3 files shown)


temperedvision:
I would also like to see this kind of feature being added, especially for large teams with different parts of a project, where linear based may only made sense up to a point (especially when testing out individual features etc)

robmorgan:
I'm waiting to see if theres further community support before merging

jvengeon:
I would like also migrate or rollback a specific version.

Timethor:
This would be a great feature addition

fabianhoward:
This would be really helpful. When merging in development branches of git it often happens that the migrations created are dated before some of the ones already deployed.

rquadling:
I would find this feature a bit of a problem for views/stored procedures/etc that are dependent upon a specific column existing that can be removed by randomly rolling back a migration.

Rolling back should by LIFO (Last In, First Out).

rquadling:
@fabianhoward, We use a mechanism to rename the migrations in a branch such that they are pushed as close to now as possible rather than when they were created. This is done prior to feature finishing. So, as far as production is concerned, these migrations have just been developed. As each feature branch is closed, its migrations are the most recent ones to add. It works for us (4 devs in 2 countries with 3 different applications and - at the moment - 18 different features being developed across the set). Of course, YMMV.

voidcontext:
+1 for this feature, it would be really helpful.

shadowhand:
@automatical I really like the idea of this change, but it can no longer be merged into the 0.4.x branch, as we are now targeting 0.5.x. Please open a new PR against the 0.5.x-dev branch. Thank you!

automatical:
@shadowhand Sure, I'll get this to you at some point during the holiday break.

PR #452, opened 2015-01-16
Multi-database support + added ENUM and SET for MySQL only

bartekrylko:
Added support for multi-databases migrations.
Added also support for ENUM and SET MySQL types. Values should be defined in 'limit' field.
More about using multi-db can be found here: https://github.com/DatingVIP/phinx/tree/0.4.x-dev


.gitignore
```diff
@@ -32,3 +32,4 @@ phpunit.xml
 
 # sqlite test database
 phinx_testing.sqlite3
+/nbproject/
```
shadowhand: These kinds of user-specific ignores should be done via [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) not at the repository level.

README.md
```diff
@@ -31,6 +31,14 @@ Phinx natively supports the following database adapters:
 * SQLite
 * Microsoft SQL Server
 
+#### Features added in this fork
```
shadowhand: These notes should be added to the documentation files, not the README.

src/Phinx/Config/Config.php
```diff
@@ -143,7 +143,7 @@ public function getEnvironments()
     /**
      * {@inheritdoc}
      */
-    public function getEnvironment($name)
+    public function getEnvironment($name, $database = null)
```
shadowhand: These changes for multi-database should be a separate PR from adding new column types, as the two are completely unrelated.

shadowhand:
The file mode changes `100644 → 100755` should be reverted... Only executable files should have the executable bit set.

src/Phinx/Db/Adapter/AdapterWrapper.php
```diff
@@ -227,7 +227,7 @@ public function isValidColumnType(Column $column)
      */
     public function hasTransactions()
     {
-        return $this->getAdapter()->hasTransaction();
+        return $this->getAdapter()->hasTransactions();
```
shadowhand: Good catch, but this should have been a separate PR. I'll make the change to master.

src/Phinx/Db/Table/Column.php
```diff
@@ -151,7 +151,7 @@ public function getType()
     /**
      * Sets the column limit.
      *
-     * @param integer $limit
+     * @param mixed $limit integer|string - can be string for SET and ENUM
```
shadowhand: This is pretty weird to me, I don't think using `limit` for ENUM options makes sense... and shouldn't ENUM options be an array, not a string?

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -871,4 +871,32 @@ public function testHasColumnReservedName()
         $this->assertTrue($tableQuoted->hasColumn('value'));
 
     }
+
+    public function testGetColumnTypes()
```
shadowhand: If the intent of this test is simply to make sure that `enum` and `set` are in the column types, that should be the only thing tested. As currently written, this test is fragile because adding new base types will cause it to fail.

src/Phinx/Db/Table/Column.php
bartekrylko: What do you suggest here? New column option called 'values' for example?
That option could take both, string and array.

shadowhand: Yes, a new column option. Calling it "values" makes sense to me.

bartekrylko:
Splitted into two separate: #464 #464 with suggested changes.

PR #454, opened 2015-01-17
(c) 2014 => (c) 2015

RQuadling:
And 2 Docblock marker typos.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -3,7 +3,7 @@
  * Phinx
  *
  * (The MIT license)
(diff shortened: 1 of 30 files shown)


robmorgan:
thanks @RQuadling !

shadowhand:
I don't necessarily agree with this change. Updates to copyright should be done a per-file basis as the files are editing, not global search and replace... there's nothing wrong with old copyright dates. This kind of change just creates noise in the log history.

RQuadling:
OK. PHP tends to update all the (C) each year. After being told twice, ... hopefully I'll remember next year. :-D

PR #455, opened 2015-01-17
Added jsonb support for postgresql 9.4

Garbee:
PostgreSQL 9.4 added [jsonb](http://www.postgresql.org/docs/9.4/static/datatype-json.html) type support. This allows for json stores to be indexed by the engine.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -690,6 +690,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
             case static::PHINX_TYPE_DATE:
             case static::PHINX_TYPE_BOOLEAN:
             case static::PHINX_TYPE_JSON:
+            case static::PHINX_TYPE_JSONB:
             case static::PHINX_TYPE_UUID:
                 return array('name' => $type);
             case static::PHINX_TYPE_STRING:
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -747,6 +748,7 @@ public function getPhinxType($sqlType)
                 return static::PHINX_TYPE_CHAR;
             case 'text':
             case 'json':
+            case 'jsonb':
```
robmorgan: shouldn't json + jsonb return the Phinx JSON types here?

Garbee: I was thinking so myself, but I just followed what was being done with the json type before. I'll update my branch and see what the CI tests show.

robmorgan:
@Garbee yeah I could see that, I'm not sure why it was mapped to text in the first place though.

Garbee:
Looking over the file history using blame you made the edit with [this commit](https://github.com/robmorgan/phinx/commit/36516f1304e42da47055150dad27b74f35d7fdbb). It claims it was to fix a PR, but no detail on what exactly was broken.

Maybe you just forgot to add the return statement within that commit?

Garbee:
My only thing with this now is... Does adding the right return for the original json type break BC? Seems like it would in which case this should go to master vs the 0.4 branch.

Garbee:
Just got through moving my migrations into phinx and currently my patch isn't working. So before this gets pulled in I need to do some digging around and figure out what else needs to be changed.

If anyone reading this has any pointers on what would cause:
>An invalid column type was specified

being thrown with this patch applied then please let me know.

Garbee:
Any update on what should be done for this to get merged in?

robmorgan:
@Garbee I can take this now. I was worried it might break some BC case I haven't tested for but I'll play with it a bit.

robmorgan:
UUID has also been implemented for both MySQL & Postgres, so I'll remove some of the code.

Issue #172, opened 2014-01-04
Suggestion to add a seed()


CMCDragonkai:
A suggestion to add a seed() function that is defined inside any migration classes. If it exists, Phinx will run the seed() only after it runs up().

I often need this to seed tables with initial data, which goes along with migration.

alexbowers-tecmark:
That seems suitable to me.

gmauricio:
is there someone working on this? @todiadiyatmo, @robmorgan ? 
If not I'd be interested in doing it.

robmorgan:
I'm not working on this at the moment, but I'd happily accept a PR. Also I'd prefer we develop it with only one command. i.e: `phinx seed`

gmauricio:
Hey @robmorgan , please check #457, there I added an insert function to the adapter interface, also added it to Table. The idea is to implement seeder classes as in Laravel, and then they would have access to the same Table as in migrations.   Like this:

```
class DatabaseSeeder extends SeederInterface
{
    function run()
    {
        $user = $this->table('user');
        
        $data = array(
            array("email" => "foo@bar.com")
        );

        $user->insert($data)->save();
    }
}
```

So, in the PR, I already enabled that for Table. 

Now I'm not sure if we should separate the interfaces for the Table used in seeders, and the Table used in migrations, to avoid being able to for example do changes to the schema in a seeder, or maybe add data in a migration. My opinion is that there's no need to separate them, it would be up to the developer to use it correctly.

I'll implement the seed command, and the Seeder base class after getting #457 accepted.

alexbowers-tecmark:
How does this handle duplicate keys, and other exceptions. Are they all thrown to the user migrating and seeding, or are they just ignored?

Also, how is the wrong data type handled. Since MySQL (and I assume other drivers) would throw an exception; does this stop execution and clear the data that has already been inserted, or is it non-transitional.


----------------------

Alex Bowers
Backend Developer

tecmark
Manchester & London
Tel:  0845 680 1146
Fax:  0845 680 1316
Email:  alex.bowers@tecmark.co.uk<mailto:alex.bowers@tecmark.co.uk>
Web:  www.tecmark.co.uk<http://www.tecmark.co.uk/>

111 Piccadilly, Manchester M1 2HY<http://maps.google.com/maps?q=111%20Piccadilly%2C%20Manchester%20M1%202HY>

[...]

PR #457, opened 2015-01-19
Added insert data method for Table and AdapterInterface

gmauricio:
This is an attempt to build #172 by first adding the insert functionality to the Table class and Adapters. I think from here we can extend this to have a seed command, or a seed function inside migrations. 
I've been playing with that a bit also and I'll be happy to finish implementing it after getting feedback on this changes.

I wrote some tests, but I'm missing the SqlServerAdapter tests. I didn't have time to set that up.


gmauricio:
The insert method implemented in this PR, is supposed to receive a list of rows, instead of a single row. 

```
$this->table('table')->insert(array(
    array('col1', 'value1'),
    array('col1', 'value1')
));
```

Do you think, that maybe I should change it to receive a single row at a time ? Like this:

```
$table->insert(array('col1', 'value1'));
$table->insert(array('col2', 'value2'));
```

alexbowers:
I think having it allow either would be good.

src/Phinx/Db/Adapter/AdapterWrapper.php
```diff
@@ -122,7 +122,7 @@ public function setOutput(OutputInterface $output)
      */
     public function getOutput()
     {
-        return $this->adapter->getOutput($output);
+        return $this->adapter->getOutput();
```
shadowhand: Good catch!

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -330,6 +330,38 @@ public function fetchAll($sql)
...
+
+        foreach ($data as $row) {
+
+            $sql = sprintf(
+                "INSERT INTO %s (",
+                $this->quoteTableName($table->getName())
+            );
+
+            foreach (array_keys($row) as $columnName) {
+                $sql .= $this->quoteColumnName($columnName) . ', ';
```
shadowhand: How about:

```php
$sql .= implode(', ', array_map(array($this, 'quoteColumnName'), array_keys($row));
```

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -330,6 +330,38 @@ public function fetchAll($sql)
...
+            }
+            $sql = substr(rtrim($sql), 0, -1) . ")";
+
+            $sql .= " VALUES (";
+
+            foreach (array_values($row) as $value) {
+                $sql .= (is_numeric($value) ? $value : "'" . $value . "'") . ', ';
```
shadowhand: Same here:

```php
$sql .= implode(', ', array_map(function ($value) {
    if (is_numeric($value)) {
        return $value;
    }
    return "'{$value}'";
}, array_values($row)));
```

src/Phinx/Db/Table.php
```diff
@@ -541,6 +566,21 @@ public function addTimestamps()
...
+     *              array(
+     *                  array("column1" => "value1", "column2" => "anotherValue1"),
+     *                  array("column1" => "value2", "column2" => "anotherValue2"),
+     *              )
+     *
+     * @return Table
+     */
+    public function insert($data)
+    {
+        $this->data = $data;
```
shadowhand: This doesn't really make sense to me... INSERTs have 3 parts:

- a table name
- **one** set of columns
- one **or more** rows

I don't think it is good practice to force the insert data to contain the column names in every line.

gmauricio:
As a clarification, after some changes, the current implementation will enable inserts this way : 

```php
$this->table('table')->insert(
    array('col1', 'col2'),
    array(
        array('col1Value1', 'col2Value1'),
        array('col1Value2', 'col2Value2'),
    )
));
```

alexbowers:
Is this suitable for migrating in then at its current stage?

gmauricio:
Yes, at this stage you would be able to do inserts from migrations

alexbowers:
@shadowhand would you be able to migrate this in?

gmauricio:
Hey @shadowhand , @robmorgan just wondering why this hasn't been accepted.
Let me know if there's something to fix or improve, I'll be happy to do it in order to get this accepted. Also I have some more code in line ready to increment this solution and have a new seed command.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -330,6 +330,38 @@ public function fetchAll($sql)
...
+        $sql .= "(". implode(', ', array_map(array($this, 'quoteColumnName'), $columns)) . ")";
+        $sql .= " VALUES ";
+
+        $rows = array();
+        foreach ($data as $row) {
+            $rows[] = "(" . implode(', ', array_map(function ($value) {
+                if (is_numeric($value)) {
+                    return $value;
+                }
+                return "'{$value}'";
```
AD7six: There really shouldn't be any string mangling going on here - otherwise this is going to fail hard for any data that contains a quote. This should be based on a pdo statement.

robmorgan:
Hi @gmauricio, the seed feature will ship in 0.5.0, so its time to take this now.

robmorgan:
However I will change the method signature to the following:

```
$this->table('table')->insert(
    array('col1' => 'foo', 'col2' => 'bar'),
    array('col1' => 'foo2', 'col2' => 'bar2')
));
```

anuj9196:
In `CakePHP 3.5+`, I have to create few default users, so in `up()` function below table `create()`
```
$this->table('users')->insert([
    [
        'name' => 'Name1',
        'email' => 'email1@example.com',
        'password' => 'password1',
        'super_admin' => true,
        'master_admin' => true,
    ], [
         'name' => 'Name2',
         'email' => 'email2@example.com',
         'password' => 'password2',
         'super_admin' => true,
    ], [
        'name' => 'Name3',
        'email' => 'email3@example.com',
        'password' => 'password3',
        'super_admin' => true
   ]
]);
```

But this does not insert any record in `users` table.

alexbowers:
I think you need to call $this->saveData() at the end

PR #459, opened 2015-01-21
Add support for UUID in Mysql and Sqlite

dakota:
Currently only SqlServer and PostgreSql support native UUID/UniqueIdentifier column types. This PR adds a mapping for Mysql and Sqlite to treat UUIDs as CHAR(36).

This allows for migration scripts originally written for SqlServer or PostgreSql to work without modification on Mysql/Sqlite.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -787,6 +787,8 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             case static::PHINX_TYPE_BOOLEAN:
                 return array('name' => 'tinyint', 'limit' => 1);
                 break;
+            case static::PHINX_TYPE_UUID:
+                return array('name' => 'char', 'limit' => 36);
             // Geospatial database types
             case static::PHINX_TYPE_GEOMETRY:
             case static::PHINX_TYPE_POINT:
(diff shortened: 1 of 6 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -837,6 +839,9 @@ public function getPhinxType($sqlTypeDef)
                     if ($limit === 255) {
                         $limit = null;
                     }
+                    if ($limit === 36) {
+                        $type = static::PHINX_TYPE_UUID;
```
shadowhand: Not 100% positive this will work all the time, but let's give it a shot.

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1142,23 +1142,7 @@ protected function getForeignKeySqlDefinition(ForeignKey $foreignKey, $tableName
...
-            'datetime',
-            'timestamp',
-            'time',
-            'date',
-            'binary',
-            'boolean',
-            'uuid',
-            'filestream'
-        );
+        return array_merge(parent::getColumnTypes(), array('filestream'));
```
shadowhand: :+1:

lorenzo:
:clap:

src/Phinx/Db/Adapter/MysqlAdapter.php
dakota: There will probably be cases where this may not work, but there isn't really any other way with MySQL but to assume that CHAR(36) is UUID. At least not that I'm aware of.

PR #464, opened 2015-01-23
Added ENUM and SET types for MySQL

bartekrylko:
Added as saparate PR - based on #452

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -966,8 +966,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $def .= '(' . $sqlType['limit'] . ')';
         }
         if(($values = $column->getValues()) && is_array($values)) {
-            foreach($values as &$value) { $value = "'{$value}'"; }
-            $def .= '('.implode(',', $values).')';
+            $def .= "('" . implode("', '", $values) . "')";
         }
         $def .= (!$column->isSigned() && isset($this->signedColumnTypes[$column->getType()])) ? ' unsigned' : '' ;
         $def .= ($column->isNull() == false) ? ' NOT NULL' : ' NULL';
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -795,6 +795,11 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             case static::PHINX_TYPE_LINESTRING:
             case static::PHINX_TYPE_POLYGON:
                 return array('name' => $type);
+            case static::PHINX_TYPE_ENUM:
+                return array('name' => 'enum');
+                break;
+            case static::PHINX_TYPE_SET:
+                return array('name' => 'set');
                 break;
             default:
                 throw new \RuntimeException('The type: "' . $type . '" is not supported.');
@@ -960,6 +965,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';
         }
+        $def .= $column->getValues() ? '(' . $column->getValues() . ')' : '' ;
         $def .= (!$column->isSigned() && isset($this->signedColumnTypes[$column->getType()])) ? ' unsigned' : '' ;
         $def .= ($column->isNull() == false) ? ' NOT NULL' : ' NULL';
         $def .= ($column->isIdentity()) ? ' AUTO_INCREMENT' : '';
@@ -1051,5 +1057,18 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
         );
 
         return $this->fetchRow($sql);
+    }
+
+    /**
+     * Returns MySQL column types (inherited and MySQL specified).
+     * @return array
+     */
+    public function getColumnTypes()
+    {
+        return array_merge(parent::getColumnTypes(),
+            [
+                'enum',
+                'set',
+            ]);
     }
 }
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -965,7 +965,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';
         }
-        if(($values = $column->getValues()) && is_array($values)) {
+        if (($values = $column->getValues()) && is_array($values)) {
             $def .= "('" . implode("', '", $values) . "')";
         }
         $def .= (!$column->isSigned() && isset($this->signedColumnTypes[$column->getType()])) ? ' unsigned' : '' ;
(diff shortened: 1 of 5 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1052,4 +1058,17 @@ public function describeTable($tableName)
...
     }
+
+    /**
+     * Returns MySQL column types (inherited and MySQL specified).
+     * @return array
+     */
+    public function getColumnTypes()
+    {
+        return array_merge(parent::getColumnTypes(),
+            [
```
shadowhand: This needs to be PHP 5.3 compatible, not using short array syntax.

src/Phinx/Db/Table/Column.php
```diff
@@ -455,6 +460,34 @@ public function getProperties()
...
+     *
+     * @param mixed (array|string) $values
+     *
+     * @return Column
+     */
+    public function setValues($values)
+    {
+        if(is_array($values))
+        {
+            foreach($values as &$value) { $value = "'{$value}'"; }
```
shadowhand: Quoting should not be done on set, but on write.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -576,7 +576,9 @@ public function testGetColumns()
               ->addColumn('column17', 'point')
               ->addColumn('column18', 'linestring')
               ->addColumn('column19', 'polygon')
-              ->addColumn('column20', 'uuid');
+              ->addColumn('column20', 'uuid')
+              ->addColumn('column21', 'set', array('values' => "'one', 'two'"))
+              ->addColumn('column22', 'enum', array('values' => ['three', 'four']));
```
shadowhand: short array needs to be long array

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -621,7 +623,9 @@ public function testGetColumnsReservedTableName()
...
```
shadowhand: short array

bartekrylko:
Please take a look on that changes.

alexbowers:
Any update on enum support?

shadowhand:
Sorry I started a new job and didn't have time to go through PRs. Am reviewing now.

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -960,6 +965,10 @@ protected function getColumnSqlDefinition(Column $column)
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';
         }
+        if(($values = $column->getValues()) && is_array($values)) {
+            foreach($values as &$value) { $value = "'{$value}'"; }
+            $def .= '('.implode(',', $values).')';
```
shadowhand: I would prefer these two lines be merged into one:

```php
$def .= "('" . implode("', '", $values) . "')";
```

src/Phinx/Db/Table/Column.php
```diff
@@ -455,6 +460,30 @@ public function getProperties()
...
     /**
+     * Sets field values.
+     *
+     * @param mixed (array|string) $values
+     *
+     * @return Column
+     */
+    public function setValues($values)
+    {
+        if(!is_array($values)) { $values = explode(',', $values); }
```
shadowhand: This needs to match PSR-2 standards:

```php
if (...) {
    ...
}
```

bartekrylko:
Updated. Please take a look.

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -960,6 +965,9 @@ protected function getColumnSqlDefinition(Column $column)
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';
         }
+        if(($values = $column->getValues()) && is_array($values)) {
```
shadowhand: Still slightly off for PSR compatibility, but close enough. (Need `if (` with a space.)

src/Phinx/Db/Table/Column.php
```diff
@@ -105,6 +105,11 @@ class Column
     protected $properties = array();
 
     /**
+     * @var string
```
shadowhand: type should be an array

src/Phinx/Db/Table/Column.php
```diff
@@ -455,6 +460,32 @@ public function getProperties()
...
+     * Sets field values.
+     *
+     * @param mixed (array|string) $values
+     *
+     * @return Column
+     */
+    public function setValues($values)
+    {
+        if(!is_array($values)) {
+            $values = explode(',', $values);
```
shadowhand: I would use `preg_split('/,\s*/', $values)` to account for `foo, bar, baz`

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -621,7 +623,9 @@ public function testGetColumnsReservedTableName()
               ->addColumn('column17', 'point')
               ->addColumn('column18', 'linestring')
               ->addColumn('column19', 'polygon')
-              ->addColumn('column20', 'uuid');
+              ->addColumn('column20', 'uuid')
+              ->addColumn('column21', 'set', array('values' => "one, two"))
```
shadowhand: this example will be slightly broken because of the issue with spaces in a comma separated list

shadowhand:
I think the comma-separated list problem should be fixed before merge. Otherwise everything looks good.

bartekrylko:
I think, current approach is not a bug. MySQL allows for ENUM/SET values with white space like ' one'. Why should we "fix" such cases?

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -960,6 +965,9 @@ protected function getColumnSqlDefinition(Column $column)
...
+            $def .= "('" . implode("', '", $values) . "')";
```
alexbowers: As per @shadowhand comment below about using preg_split

shadowhand:
> MySQL allows for ENUM/SET values with white space like ' one'. 

That's absolutely true, but I think the expectation for a string list of values `"a, b, c"` would be that it produce `"a", "b", "c"` and not `"a", " b", " c"`.

bartekrylko:
I'm still not 100% sure about that change in Column class. But updated as you suggested. Hopefully, expectations are as you mentioned.

shadowhand:
Thanks!

bartekrylko:
thanks

alexbowers:
Any chance of a documentation PR @bartekrylko

bartekrylko:
Sure, here you go. Two simple examples:
```
->addColumn('enum_col_name', 'enum', array('values' => "one, two", 'default' => 'one'))
->addColumn('set_col_name', 'set', array('values' => array('one','two'), 'null' => true))
```
For both types (ENUM and SET), 'values' option supports string and array.

alexbowers:
The string being comma separated with or without spaces?

Is this feature MySQL only. Or supported in all drivers?

alexbowers:
#474 Has provided proper documentation for it.

bartekrylko:
This is supported for MySQL only.
> The string being comma separated with or without spaces? @alexbowers 

Doesn't matter. Value will be fetched without spaces anyway.

Issue #466, opened 2015-01-25
init enhancement : specify config type


JeromeGill:
Init should allow specifying php, yaml or json config

Each have sufficient merit in different projects to warrant equal exposure as supported formats right from the get-go.

Would you accept a PR for this?

shadowhand:
I like the idea.

dereuromark:
Are you able to provide a pull request with your suggested changes?

dereuromark:
ping

otherwise we have to close here.

MasterOdin:
@dereuromark I can do this.

Looking at the init command, I saw:
https://github.com/cakephp/phinx/blob/0429021232229a9ea8da24b064b70c002dcae5f0/src/Phinx/Console/Command/Init.php#L138-L143

Could this be simplified to always look in one directory now for templates for the three different formats (say `src/Phinx/data` or `src/data`) and then stick the templates in there? Or just make all three of them at the root directory (though I'd probably remain the existing `phinx.yml` to `phinx.yml.dist` and then have `phinx.php.dist` and `phinx.json.dist` as well)?

And here's my proposed change to the command:
```
bin/phinx init --help
Usage:
  init [options] [--] [<path>]

Arguments:
  path                  Which path should we initialize for Phinx?

Options:
  -f, --format=FORMAT   What format should we use to initialize? [default: "yml"]
```

Issue #55, opened 2013-02-19
Interactive init


andho:
Allow to specify the migrations dir and other configurations.
Also should create the migration dir if it does not exist.

robmorgan:
was implemented in #468.

PR #468, opened 2015-01-26
interactive setup

JeromeGill:
Implements #55 and #466

On running init, the user is asked to specify the path (defaulting to cwd as before) followed by selecting one of the three supported config formats.

### DIFF
--- a/phinx.json
+++ b/phinx.json
@@ -1,0 +1,36 @@
+{
+  "paths": {
+    "migrations": "%%PHINX_CONFIG_DIR%%/migrations"
+  },
+  "environments": {
+    "default_migration_table": "phinxlog",
+    "default_database": "development",
+    "production": {
+      "adapter": "mysql",
+      "host": "localhost",
+      "name": "production_db",
+      "user": "root",
+      "pass": "",
+      "port": 3306,
+      "charset": "utf8"
+    },
+    "development": {
+      "adapter": "mysql",
+      "host": "localhost",
+      "name": "development_db",
+      "user": "root",
+      "pass": "",
+      "port": 3306,
+      "charset": "utf8"
+    },
+    "testing": {
+      "adapter": "mysql",
+      "host": "localhost",
+      "name": "testing_db",
+      "user": "root",
+      "pass": "",
+      "port": 3306,
+      "charset": "utf8"
+    }
+  }
+}
--- a/src/Phinx/Console/Command/Init.php
+++ b/src/Phinx/Console/Command/Init.php
@@ -29,8 +29,9 @@
 namespace Phinx\Console\Command;
 
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class Init extends Command
@@ -42,7 +43,6 @@ class Init extends Command
     {
         $this->setName('init')
             ->setDescription('Initialize the application for Phinx')
-            ->addArgument('path', InputArgument::OPTIONAL, 'Which path should we initialize for Phinx?')
             ->setHelp(sprintf(
                 '%sInitializes the application for Phinx%s',
                 PHP_EOL,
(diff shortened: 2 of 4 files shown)


src/Phinx/Console/Command/Init.php
```diff
@@ -77,8 +77,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ));
         }
 
+        // set the config format
+        $getFormat = new ChoiceQuestion(
+            'Would you like Phinx to use json yml or php based configuration?',
```
shadowhand: Grammar would be better as:

> Would you like to use JSON, YML, or PHP configuration files?

src/Phinx/Console/Command/Init.php
```diff
@@ -77,8 +77,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $type = $helper->ask($input, $output, $getFormat);
+
+        switch ($type) {
+            case 'php':
+            case 'json':
+            case 'yml':
+                break;
+            default:
+                throw new \InvalidArgumentException(sprintf(
+                    'The format "%s" is not one of json, php or yml',
```
shadowhand: No need for this check, ChoiceQuestion already enforces it.

phinx.json
```diff
@@ -0,0 +1,36 @@
+{
+  "paths": {
+    "migrations": "%%PHINX_CONFIG_DIR%%/migrations"
```
shadowhand: Rather than creating duplicated configuration files, wouldn't it be better to ship with a PHP file, and then translate it to YML and JSON at run time?

src/Phinx/Console/Command/Init.php
JeromeGill: Yeah your right, I wasn't using ChoiceQuestion at first but forgot to get rid of the check.

src/Phinx/Console/Command/Init.php
JeromeGill: Ok no problem.

phinx.json
JeromeGill: I did think about at that but came to the conclusion that I wasn't going to interactively generate the DB config as there is no advantage to it over just editing the provided file. 

The advantage of leaving them in place is they serve as templates for the layout of the various formats for people not using the init command, or with different formats they want to change at a later date. Generating them seemed like overkill for little advantage to me, you may disagree?

shadowhand: I don't mean interactively generate it, I mean transmute the default PHP config into each type on demand. For instance, if I request YAML config:

- load the PHP config
- convert to YAML
- write config file

Repeat the same process for JSON. This prevents the YAML/JSON from getting out of sync with PHP.

JeromeGill: Yeah I would prefer that too. 

Hmm, well doing that with JSON is no issue at all with json_encode. Converting to yaml, however requires either the yaml php extension, writing a yaml generator or introducing one as a dependency. Would it be worth introducing these dependencies/complexities over simply ensuring all three are unit tested in the appropriate place?

JeromeGill: On second thought, users wishing to use yaml will likely have the extension anyway. Ok. I'll go ahead with this.

shadowhand: Phinx currently [requires YAML](https://github.com/robmorgan/phinx/blob/cd2009b71e16ca8363e5e95ee19a991ae4cfcb5a/composer.json#L23). Regardless, you're right that anyone wanting to use YAML would have the extension installed.

JeromeGill:
@shadowhand Updated PR with your suggested changes

mdwheele:
Is there still interest in implementing the changes suggested in this PR? I was considering a less-interactive implementation that exposed command-line options rather than prompts, but the value provided is the same. I think it's a worthwhile addition and would be interested in continuing the work in this PR or starting a new branch that accomplishes this via command-line options.

robmorgan:
@mdwheele not at the moment, the default yaml template is still good enough for most users.

mdwheele:
Thanks @robmorgan. There was work done in another PR (forget the number) that actually helps resolve some of *my* specific reasoning for this type of feature by adding `phinx.php` configuration to the scan path. At that point, the last 20% is just creating the file and I can do that myself. 

I appreciate the hard work on this project and use it every day. Looking forward to using the Seeder :smile:.

robmorgan:
@mdwheele thanks mate, I hope you enjoy it!

robmorgan:
closing for now.

PR #469, opened 2015-01-29
Mysql binary should use LONGBLOB

dakota:
`BLOB` fields only supports 2^16 (65536) bytes, whereas `LONGBLOB` supports 2^32 bytes (4096 MiB).


shadowhand:
Not sure I agree with this. If you want longer blob support, it should be implemented like `LONGTEXT`, not as a new column type.

dereuromark:
Refs https://github.com/robmorgan/phinx/issues/45

mamchenkov:
Here's the PR #555 to that does the BLOB sizes.

mamchenkov:
Now that #555 was merged in, this can be closed I guess.

PR #470, opened 2015-02-01
Make quotes more ANSI compatible.

### LINKED ISSUE #453: 1054 Unknown column '20150116211834' in 'field list'
Running an empty migration as detailed in the install modal results in an error:

```
php vendor/bin/phinx create MyFirstMigration
php vendor/bin/phinx migrate -e development
Phinx by Rob Morgan - https://phinx.org. version 0.4.1

using config file ./phinx.yml
using config parser yaml
using migration path /home/mark/Projects/accounts/migrations
using environment development
using adapter mysql
using database metrodb_test

 == 20150116211834 MyFirstMigration: migrating


 [PDOException]                                                                           
  SQLSTATE[42S22]: Column not found: 1054 Unknown column '20150116211834' in 'field list'
```

Removing this migration, creating a new one, adding some commands to rename a column results in the same error (different column name).

markkimsal:
Using double quotes for values is MySQL specific.  Running against
a MySQL server with ANSI_QUOTES mode set results in an inability to
record migrations in the database as the migration name value is
treated as a column name.

This change makes the use of single quotes more consistent with all
existing database drivers and more compatible with ANSI usage.

This is more of a bug fix than a feature and will close #453

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -352,9 +352,9 @@ abstract class PdoAdapter implements AdapterInterface
                 'INSERT INTO %s ('
                 . 'version, start_time, end_time'
                 . ') VALUES ('
-                . '"%s",'
-                . '"%s",'
-                . '"%s"'
+                . '\'%s\','
+                . '\'%s\','
+                . '\'%s\''
                 . ');',
                 $this->getSchemaTableName(),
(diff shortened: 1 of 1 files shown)


shadowhand:
Are there other places where double quotes are used?

markkimsal:
I don't think so.  This change makes phinx functional on my setup, and I just did a quick scan of double quotes in all the adapters.

PR #472, opened 2015-02-05
Allow aliases for the template creation class names.

rquadling:
We have a set of migration template creation classes that are namespaced. Allowing them to be accessed via an alias which is part of our Phinx setup simplifies things for our developers.

One thing I wasn't sure about is making ```Config::getAlias()``` part of ```ConfigInterface```.

If I did, any pre-existing class implementing that interface will now require an empty function, assuming they didn't want to use any aliases.

So, hopefully, just where it is is OK.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1,6 +1,6 @@
 .. index::
    single: Configuration
-   
+
 Configuration
 =============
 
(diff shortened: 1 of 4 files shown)


RQuadling:
Is there anything to be done about AppVeyor failing?

rquadling:
Any chance this could be merged please?

rquadling:
Thank you.

robmorgan:
:+1:

PR #474, opened 2015-02-05
Update documentation to show enum and set

alexbowers:


docs/migrations.rst
```diff
@@ -413,6 +413,8 @@ Column types are specified as strings and can be one of:
 -  date
 -  binary
 -  boolean
+-  enum
```
shadowhand: Like the Postgres note, I think these two should go in a separate MySQL section.

alexbowers:
@shadowhand Updated

shadowhand:
Thanks!

PR #477, opened 2015-02-06
Add port information when using sqlsrv adapter

tuimz:
Hi,

When using sqlsrv to connect to a named instance it had me searching for quite a while to find out that if you connect to an instance it will negotiate which port to use. So you need to omit the port in the configuration.


shadowhand:
Thanks!

Issue #480, opened 2015-02-11
Postgres: Syntax Error


daniel-lang:
I noticed a couple of problems with the changeColumn method of the PostgresAdapter:
* Changing the default doesn't work since this has to be done with a separate `ALTER TABLE` just like changing the NULL constraint
* Replacing the `NULL` and `NOT NULL` with an empty string in the getColumnSqlDefinition function doesn't work when specifying the DEFAULT value since it is not the end of the query

m1ome:
:+1:

shadowhand:
Fixed by #481

PR #481, opened 2015-02-11
Postgres: fixed problem with changeColumn and default values

daniel-lang:
When trying to change the default value in the changeColumn method a syntax error occurred.
Added the ability to drop the default constraint by leaving default null.


daniel-lang:
fixed #480

shadowhand:
Thanks!

Issue #482, opened 2015-02-11
Support for BIGSERIAL and JSONB in PGSQL Adaptor


jwoodcock:
Right now I've found no support for BIGSERIAL nor JSONB in the PGSQL adaptor. 

I've tested out a fix locally and will be submitting a PR with it once I make sure to have not messed up test coverage.

tacman:
Can you post your fix or your fork of the repo?  The ability to use jsonb  would be enormously helpful.

jwoodcock:
@tacman You can see all the code changes here:
https://github.com/robmorgan/phinx/pull/483

I just ran out of time and couldn't make the changes they asked. In addition, we've ran into a number of other limitations that has forced us to use only the exec/query commands.

robmorgan:
Closing this for now, please open a separate issue for `BIGSERIAL` and I'll merge if I see further support from the community.

PR #483, opened 2015-02-11
Postgres Adaptor Support for JSONB and BIGSERIAL

jwoodcock:
This PR fixes issue #482 with test and documentation. 

Recap: Postgres adaptor does not current support JSONB, Postgres 9.4, nor BIGSERIAL, ~8.1. This PR fixes those.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -752,6 +752,8 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
                 return static::PHINX_TYPE_JSONB;
             case 'bigserial':
                 return static::PHINX_TYPE_BIGSERIAL;
+            case 'inet':
+                return static::PHINX_TYPE_INET;
             case 'int':
             case 'int4':
             case 'integer':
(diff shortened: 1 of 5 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -748,6 +751,12 @@ public function getPhinxType($sqlType)
             case 'text':
             case 'json':
                 return static::PHINX_TYPE_TEXT;
+            case 'jsonb':
+                return static::PHINX_TYPE_JSONB;
```
shadowhand: Is there any way to combine JSON and JSONB together via column options?

jwoodcock: Would you mind elaborating on your request? JSON and JSONB are two distinct column types in Postgres 9.4 create statements.

shadowhand: Right, but the only difference is in storage, not in structure. Ideally, both would share a common `JSON` type and have a column flag for "is this a binary column?". Similar to how INT columns have an "identity" flag.

robmorgan:
This PR needs to be broken up as Phinx now supports JSONB for Postgres.

robmorgan:
I'm closing for now.

PR #485, opened 2015-02-12
Bugfix: dropForeignKey SQLite adapter

boedy:
Removing a foreign key with SQLite adapter caused the table to be removed. Also the create syntax left out the original column.


shadowhand:
Thank you!

Issue #429, opened 2015-01-02
AppVeyor builds are failing for no apparant reason


shadowhand:
https://ci.appveyor.com/project/robmorgan/phinx/build/1.0.388

![screen shot 2015-01-01 at 21 33 41](https://cloud.githubusercontent.com/assets/38203/5594073/edcc8dc4-91fd-11e4-9044-713b7831f0a6.png)

Tests are passing, but fail with "The batch file cannot be found".

robmorgan:
ahh now I think I know what broke this.

alexbowers:
@robmorgan Any update on getting this fixed? Quite a few pull-requests are failing due to a build error

robmorgan:
Appveyor is not running the tests at all right now. Looking into it...

robmorgan:
I've narrowed the batch file cannot be found error down to `InitTest.php`.

PR #486, opened 2015-02-13
Added support for parameterised queries.

courtney-miles:
Added support for parameterised/prepared queries, based on the work done at #256.

Note that I have introduced `test/lib/PdoMockable.php` so I can test the `PdoAdapter` with a mock `PDO` object.  I'm not sure if that's where you would want such code.

Also, this change includes PHPUnit tests the first assertions for the `query()` method return value.  Despite what the documentation says, PHPDocs say, and the mock implementations of `query()`, it does not return an array.  It isntead returns a `PDOStatement` so that's what I'm testing on.


courtney-miles:
... will fix the Travis CI build failures as soon as I can.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -292,17 +292,68 @@ public function disconnect()
...
+     * @return \PDOStatement
+     */
+    protected function executeParameterisedQuery($sql, array $inputParams)
+    {
+        $statement = $this->getConnection()->prepare($sql);
+
+        $position = 1;
+        foreach ($inputParams as $key => $bind) {
+            $value = $bind;
+            $type = \PDO::PARAM_STR;    // This is PDO's default type already.
```
eridal: how about if you leave pdo to figure out the default `$type` ... it does a really decent job

```php
$type = null;
...
if ($type) {
  $statement->bindValue($parameter, $value, $type);
} else {
  $statement->bindValue($parameter, $value, $type);
}
```

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -44,6 +44,12 @@
  */
 abstract class AbstractMigration implements MigrationInterface
 {
+    const PHINX_PARAM_BOOL = \PDO::PARAM_BOOL;
+    const PHINX_PARAM_NULL = \PDO::PARAM_NULL;
+    const PHINX_PARAM_INT = \PDO::PARAM_INT;
+    const PHINX_PARAM_STR = \PDO::PARAM_STR;
+    const PHINX_PARAM_LOB = \PDO::PARAM_LOB;
```
eridal: maybe it's just me, but don't use just PDO constants? I dont see the benefit.

eridal:
Hey courtney-miles,
I leaved some small nits, but otherwise great work!!

src/Phinx/Db/Adapter/PdoAdapter.php
courtney-miles: Hey eridal,

PDO will always use \PDO::PARAM_STR if you don't otherwise specify:

    PDOStatement::bindValue ( mixed $parameter , mixed $value [, int $data_type = PDO::PARAM_STR ] )

I don't mind adopting your suggestion, but the outcome is the same.  Let me know.

src/Phinx/Migration/AbstractMigration.php
courtney-miles: This was a design from #256.

I believe the goal is to abstract away the adapter that is being used in the migrations your write.

It's not something I would have thought to do myself, but I believe the author, @cyrusboadway, was right to do it this way.

There is also an abstract `getAdapterBindParamType()` method now such that if an alternative to PDO is introduced, this method will accept one these constants and return a value suitable for the new adapter.

Phinx is clearly designed to accommodate different adapters, so I think it makes sense to provide constants that are not specific to one adapter (even if the constants values are based on one in particular).

docs/migrations.rst
```diff
@@ -205,6 +205,84 @@ Queries can be executed with the ``execute()`` and ``query()`` methods. The
...
+        use Phinx\Migration\AbstractMigration;
+
+        class MyNewMigration extends AbstractMigration
+        {
+            /**
+             * Migrate Up.
+             */
+            public function up()
+            {
+                $queryParams = array(
```
cyrusboadway: So this is the same query format that I'd recommended in the RFC way back, so I clearly think this is the _good_ solution, but I worry about arrays, especially multi-dimensional arrays, as a way of specifying behaviour. Phinx is already doing a tonne of this (such as [setting column attributes in a column definition](http://docs.phinx.org/en/latest/migrations.html#creating-a-table)). My problem is that the user has to know this particular data structure to define the correct behaviour, in this case bind-typing. This is some dense logic:

> If a `$queryParams` item value is not an array, then the array item value is the bind-value. Otherwise, the item-value array must have an item with key `value` to specify the bind-value, and optionally an item with key `type` specifying the bind-type.

Gross. Try getting your dev team to internalise that logic. And I learn this where, in the documentation, or by reading examples? Not to mention, if this migration tool starts supporting [JSON column types](https://mariadb.com/kb/en/mariadb/column_json/), the bind value might legitimately be an array….

So what about an object oriented approach, where the binds are all defined in classes? The method signatures explain the behaviour, and where there's any confusion, the docblocks explain the rest:

```php
// This example has sorely insufficient documentation
interface QueryBind {
    const TYPE_INT = \PDO::PARAM_INT;
    const TYPE_BOOL = \PDO::PARAM_BOOL; // etc…
    /**
     * @param mixed $bindValue
     */
    public function __construct($bindValue);
    /**
     * @param int $bindType Should match QueryBind::TYPE_*
     * @return $this
     * @see \PDO
     * @link http://php.net/manual/en/pdo.constants.php
     */
    public function setBindType($bindType);
    /**
     * @return int By default `static::TYPE_STR`
     * @link http://php.net/manual/en/pdo.constants.php
     */
    public function getBindType();
}

class AbstractMigration {
    /**
     * @param string $query
     * @param QueryBind[] $binds
     */
    public function execute($query, array $binds = null){ /* magic */ }
}

class SomeMigration extends AbstractMigration{
    public function up(){
        $binds = [
            'bindA' => new QueryBind('somestring'), // default is string bind
            'bindB' => (new QueryBind(5))->setBindType(QueryBind::TYPE_INT),
            'bindC' => new QueryBind(5, QueryBind::TYPE_INT), // or perhaps this as an alternative?
        ];
        $this->execute('SELECT … WHERE a=:bindA AND b=:bindB AND c=:bindC', $binds);
    }
}
```

The benefit here is the behaviour is prescribed by the object's methods and parameters, rather than in some array format.

Anyway, all this to say that when I did my version I still went with the array model. I just thought OOO alternative version should be discussed.

src/Phinx/Migration/AbstractMigration.php
cyrusboadway: I'm all for aliasing the `PDO` params to something local-to-the-class. `PDO` is a dependency, to which direct references should be discouraged.

src/Phinx/Db/Adapter/PdoAdapter.php
cyrusboadway: :+1: leaving it as is. The behaviour is defined within the class: i.e. less surprising.

docs/migrations.rst
courtney-myaffiliates: I agree with benefits of defining a bind class.  

I'm was not comfortable with using structured arrays (they are too free-form), but I was avoiding taking liberties with the solution already proposed.

If there are no objection, I'm happy to implement the above suggestion.

cyrusboadway: That proposal was meant only to begin a discussion about a mechanism for bound params. I'd much rather see an object oriented API, so if my vote's worth anything, I :+1: that.

courtney-miles:
I'm keen to see this finished, but have been waiting for guidance.

The the project maintainers have no preferences, I will continue with changes discussed above.

src/Phinx/Migration/AbstractMigration.php
shadowhand: :+1: I agree that these constants should exist.

docs/migrations.rst
shadowhand: I like the idea of including a bind class, but I don't think it's entirely necessary for *all* binds. Of the listed syntaxes, I think that this one is best:

```php
new QueryBind(5, QueryBind::TYPE_INT)
```

However, for scalar types, I don't think it's entirely necessary. It should be possible to simply have:

```php
$bindValues = array(
    'a' => 'foo',
    'b' => 15
);

And have this automatically be recognized as `TYPE_STRING` and `TYPE_INT`. I see the QueryBind being most useful when dealing with complex data. With simple data, it needs to get out of the way or it will be cumbersome.

courtney-myaffiliates: I agree that you should be able to use named parameters without using the `QueryBind` class.  But if you don't use `QueryBind` then the bind type should always be a string -- this mimics the default behaviour of PDO.

zeroedin-bill:
Please merge this. I would greatly appreciate having this capability, as it is the only thing I have found lacking in Phinx.

courtney-miles:
I have pushed changed to include using a `QueryBind` class for if you need to bind a value to a query as a specific SQL type.

The continuous-integration/appveyor build failed, but I'm not clear if it's something I have caused, sorry.

ncareau:
Any progress on this ? or are there any other way to bind params to a query right now ? 
Also, i like the idea of the object QueryBind but also would like to be able to pass an array of name=>key that automaticaly binds to the query. I think i'ts easier and more in line with PDO.
Thanks a lot in advance.

courtney-miles:
Hey @ncareau 

I'm afraid I don't know what to do next to get this pull request accepted.  The appveyor test has failed for a reason I don't think is caused by my changes.

Note that you will have the options to pass an array of key/values.  `queryBind()` is only for if you need to bind a value as a specific type.

The work around until this feature is accepted is to grab the PDO adapter and use it instead.  To do that, use this line in your migration method.

    $pdo = $this->getAdapter()->getConnection();

ncareau:
@courtney-miles Thanks for the quick reply, i'll try that.

ncareau:
@courtney-miles Also, look into #429. It seems the issue with appveyor was fixed, so maybe try to push anything to force a new build.

courtney-miles:
Pushed a commit to trigger, but now seeing, "AppVeyor was unable to build non-mergeable pull request."

I have been merging with `master`.  I'm thinking I need to merge with `0.4.x-dev`.  Can anyone confirm before I do?

rquadling:
I rebase my features (PRs) onto 0.4.x-dev, rather than master. That has worked for the features I've provided.

courtney-miles:
Should I be merging the `0.4.x-dev` branch into `queryparams` again to resolve these conflicts and get the ball rolling again?

rquadling:
@courtney-miles You have to resolve the conflicts in your branch. It isn't the package owner's job to do that. The best way to achieve this is to rebase your changes onto the ```0.4.x-dev``` branch.

Once your branch is uptodate with all the latest changes, make sure all the unit tests run and pass. Fix your code to pass the tests or, if your code has replaced existing functionality, then new tests and retire the dead ones (or at least change them to expected to fail tests).

cyrusboadway:
@rquadling: @courtney-miles rebased [six months ago](https://github.com/courtney-miles/phinx/commit/6ea728be16e60b60ebf7c223e5bc7c8cd65cb79c) with all tests passing. He also [politely requested feedback](https://github.com/robmorgan/phinx/pull/486#issuecomment-75186446) from the package maintainers, without any response. I suspect he's politely asking whether it's worth the effort.

rquadling:
@cyrusboadway Didn't notice the time line. Infrequent bursts of activity from the package maintainers. I wonder if they need more volunteers.

BallisticPain:
@courtney-miles 
I realize you did a re-base a while back and no one merged your result.
Could you please indicate if you intend to try again? If not, I would be more than happy to try myself by submitting a new PR and referencing this one.

@robmorgan 
If @courtney-miles does this, can you merge soonish? I can only imagine you're quite busy.

Thanks!

courtney-miles:
I'm not sure these check failures are due to my contributions.

I merged in `0.4.x-dev` but I see there is a `0.5.x-dev` now and I'm not sure if I should be merging that branch now.

I also mistakenly merged `master`.  If anyone believes that may be a problem, I will revert the merges and try again.

shadowhand:
@courtney-miles Thank you for contributing! The branch you made this PR against is no longer accepting patches. This probably happened because we didn't merge your changes fast enough. :sob:

wesleywesley:
So what is the correct syntax for this? I tried some of the code samples above but that only results in a syntax error message.. Where is the documentation for this feature??

BallisticPain:
@wesleywesley

If you use the code sample grabbing the PDO connection as stated here: https://github.com/robmorgan/phinx/pull/486#issuecomment-105323881 Then you just use the PDO documentation found on PHP.net

Hope this helps,
Jarvis

raymondjplante:
Wish there was a way to do this without having to fall back to underlying connector :/

raymondjplante:
Is there documentation or a full example on using parameterized queries  vs. falling back on the PDO adapter?

courtney-miles:
A new PR has been raised for the 0.5.x-dev branch #850

dchao6:
Pull request #1138 made in 0.9.0 requires a change to the code sample grabbing the PDO connection

**Previous to 0.9.0**: https://github.com/cakephp/phinx/pull/486#issuecomment-105323881
`$pdo = $this->getAdapter()->getConnection();`

**0.9.0 and after**: https://github.com/cakephp/phinx/issues/1182#issue-257084065
`$pdo = $this->adapter->getAdapter()->getConnection();`

> ... you can check if getAdapter is returning a wrapper and called getAdapter again until it is not?
https://github.com/cakephp/phinx/issues/1182#issuecomment-328904010

PR #487, opened 2015-02-13
Sqlite changeColumn bugfix

boedy:
When changing a primary key column to a BIGINTEGER this doesn't work because SQLite doesn't support a BIGINT auto incrementing column. However SQLite uses a dynamic type system. We can therefor use INTEGER. See: http://www.sqlite.org/datatype3.html.

What happens if you run the code from the new test All other columns would be removed from the table.


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -793,7 +793,7 @@ public function dropForeignKey($tableName, $columns, $constraint = null)
     /**
      * {@inheritdoc}
      */
-    public function getSqlType($type, $limit = null)
+    public function getSqlType($type, $limit = null, $identity = false)
```
shadowhand: I don't recommend modifying this signature, as it is defined by the AdapterInterface.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -979,7 +981,7 @@ protected function getDefaultValueDefinition($default)
      */
     protected function getColumnSqlDefinition(Column $column)
     {
-        $sqlType = $this->getSqlType($column->getType());
+        $sqlType = $this->getSqlType($column->getType(), $column->isIdentity());
```
shadowhand: This would be the better place to make the modification... If `$column->getType()` is `PHINX_TYPE_BIG_INTEGER` and `$column->isIdentity()`, then send `PHINX_TYPE_INTEGER` to `getSqlType` instead.

robmorgan:
this PR shouldn't modify any interface signatures, it must be cleaned up first.

shadowhand:
Thank you for trying to contribute. Due to the lack of response on the changes that were requested, I am going to close your pull request.

PR #492, opened 2015-02-16
Added table name as part of index for SQLite adapter

sorend:
The SQLite adapter uses named indexes (assuming because SQLite needs named indexes), and the current implementation does not use table name as part of the index. This patch changes that.

When two tables define an index on a field with the same name, only one of the indexes is created. No errors are given from Phinx in this case.

A work-around is to use named indexes. (i.e. addIndex(array("field"), array("name" => "my_index")) ).

I've executed the tests after making the change, hoping it won't break anything, let me know if more is required :-)


shadowhand:
It would be nice if this included unit tests.

sorend:
@shadowhand I have created a separate pull request for the unit test. Sorry about not including in first case, I'm a little new to PHP :-)

PR #495, opened 2015-02-22
Added test for two indexes with same name in different tables, SQLite.

sorend:
This is the test for the previous patch. Two tables with same indexed attribute in SQLite.


shadowhand:
thanks!

PR #497, opened 2015-02-23
Initial implementation of a breakpoint for migration rollback.

rquadling:
See https://github.com/robmorgan/phinx/issues/494 for the issue.


robmorgan:
waiting to see if theres further support from the community

pedanticantic:
This is a really good feature and I would love to see if incorporated into phinx! It means you can have a "safety net" so you don't accidentally roll back _all_ your database changes. It also means that when you start a new task, you can put a breakpoint in on the "current" change, and it's really easy to roll back to that point while you're working on that task.

shadowhand:

PR #502, opened 2015-02-23
No prev migration check on rollback

Ducain:
When a team is working on various branches, a situation can arise where rolling back the most recent migration means the previous migration doesn't exist, because it was executed when another branch was checked out. It seems logical to me that executing a rollback with no target should not fail if the previous migration doesn't exist on the current branch, as no action is being taken on that previous migration.

This change executes the rollback of the most recent migration, without checking for the existence of the previous migration.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -253,7 +253,7 @@ class Manager
                 }
             }
 
-            exit();
+            return;
         } else {
             // Get the first migration number
             $first = reset($versions);
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -246,13 +246,19 @@ class Manager
 
             $this->getOutput()->writeln('<info>Attempting to rollback version: </info><comment>' . $version . '</comment>');
 
+            $migration_found = false;
+
             foreach ($migrations as $migration){
                 if ($migration->getVersion() == $version) {
-                  $this->executeMigration($environment, $migration, MigrationInterface::DOWN);
-                  break;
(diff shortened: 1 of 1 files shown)


temperedvision:
This change makes sense to me and would go some way to deal with the scenario mentioned in the pull request (developing and testing migrations in isolation against a shared database)

RQuadling:
If you are using branches with each having its own migrations which are isolated from each other until feature finished, when you switch branches without first rolling back the outgoing branch and migrating the incoming branch, your DB will be a bit of a mess.

Unless you have a different pattern of behaviour.

How do you manage that?

Ducain:
@RQuadling As any branch can have multiple migrations (most often the addition of tables and columns), and during development new data added for feature testing, rolling back all migrations for a branch before switching to another would become inefficient very quickly.  Additionally, the existence of tables or table columns doesn't constitute 'a mess' in our minds.  It would be odd for the existence of a table and/or columns to break functionality on another branch. 

I created a new phinx phar with the changes I made here, and we've been using it in development. Definitely an improvement in our workflow.

RQuadling:
@Ducain, ah. OK. Our migrations include views/sp/triggers/etc. Not just an additional column here or there. For our RI, the public interface of our DB for our repo mechanics need to tie up, so the rollback/migrate works well for us. I think using solid state drive has spoiled me rotten.

Ducain:
@RQuadling Makes sense. We use very few views/sp/triggers because we're in MySQL and try to limit the existence of such things. Our migrations are almost completely tables/columns.

shadowhand:

PR #465, opened 2015-01-23
Added multi-database support

bartekrylko:
Based on #452 - with suggested changes.

PR #503, opened 2015-02-24
Multidb support

bartekrylko:
This is based on #465 PR. All changes are in one commit now.
Configuration details are in `docs/configuration.rst` file.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -143,7 +143,7 @@ class Config implements ConfigInterface
     /**
      * {@inheritdoc}
      */
-    public function getEnvironment($name)
+    public function getEnvironment($name, $database = null)
     {
         $environments = $this->getEnvironments();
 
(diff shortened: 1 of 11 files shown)


src/Phinx/Config/Config.php
```diff
@@ -153,6 +153,49 @@ public function getEnvironment($name)
...
+                if ($database != $environments[$name]['name']) {
+                    throw new \InvalidArgumentException(sprintf(
+                        'The database "%s" does not exists in environment "%s"',
+                        $database,
+                        $name
+                    ));
+                }
+                $environments[$name]['name'] = $database;
+            }
+            elseif (!empty($environments[$name]['databases']) && $database !== null ) {
```
shadowhand: But if `$database = false`, is it still okay?

bartekrylko: Moslt likely not. Proper value is verified in next condition.
`$database !== null` is in condition to verify if parameter was sent, or default value is provided.
Imho no need to check if `$database === false` but if you want, i can add it.

bartekrylko:
hello, do we know something more here?

rlx123:
Hey guys, any news on multidb support ? I don't conceal the fact that it will be very useful.

src/Phinx/Config/Config.php
shadowhand: I would recommend using `!empty($database)` instead.

shadowhand:

PR #507, opened 2015-02-27
Update configuration.rst

rdeutz:
added MySQL prefix and suffix documentation and structured the adapter area a bit

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -89,6 +89,25 @@ file:
 
     export PHINX_ENVIRONMENT=dev-`whoami`-`hostname`
 
+
+Table Prefix and Suffix
+------------------
+
+You can define a table prefix and table suffix:
+
+.. code-block:: yaml
+
+    environments:
+        development:
+            ....
+            table_prefix: dev_
+            table_suffix: _v1
+        testing:
+            ....
+            table_prefix: test_
+            table_suffix: _v2
+
+
 Socket Connections
 ------------------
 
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -143,6 +143,26 @@ Phinx currently supports the following database adapters natively:
 * `SQLite <http://www.sqlite.org/>`_: specify the ``sqlite`` adapter.
 * `SQL Server <http://www.microsoft.com/sqlserver>`_: specify the ``sqlsrv`` adapter.
 
+MySQL
+`````````````````
+
+Since version 0.4.0 you can define a table prefix and table suffix:
```
temperedvision: Documentation should always refer to the current version, so there really shouldn't be any references to versions or "works since" notes as this is irrelevant info, and should be kept to changelogs.

rdeutz: fine by me, needed something so that it isn't there without a comment

docs/configuration.rst
```diff
@@ -143,6 +143,26 @@ Phinx currently supports the following database adapters natively:
 * `SQLite <http://www.sqlite.org/>`_: specify the ``sqlite`` adapter.
 * `SQL Server <http://www.microsoft.com/sqlserver>`_: specify the ``sqlsrv`` adapter.
 
+MySQL
```
shadowhand: Prefixing should work with all adapters, not just MySQL.

PR #511, opened 2015-03-06
Support for smallint to PostgreSQL database

vitormattos:
Added support for numeric columns of type smallint similarly the implementation of column types used in the version of MySQL.

Updated the documentation with the content of the changes made.

Implemented unit testing.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -36,6 +36,8 @@ use Phinx\Migration\MigrationInterface;
 
 class PostgresAdapter extends PdoAdapter implements AdapterInterface
 {
+
+    const INT_SMALL   = 65535;
     /**
      * Columns with comments
      *
@@ -708,6 +710,13 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
     {
         switch ($type) {
             case static::PHINX_TYPE_INTEGER:
+                if ($limit && $limit == static::INT_SMALL) {
+                    return array(
+                        'name' => 'smallint',
+                        'limit' => static::INT_SMALL
+                    );
+                }
+                return array('name' => $type);
             case static::PHINX_TYPE_TEXT:
             case static::PHINX_TYPE_DECIMAL:
             case static::PHINX_TYPE_TIME:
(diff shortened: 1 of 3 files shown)


tests/Phinx/Db/Adapter/PostgresAdapterTest.php
```diff
@@ -415,18 +415,19 @@ public function testGetColumns()
...
-              ->addColumn('column6', 'decimal')
-              ->addColumn('column7', 'time')
-              ->addColumn('column8', 'timestamp')
-              ->addColumn('column9', 'date')
-              ->addColumn('column10', 'boolean')
-              ->addColumn('column11', 'datetime')
-              ->addColumn('column12', 'binary')
-              ->addColumn('column13', 'string', array('limit' => 10));
+              ->addColumn('column2', 'integer', array('limit' => PostgresAdapter::INT_SMALL))
+              ->addColumn('column3', 'integer')
```
shadowhand: Why were all these other lines changed?

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -772,7 +781,13 @@ public function getPhinxType($sqlType)
             case 'text':
             case 'json':
                 return static::PHINX_TYPE_TEXT;
+            case 'smallint':
+                return array(
+                    'name' => 'smallint',
+                    'limit' => static::INT_SMALL
+                );
             case 'int':
+            case 'int2':
```
shadowhand: What does `int2` have to do with smallint?

tests/Phinx/Db/Adapter/PostgresAdapterTest.php
vitormattos: The smallint column was inserted before the column integer, to keep the sequence, I added one to the name for all columns.

src/Phinx/Db/Adapter/PostgresAdapter.php
vitormattos: I asked the same question but in the rush to use smallint in my project, I did not analyze precisely because the `getPhinxType()` method, has types that are not used in the PostgreSQL adapter.

Following my analysis:
In commit 7bad4f9d5436be08ae40f98fc77e1b638d780534 @czogori added some types in `getPhinxType()` method that are not returned by `getColumns()` method. I believe that the intention of @czogori was to use the column [udt_name] (http://www.postgresql.org/docs/8.4/static/infoschema-columns.html) of the PostgreSQL table `information_schema.columns`.

At this time, I removed the `int2` and I will make another pull request with a refactor to improve the use of types to postgreSQL, okay? These other types are not implemented in Phinx but may be helpful for future implementation of other types used by PostgreSQL.

Note: `int2` is `udt_name` to smallint in PostgreSQL

vitormattos:
what is the status of this pull request?
What do I need change more?
The fail in CI is not because my changes.

PR #514, opened 2015-03-08
SQLiteAdapter fixes for transaction, createTable() with foreign key and addForeignKey()

allcode:
This PR contains the following three fixes within the SQLiteAdapter:

#### beginTransaction
Fixed `beginTransaction()` query. It now uses `BEGIN TRANSACTION` (see: https://www.sqlite.org/lang_transaction.html). Using `START TRANSACTION` with PDO error mode set to exception/error (see http://php.net/manual/en/pdo.error-handling.php) will result in an error. Added `testBeginTransaction` test.

#### createTable with foreign key
Fixed erroneous create table query in `createTable()` when the table has one or more foreign keys.
`createTable()` now adds the foreign key statements to the column definitions, inside the parentheses, instead of after (see: https://www.sqlite.org/lang_createtable.html). Added `testCreateTableWithForeignKey` test.

#### addForeignKey
Fixed two issues in `addForeignKey()`. The `INSERT INTO ... SELECT` query did not contain any column information, resulting in an error when PDO error mode was set to exception/error. This was due to an always empty `$columns` array. This is fixed by assigning the `pragma table_info` result to `$rows` and using it to build the `$columns` array. The table names in the `INSERT INTO ... SELECT ` query are now quotes to prevent issues when using reserved words. Added `testAddForeignKeyWithPdoExceptionErrorMode` test.

### DIFF
--- a/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
@@ -455,6 +455,24 @@ class SQLiteAdapterTest extends \PHPUnit_Framework_TestCase
         $this->assertTrue($this->adapter->hasForeignKey($table->getName(), array('ref_table_id')));
     }
 
+    public function testAddForeignKeyWithPdoExceptionErrorMode()
+    {
+        $this->adapter->getConnection()
+            ->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+        $refTable = new \Phinx\Db\Table('ref_table', array(), $this->adapter);
+        $refTable->addColumn('field1', 'string')->save();
+
+        $table = new \Phinx\Db\Table('table', array(), $this->adapter);
+        $table->addColumn('ref_table_id', 'integer')->save();
+
+        $fk = new \Phinx\Db\Table\ForeignKey();
+        $fk->setReferencedTable($refTable)
+            ->setColumns(array('ref_table_id'))
+            ->setReferencedColumns(array('id'));
+
+        $this->adapter->addForeignKey($table, $fk);
+    }
+
     public function testDropForeignKey()
     {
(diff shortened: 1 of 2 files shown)


tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -41,6 +41,13 @@ public function testConnection()
         $this->assertTrue($this->adapter->getConnection() instanceof \PDO);
     }
 
+    public function testBeginTransaction()
+    {
+        $this->adapter->getConnection()
+            ->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+        $this->adapter->beginTransaction();
```
shadowhand: No assertion?

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -434,6 +455,24 @@ public function testAddForeignKey()
...
+
+        $table = new \Phinx\Db\Table('table', array(), $this->adapter);
+        $table->addColumn('ref_table_id', 'integer')->save();
+
+        $fk = new \Phinx\Db\Table\ForeignKey();
+        $fk->setReferencedTable($refTable)
+            ->setColumns(array('ref_table_id'))
+            ->setReferencedColumns(array('id'));
+
+        $this->adapter->addForeignKey($table, $fk);
```
shadowhand: Missing assertions.

allcode: Added a hasForeignKey assertion.

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
allcode: Added `assertTrue(true)` assertion as we are only testing the validity of the query.

PR #523, opened 2015-03-20
Fixed change comment block

LukePeterson:
Improves the ability to uncomment the block in one hit.  Also makes the markdown render the commented block correctly (removing the red highlighting)


robmorgan:
Keep track of: https://github.com/robmorgan/phinx/issues/287

PR #527, opened 2015-03-24
Documentation for init() method

oytuntez:
init() method from AbstractMigration class was not documented, but it can actually be very useful in some circumstances.


robmorgan:
I'm closing this. I'd prefer to leave the `init()` method undocumented. `change()` is now the default anyway.

PR #532, opened 2015-04-08
Added support for ALTER TABLE ... AFTER in changeColumn()

vingborg:
Adding support for 'ALTER TABLE ... AFTER ...' feature in changeColumn() for mysql adapter. Used the same as for addColumn(), e.g. $table->changeColumn('colname', 'integer', array('after' => 'anothercolumn')).


daniel-lang:
:+1:

PR #536, opened 2015-04-16
Added documentation for the hasColumn method.

shanahanjrs:
Phinx documentation does not include the function `hasColumn` which is very useful and could help many people write cleaner, smaller code.


robmorgan:
@shanahanjrs thanks

Issue #265, opened 2014-06-30
A column can not be "NOT NULL" unless it has a default value (SQLite)


nixxquality:
In the SQLite adapter you can not use just `array('null' => false)` to specify that a column is "NOT NULL", you also have to specify a default value[[1]](https://github.com/robmorgan/phinx/blob/0.3.x-dev/src/Phinx/Db/Adapter/SQLiteAdapter.php#L945).

``` php
$def .= ($column->isNull() || is_null($default)) ? ' NULL' : ' NOT NULL';
```

This behavior is different from the other adapters[[2]](https://github.com/robmorgan/phinx/blob/0.3.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L843)[[3]](https://github.com/robmorgan/phinx/blob/0.3.x-dev/src/Phinx/Db/Adapter/PostgresAdapter.php#L821), so I am assuming it is a bug.

``` php
$def .= ($column->isNull() == false) ? ' NOT NULL' : ' NULL';
$buffer[] = $column->isNull() ? 'NULL' : 'NOT NULL';
```

robmorgan:
I've merged a test which is reproducing this behaviour, but I've marked it as skipped for now.

I tried to fix this bug but it turns out its not as simple as it seems. I can make it work when creating a table, however altering one is a different story. SQLite only supports a limited subset of ALTER TABLE:

> The new column may take any of the forms permissible in a CREATE TABLE statement, with the following restrictions:
 > * If a NOT NULL constraint is specified, then the column must have a default value other than NULL.

See: http://www.sqlite.org/lang_altertable.html for more information.

I guess Phinx would have to work around this behaviour and inject a default value say an empty string or 0 depending on the column type when altering a table.

I'll leave this open for now.

K4T:
Hi, do we have any progress with that bug?

aldas:
having `['null' => false, 'default' => '']`in DDLs created by create table is pretty insane. basically you allow creating rows with almost as bad as with null values enabled. 

Maybe `getColumnSqlDefinition` could know in which ddl context it is called (CREATE/ALTER/...) and in sqlite not force you to use default?

aldas:
I would like to point out that Sqlite `ALTER TABLE` does not support altering existing column. So there is no way you could alter NOT NULL column to NULL value - therefore I think CREATE should not force column NOT NULL with DEFAULT set. 

and adding new column with NOT NULL and without DEFAULT would trigger database level error anyway
```
[...]

PR #546, opened 2015-05-03
Added test for null/default value issue in SQLiteAdapter

sorend:
This is a (possible) test for Issue #265.


robmorgan:
Merged but see #265 for more information.

PR #550, opened 2015-05-06
Throw RuntimeException when default CURRENT_TIMESTAMP on DATETIME field not supported

HughbertD:
Ran into this issue: https://github.com/robmorgan/phinx/issues/261

This PR simply creates a little more visibility to the problem which otherwise isn't very obvious. 

If the version of MySQL is too low to support a CURRENT_TIMESTAMP DEFAULT on a DATETIME column, throw an Exception to give some visibility to the user.


robmorgan:
I'm closing this, I think if anything it should be mentioned in the docs, but its not a good idea to sniff specific versions of MySQL at the state Phinx is currently in.

Issue #549, opened 2015-05-05
Adding "bootstrap" handling (via an external PHP file)


enygma:
I have a situation where it would be helpful to be able to call a PHP bootstrap file prior to the execution of the migrations. Ideally it would be able to set configuration information back into the application, possibly called either from the command line:

```
vendor/bin/phinx migrate --bootstrap=/path/to/bootstrap.php
```

or in the configuration file:

```
paths:
    migrations: %%PHINX_CONFIG_DIR%%/migrations
    bootstrap: /path/to/bootstrap.php
```

I don't know if this functionality might be useful to anyone else out there but I could see it being handy for my personal needs.

sergiors:
:+1:

cassiosantos:
:+1:

m1ome:
This is obvious a copy-of-functionality.
To achieve this you can use [php configuration files](http://docs.phinx.org/en/latest/commands.html#configuration-file-parameter)

PR #553, opened 2015-05-07
Add bootstrap file handling

enygma:
This PR is related to functionality mentioned in Issue #549 about executing a bootstrap file when any command is called.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -156,6 +156,22 @@ In case with PHP array you can provide ``connection`` key with existing PDO inst
                 )
             );
 
+Bootstrap File Parameter
+----------------------------
+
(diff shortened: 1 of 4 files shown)


MasonM:
:+1: this would be very handy

enygma:
Any chance of this getting merged? Looks like the Scrutinizer build failed for no good reason...

m1ome:
You can use a config file as a bootstrap file.

enygma:
Um, correct me if I'm wrong but the configuration files are yaml files, not PHP and those don't allow loading or calling out to external scripts.

m1ome:
@enygma, you can use a ".php" file for configuration, here you can do you "bootstrap" thing, please check the [docs](http://docs.phinx.org/en/latest/commands.html#configuration-file-parameter) you can do what you want in this file and then "return" a **Phinx** configuration.

robmorgan:
@enygma please refer to the docs. PHP-based configs are now documented and should work for you.

enygma:
@robmorgan thanks for the update! I'll check into it.

PR #554, opened 2015-05-11
Fixed: When the column option 'scale' was ZERO, it would use 'limit' …

SBhojani:
…instead of 'precision' to create the SQL.

The code:

    $table->addColumn('max_amount', 'decimal', [
            'default' => null,
            'precision' => 50,
            'scale' => 0,
            'null' => false,
            'after' => 'min_amount',
        ]);

was creating a `decimal(10,0)` instead of a `decimal(50,0)`. This fixes that.


robmorgan:
we need a unit test to wrap this.

shadowhand:

Also please add tests if you recreate this.

PR #557, opened 2015-05-17
Improved docs and fixed Commands interfaces

corpsee:
Improved docs and fixed Commands interfaces


robmorgan:
hi @corpsee, a few things

* there is no need to return 0 in the Phinx commands as the Symfony Command class will automatically default to 0.
* the phpdoc new lines are inconsistent with the rest of the project.
* However some of your phpdocs are correct, with a null | int return type compared with our void.

I'm closing this for now.

PR #559, opened 2015-05-20
simplify check for PostgresAdapter::hasTable

steffkes:
Trying to use my own connection, i've got `Notice: Undefined offset: 0` because i'm using `PDO::FETCH_ASSOC` which doesn't return numeric indexes for its result.

Since we're only interested in one column anyway, we could change it to `PDOStatement::fetchColumn` which would work regardless of the underlying connection and its options.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -161,7 +161,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
             )
         );
 
-        return $result->rowCount() !== 0;
+        return $result->rowCount() === 1;
     }
 
     /**
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -151,10 +151,8 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
     public function hasTable($tableName)
     {
         $tables = array();
-        $rows = $this->fetchAll(sprintf('SELECT table_name FROM information_schema.tables WHERE table_schema = \'%s\';', $this->getSchemaName()));
-        foreach ($rows as $row) {
-            $tables[] = strtolower($row[0]);
-        }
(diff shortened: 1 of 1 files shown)


steffkes:
I have no idea why the Scrutinizer Build is failing? Doesn't look like it's change related, is it?

steffkes:
@robmorgan any idea why that scrutinizer build failed? on the other hand, any thoughts on the PR?

steffkes:
@robmorgan any feedback on this PR at all?

rquadling:
I may have missed the reason, but why are **all** the tables queried, retrieved, in_array(), discarded?

Surely the query should look for the specific table?

steffkes:
@rquadling uhm, that's a valid question - i didn't look much at the context just fixed my problem, that's a good idea, will include that one. indeed no need to retrieve all the information :)

steffkes:
mind having another look @rquadling? otherwise, i'd say we're ready to go @robmorgan

robmorgan:
looks good to me, but i'll test it tomorrow. thoughts @rquadling ?

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -150,12 +150,18 @@ public function quoteColumnName($columnName)
...
+                'SELECT *
+                FROM information_schema.tables
+                WHERE table_schema = %s
+                AND lower(table_name) = lower(%s)',
+                $this->getConnection()->quote($this->getSchemaName()),
+                $this->getConnection()->quote($tableName)
+            )
+        );
+
+        return $result->rowCount() !== 0;
```
rquadling: Should/would this be better as ```return $result->rowCount() === 1;``` as we are expecting exactly 1 row for a truthy value. Because this is PHP, ```false`, ```-1```, ```99```, ```'who cares'```, ```null```, would all be true.

steffkes: True that, changed the comparison to `=== 1`

steffkes:
Scrutinizer check failed for unknown reason, again :/

PR #560, opened 2015-05-27
Correctly described query() method for Phinx 0.4.3

crantok:
Returns a PDOStatement, not an array.


crantok:
Oops. Didn't notice the doc was and RST rather than an MD. Have not encountered an RST before.

robmorgan:
fixed in a7510f32957b016eb9221cf4f50c2529952a138e

PR #581, opened 2015-07-02
Fix #535

### LINKED ISSUE #535: Sanity check for addTimestamps() function
I tried using the addTimestamps() function today and what happened seemed to me to be incorrect behavior.  When doing a DESC on my mysql table I got back the following:
```
+--------------+--------------+------+-----+-------------------+-----------------------------+
| Field        | Type         | Null | Key | Default           | Extra                       |
+--------------+--------------+------+-----+-------------------+-----------------------------+
| created_at   | timestamp    | NO   |     | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
| updated_at   | timestamp    | YES  |     | NULL              |                             |
+--------------+--------------+------+-----+-------------------+-----------------------------+
```
I would expect the Extra field of 'on update CURRENT_TIMESTAMP' to be attached to the `updated_at` field not the `created_at` field.  I don't know that this is incorrect behavior, but it is not what I would have expected to occur, so, I figured I would at least point it out.  If you could please let me know if this is the intended behavior which Phinx wants or if this is an actual issue.

Thanks.

elliotwms:
`created at` shouldn't have an updates at

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -556,10 +556,13 @@ class Table
      */
     public function addTimestamps()
     {
-        $this->addColumn('created_at', 'timestamp')
+        $this->addColumn('created_at', 'timestamp', array(
+                'default' => 'CURRENT_TIMESTAMP',
+                'update' => ''
+            ))
              ->addColumn('updated_at', 'timestamp', array(
-                 'null'    => true,
-                 'default' => null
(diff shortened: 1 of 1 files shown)


elliotwms:
Note: Travis failed on 5.5 and hhvm because of a composer installation issue

robmorgan:
thanks

PR #585, opened 2015-07-05
Add function for soft deleting table column

elliotwms:
Adds the ability to add a `deleted_at` column to a table using either of two methods:

- As a function on the table

```php
$table->addSoftDelete()
```

- As an argument in the `addTimestamps()` function

```php
$table->addTimestamps(true)
```

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -558,13 +558,13 @@ class Table
     public function addTimestamps($softDelete = false)
     {
         $this->addColumn('created_at', 'timestamp', array(
-                'default' => 'CURRENT_TIMESTAMP',
-                'update' => ''
-            ))
-             ->addColumn('updated_at', 'timestamp', array(
+            'default' => 'CURRENT_TIMESTAMP',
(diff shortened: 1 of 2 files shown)


elliotwms:
I was going to write some documentation for this but there doesn't seem to be anything on `addTimestamps()` itself in the docs so I was unsure of where to put it

robmorgan:
thanks @elliotwms most of our users rely on `created_at` or `updated_at`, hence why I included the `addTimestamps` method. When it comes to deletion, rows are normally simply discarded. However if this feature proves popular then I'll consider merging it.

elliotwms:
Fair enough, I've just noticed that I'd been implementing this manually in many migrations as an application I'm developing relies on data not being permanently discarded (for auditing purposes and the like). However I agree that this use case may only apply to a small subset of users, so I understand your position

eddturtle:
:+1: for this idea

w2pc:
I like this idea.  I have been creating them in migration. Would be nice to
have it automatically with a parameter setting

PR #608, opened 2015-07-24
Documentation improvements for valid column types

mamchenkov:
* Added ```uuid``` to the list of columns
* Added ```jsonb``` to Postgres
* Sorted column types alphabetically

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -417,7 +417,7 @@ Column types are specified as strings and can be one of:
 
 In addition, the MySQL adapter supports ``enum`` and ``set`` column types.
 
-In addition, the Postgres adapter supports ``json`` column types
+In addition, the Postgres adapter supports ``json`` and ``jsonb`` column types
 (PostgreSQL 9.3 and above).
 
 For valid options, see the `Valid Column Options`_ below.
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -413,10 +413,11 @@ Column types are specified as strings and can be one of:
 -  date
 -  binary
 -  boolean
+-  uuid
 
 In addition, the MySQL adapter supports ``enum`` and ``set`` column types.
 
-In addition, the Postgres adapter supports ``json`` and ``uuid`` column types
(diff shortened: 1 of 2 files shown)


w2pc:
IMO uuid should not be added to the list here as that is not universal.
It's already noted that uuid is available for the Postgres adapter

On Fri, Jul 24, 2015 at 5:00 AM, Leonid Mamchenkov <notifications@github.com

mamchenkov:
@w2pc to me it looks like the documentation is misleading, since it says that only Postgres supports UUID.  We are using UUID just fine with MySQL and SQL Server.  I took a quick look around and it seems that SQLite adapter supports it too (which makes it 4 out of 4).  PDO adapter, which all other ones are extending, supports it.  And the phinx type constant is defined in AdapterInterface, which all adapters implement.

To me it looks pretty universal.

w2pc:
But a UUID in MySQL is a column type of 16byte "binary" which is already in
the column types.

On Fri, Jul 24, 2015 at 6:34 AM, Leonid Mamchenkov <notifications@github.com

mamchenkov:
@w2pc not all databases support UUID natively.  Those that don't, are using ```char(36)```.  The adapters are mapping this difference pretty nice, so that from the point of view of migrations, you don't even need to know if the engine supports uuid column type.

Currently, the documentation suggests that uuid column type is only supported by Postgres, so you probably should (implicitly) check the database engine in your migration and use ```char(36)``` for everything that is not Postgres.  Which is not the case.

Just trying help people keep migrations simple and abstract. :)

mamchenkov:
Just checked to be sure, and MysqlAdapter uses ```char(36)``` for UUID:

* [Line 814](https://github.com/robmorgan/phinx/blob/master/src%2FPhinx%2FDb%2FAdapter%2FMysqlAdapter.php#L814)
* [Line 871](https://github.com/robmorgan/phinx/blob/master/src%2FPhinx%2FDb%2FAdapter%2FMysqlAdapter.php#L871)

PR #610, opened 2015-07-26
Schema n seeds

rockpunk:
Hi! 

At Metric Insights, we've been using Phinx to handle migrations of our rather complicated database (360 tables, 60 seed tables, some foreign keys, procedures, views, etc.) However, we needed a clean and easy way to create a base schema including seed data to allow us to install a fresh version of our product on customer machines. (Instead of running through all the migrations, it is easier to create the schema once, and then add any seed data.) 

I know some work on adding a seed() function to migrations, but I was looking for something a little more akin to the Ruby On Rails method of automatically dumping a schema structure on each migration. As a result, I spent the past week working on implementing this in a pull request. The main features and design goals that I wanted were as follows:

* feature1 - automatically dump modified schema structure when a developer runs a new migration
* feature2 - be able to create seed data from an existing database (that may not already have a phinx migration for)
* feature3 - be able to reset the database to a 'known' clean version during development, similar to rail's db:reset functionality
* keep it as simple as possible for the user.  don't add a bunch of unnecessary phinx commands or collections of commands
* intentionally do NOT support views, stored procedures, and functions. This is more like code and should be loaded in after the basic schema structure has been loaded.
* use sql-based (database specific) schema dumps (no ORM)
* support as many database adapters as possible
* make `phinx migrate` continue to work with existing phinx.yml configurations

The result is this pull request. (Sorry it's kinda big! Most of the code is tests) It includes two new user-facing methods:

`phinx reset` - reset the database to the state known in schema.sql
`phinx dump` - dump the current version of the target environment into a schema.sql (this will automatically be called by `phinx migrate` and `phinx rollback`, similar to rails, but can be turned off with the --no-schema-dump option)

I have tested this branch with MySQL, Postgres, and SQLite adapters and all tests pass, including with potentially complicated foreign key dependency schemas. However, I was unable to test SQLServer so I would love for someone to help with that.

I'm also open to feedback, both positive and negative, and would be willing to work with the community more to get this functionality upstream.  There are still some things that could use doing, like:

* integrating with proposed Migration::seed() if necessary
* testing SQLServer
* updated user documentation about working with seeds, handling code like views, procedures, etc.
* a database setup/create method similar to rails?

Thanks for all the good work you guys have done on this project, it was relatively easy to modify!

-Steve

p.s. - if you are running MySQL 5.5 or earlier, and you point the phinx test suite against said database with a lot of innodb tables, you might need to `set global innodb_stats_on_metadata=0;`, otherwise the code dump the schema is pretty horribly slow. See https://www.percona.com/blog/2013/12/03/innodb_stats_on_metadata-slow-queries-information_schema/ for more info. Luckily, innodb_stats_on_metadata defaults to OFF for mysql 5.6 and later.


rockpunk:
Hm... so i had a bout of php 5.3 incompatibilities, which I have verified fixed on a php 5.3 debian squeeze docker image. The latest build, however, seems to die due to a travis configuration error. Php 5.4 build is trying to create postgres user and blocking on user input:

Shall the new user be a super user? (y/n)

I'll commit a dummy commit later to see if the build machine magically fixed itself (because before the 5.4 build built just fine?)

rockpunk:
hey @robmorgan! i've been perusing through the open tickets and i gather that you're working on a seed functionality for 0.5.0. From what little I could piece together, I think you're leaning more to a separate command like `phinx seed` which will seed the database with a local seeder, as well as separate migrations out to include up/down and a separate seed method. 

While that functionality may prove useful for many folks, it's not really what we needed at MI. Hence this PR.

I very much hope that what I've done here is compatible with your anticipated changes as I'd love to see it merged. Would you mind giving it a review when you get the chance? Please let me know if there's anything I need to change to make it work with what you're planning. I'm happy to work with you in whatever way you need to get this functionality upstream! Thank you!

robmorgan:
hi @rockpunk thanks for such a big PR. I've just started skimming over it. First thing I can spot is some of the code is not properly PSR formatted. The other thing - I like the reset + dump commands <--- they have been commonly requested in the past. The seed logic though will need to be expanded on in order to meet everyone's needs. I can probably integrate some of my work in here, I'll keep going over it. Cheers, Rob

rockpunk:
Thanks Rob, yes, you are correct re: psr formatting. I ran everything I wrote through php-cs-fixer so it should be kosher now. (i don't normally write php and only touch it if i have to ;) ) I'm happy to hear that this functionality can likely be expanded on. Please don't hesitate to let me know if you have any questions if/where the code doesn't speak for itself.

I'm looking forward to seeing how the seed logic will be expanded!

-Steve

thomasbiddle:
+1 on being able to dump schema. Looks like I searched for this feature at the perfect time as I'm seeing a lot of active development on it. Great work guys

shadowhand:

PR #622, opened 2015-08-19
allow setting template file or class in the config file

bobmajdakjr:
these updates allow you to define templates and classes in the config file instead of having to use --template and --class all the time from the command line. the same rules about not using both at the same time apply.

templates:
    file: path/to/file
    class: f/q/c/n


bobmajdakjr:
closing because of course i had a better idea. brb.

w2pc:
a better idea.... i have to see this :d I liked this one.

PR #623, opened 2015-08-19
allow setting template file or class in the config file. this time with feeling.

bobmajdakjr:
these updates allow you to define templates and classes in the config file instead of having to use --template and --class all the time from the command line. the same rules about not using both at the same time apply. using --class or --template from the cli will overrule the config file.

    templates:
        file: path/to/file
        class: f\q\c\n

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -243,6 +243,34 @@ class Config implements ConfigInterface
     }
 
     /**
+     * Get the template file name.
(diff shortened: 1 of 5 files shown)


bobmajdakjr:
any thoughts? ^_^

IngwiePhoenix:
Still hoping for this to get in. There is a PR for it after all...

robmorgan:
i'll take it.

PR #624, opened 2015-08-19
Added ability to roll back by date string.

jpohlmann:
I've run into this issue more than once, where I want to roll back my database to my version's release date before I roll the code back.

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -46,12 +46,14 @@ class Migrate extends AbstractCommand
         $this->setName('migrate')
              ->setDescription('Migrate the database')
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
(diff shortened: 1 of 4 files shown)


stoiev:
+ 1 for this pull request.

IngwiePhoenix:
I second that!

w2pc:
I upvote this.

rquadling:
Can you recheck why some of the input value are now optional?

If I use a ```-t``` parameter, I will be expected to supply a timestamp. Similarly, if I use a ```-d``` parameter, I'm expecting a timestamp.

In neither case is the timestamp optional - as that implies no target.

jpohlmann:
So I considered doing that, and I certainly still can.  But currently when I run phinx migrate without a -t, it updates to the current revision.  When I run phinx rollback without the -t, it rolls back one revision.  So it didn't look like -t was actually required.  I didn't want to mess that up by raising an error.

Having said that, I can add that in if it makes sense.

robmorgan:
I like this and thanks for the tests as well.

PR #627, opened 2015-08-24
all migrations now run inside a single transaction

schlndh:
As was explained in robmorgan/phinx#6 all migrations should run in a single transaction - that way either all migrations finish successfully or none will.

I also fixed a rather strange behavior - previously rollback was never called even if a migration failed somewhere which could lead to some changes being commited when error happened outside of SQL query and autocommit was enabled (which is the default behavior for mysql, I don't know about other databases).

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -223,29 +223,46 @@ class Manager
         // are we migrating up or down?
         $direction = $version > $current ? MigrationInterface::UP : MigrationInterface::DOWN;
 
-        if ($direction == MigrationInterface::DOWN) {
-            // run downs first
-            krsort($migrations);
+        $adapter = $this->getEnvironment($environment)->getAdapter();
+        if ($adapter->hasTransactions()) {
+            $adapter->beginTransaction();
+        }
+
+        try {
+            if ($direction == MigrationInterface::DOWN) {
+                // run downs first
+                krsort($migrations);
+                foreach ($migrations as $migration) {
+                    if ($migration->getVersion() <= $version) {
+                        break;
+                    }
+
+                    if (in_array($migration->getVersion(), $versions)) {
+                        $this->executeMigration($environment, $migration, MigrationInterface::DOWN);
+                    }
+                }
+            }
+
+            ksort($migrations);
             foreach ($migrations as $migration) {
-                if ($migration->getVersion() <= $version) {
+                if ($migration->getVersion() > $version) {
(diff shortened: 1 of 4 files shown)


schlndh:
I found out that it is far more complicated than I thought, at least on MySQL (I don't know about others, but it seems that at least PostgreSQL [supports transactional DDL](http://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis)).

MySQL performs implicit commit before (and after) [some commands](http://dev.mysql.com/doc/refman/5.5/en/implicit-commit.html) including `ALTER TABLE` and these therefore cannot be rolled back as well as any query that came before them. (I found a bug in MariaDB bug tracker which suggest that transactional DDL [will be implemented](https://mariadb.atlassian.net/browse/MDEV-4259) for MariaDB).

Also be sure to use database engine that actually supports transaction (eg. not MyISAM).

To sum it up I think that this PR does make things better than they were before but it's far from perfect because of limited support for transactions in MySQL. If you use MySQL you'll have to write all migrations so that they will finish successfully no matter what.

robmorgan:
I am holding off on this PR for the moment until the other adapter database vendors better support transactional DDL.

schlndh:
@robmorgan Thanks for feedback. According to [this](http://stackoverflow.com/a/4736346/2503810) the only one missing (from the list of databases supported by phinx) is MySQL (which could hopefully change  in the future at least for MariaDB).

I guess we could put a warning for MySQL, but since it benefits all other supported databases would you reconsider merging it?

lukecyca:
+1

This is a hugely important feature, as otherwise a failed migration leaves the database in an unknown state with no way to recover! It's disappointing that MySQL can't support it, but that should not prevent this feature from becoming available for Postgres et al.

schlndh:
@lukecyca Thanks. If you try it on Postgre please provide some feedback, I can't really test it in real application as we use MySQL (but we use this feature anyway).

lorenzo:
@robmorgan You will probably sit and wait forever until all vendors implement this feature. In the meantime I think that it is a key feature for people using postgres

shadowhand:
This won't work correctly with MySQL, despite MySQL reporting that it supports transactions. Most likely, a new `hasDefinitionTransactions` method should be added to support this feature.

Also, this PR is currently targeting the 0.4.x-dev branch, which can no longer accept new features. Please create a new PR against the 0.5.x-dev branch.

Thank you!

PR #607, opened 2015-07-22
Added Boolean as one of signed types

darensipes:
For the MysqlAdapter it make sense to me to allow the signed option to work with boolean since MySQL is using tinyint for boolean. I prefer to make my tinyint(1) values unsigned when used for boolean purpose.

PR #630, opened 2015-08-25
Allow Boolean Columns to Accept the Signed Option

### LINKED ISSUE #490: Boolean fields are not showing up as unsigned in MySQL
Given this method call...

```php
$table->addColumn('is_visible', 'boolean', array('length'=>1, 'signed' => false))
```

The column that is created in mysql shows type TINYINT(1) but remains signed, so it is ignoring my 'signed'=>false flag. Expected behavior is that it respects the flag at minimum, and even makes unsigned be the default state for boolean fields. 

The best solution, of course, is to improve the MySQL adaptor's code to translate this "boolean" type to "bit" in MySQL.

kamioftea:
This pull request
- Add boolean to array of allowed signed types in the MySQL Adapter - which seems the only adapter the signed option applies to.
- Adds a test for adding an unsigned boolean
- Update docs to indicate boolean and decimal values can be signed when using the MySQL adapter

It fixes #490.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -924,6 +924,7 @@ Option    Description
 ========= ===========
 precision combine with ``scale`` set to set decimial accuracy
 scale     combine with ``precision`` to set decimial accuracy
+signed    enable or disable the ``unsigned`` option *(only applies to MySQL)*
 ========= ===========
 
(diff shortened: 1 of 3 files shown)


kamioftea:
I missed that PR #607 already existed when writing this, but the code change is exactly the same.

PR #637, opened 2015-09-08
Update symfony console to 3.0

stevemarksmonii:
Update symfony console to 3.0. Tests still passing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     }],
     "require": {
         "php": ">=5.3.2",
-        "symfony/console": "~2.7",
+        "symfony/console": "^3.0@dev",
         "symfony/config": "~2.7",
         "symfony/yaml": "~2.7"
     },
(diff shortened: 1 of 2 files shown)


robmorgan:
unfortunately symfony console 3.0 requires > PHP 5.5 so I can accept this until we drop support for the older PHP versions.

willdurand:
@robmorgan that would be nice actually, because right now I can't use Phinx in my project... I think supporting both `^2.7` and `^3.0` should allow anyone to use PHP 5.3, even though it is more than dead (but I won't argue about that here).

robmorgan:
@willdurand I ran a survey and its clear only a very small percentage are using 5.3 now so I'm going to axe it in the next release. People can continue to use 0.5.x if required. Will write a blog post later.

In other news I can accept this PR now.

robmorgan:
fixed in 7fe73eff94fa65cc6123ec839d3d4ef93667c5de.

Issue #243, opened 2014-05-05
Unable to create log table in mysql


murraytm:
Using phinx 0.3.4.

With a MySQL 5.7.4 server running in STRICT_TRANS_TABLE mode, phinx is unable to create the log table (phinxlog) and exits with an error about an invalid default value for end_time.

The problem is that phinx is trying to make the default value '0000-00-00 00:00:00' on a column with the TIMESTAMP data type.

STRICT_TRANS_TABLE enabled the NO_ZERO_DATE server mode so that is an invalid date.

The correct minimum value for a TIMESTAMP column is '1970-01-01 00:00:01'. When I manually created the table with that default value for end_time, phinx was able to proceed with no problems.

willdurand:
:+1:

willdurand:
Using MySQL, I fixed this issue by monkey patching `PdoAdapter`:

```php
$table = new Table($this->getSchemaTableName(), $options, $this);
$table->addColumn('version', 'biginteger', array('limit' => 14))
    // Fix "SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'start_time'"
    ->addColumn('start_time', 'timestamp', ['default' => 'CURRENT_TIMESTAMP'])
    ->addColumn('end_time', 'timestamp', ['default' => 'CURRENT_TIMESTAMP'])
    ->save();
```

shadowhand:
Have you located where the default value of `0000-00-00 00:00:00` comes from? I can't find anything in the code that suggests it should act this way... no default is set in `createSchemaTable`.

stephenmelrose:
I just got this issue with v0.4.2 and MySQL v5.7.5-m15. I upgraded from v0.3.7. I can confirm downgrading back to v0.3.7 fixes this.

stephenmelrose:
I just got this issue with `v0.4.2.1` and MySQL `v5.7.5-m15`.

I tried downgrading (as I just upgraded), and I got the same issue with `v0.3.7` and `v0.3.8`.

The above fix worked for me.

murraytm:
Of course it's been quite some time since I had to look at this but I'm pretty sure that 0 date value is coming from PHP or PDO itself.

twoixter:
I'm having the same issue. @willdurand solution seems fine to me, why this issue is not closed yet?

robmorgan:
It can be closed, can't take @willdurand solution at the moment due to the MySQL version on Travis.

[...]

PR #638, opened 2015-09-09
Update PdoAdapter.php

Awnage:
Fix for #243 to make sure more recent versions of MySQL are happy without Travis complaining too much.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -421,10 +421,21 @@ abstract class PdoAdapter implements AdapterInterface
             );
 
             $table = new Table($this->getSchemaTableName(), $options, $this);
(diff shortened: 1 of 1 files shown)


Awnage:
lol, Travis fails on PHP 5.3. GOOD! No one should be using it anyway. =)

enygma:
:+1: on this - it's a definite must for future use of Phinx. Otherwise I think quite a few people will have to find another tool... :(

robmorgan:
@Awnage please make sure those arrays use the 5.3 syntax otherwise the old PHPs will still complain (even if kept in a conditional block)

Awnage:
Look good, or would you prefer a single comprehensive patch.

robmorgan:
no it looks okay, its not perfect PSR, so I'll clean it up a bit.

PR #639, opened 2015-09-09
Adding explanation on how to add/drop new records.

gmansilla:


robmorgan:
@gmansilla I don't want to accept this yet, because I'm about to change the signature for inserting data to something like:

```
$data = [
    ['id' => 1, 'name' => 'foo'],
    ['id' => 2, 'name' => 'bar']
];
$table = $this->table('status');
$table->insert($data);
```

Rob

robmorgan:
reworded slightly and fixed in 3119fa8de71e2b83b8c3c58d0f503777c376371d

PR #640, opened 2015-09-10
Using a binary type with the MySQL driver should produce a binary column

HavokInspiration:
Currently, if you define a column as ``binary`` when using the MySQL driver, you get a ``blob``.
The problem is that ``binary`` is a valid type for MySQL and it is currently not possible to get one.

This PR aims to correct this.
A new ``BLOB`` type was added to the constants of the ``AdapterInterface``.
If it is used with other drivers than MySQL, it will produce the same result as using the ``binary`` type.


mamchenkov:
:+1:

lorenzo:
:+1:

dereuromark:
:+1:

robmorgan:
thx

Issue #650, opened 2015-09-26
Undefined Index on migration


jimdoescode:
**Problem**
```
Notice: Undefined index: name in /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php on line 190

Call Stack:
    0.0014     222568   1. {main}() /home/vagrant/project/vendor/robmorgan/phinx/bin/phinx:0
    0.0900    2004912   2. Symfony\Component\Console\Application->run() /home/vagrant/project/vendor/robmorgan/phinx/bin/phinx:28
    0.1101    2349360   3. Phinx\Console\PhinxApplication->doRun() /home/vagrant/project/vendor/symfony/console/Application.php:126
    0.1108    2351128   4. Symfony\Component\Console\Application->doRun() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Console/PhinxApplication.php:80
    0.1109    2352016   5. Symfony\Component\Console\Application->doRunCommand() /home/vagrant/project/vendor/symfony/console/Application.php:195
    0.1110    2352520   6. Symfony\Component\Console\Command\Command->run() /home/vagrant/project/vendor/symfony/console/Application.php:878
    0.1112    2357952   7. Phinx\Console\Command\Migrate->execute() /home/vagrant/project/vendor/symfony/console/Command/Command.php:259
    0.1765    2986120   8. Phinx\Migration\Manager->migrate() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Console/Command/Migrate.php:110
    0.1896    3128984   9. Phinx\Migration\Manager\Environment->getVersions() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Migration/Manager.php:204
    0.1896    3129440  10. Phinx\Migration\Manager\Environment->getAdapter() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php:201
    0.1924    3150472  11. Phinx\Db\Adapter\AdapterFactory->getAdapter() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php:271
    0.2045    3551504  12. Phinx\Db\Adapter\PdoAdapter->__construct() /home/vagrant/project/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterFactory.php:130
[...]

PR #651, opened 2015-09-29
Display error if database name can't be determined.

jimdoescode:
According to issue #650, database name **is** required for some commands (in particular migrate). There is currently no check that this required parameter is actually set. This change would address that issue.


jimdoescode:
Sorry, I thought I'd be able to fix the broken test remotely. I'll fix it when I get home from work tonight.

jimdoescode:
@robmorgan is there something else I need to do to get this merged in? I think it's important to provide some error reporting here.

robmorgan:
improves #650.

Issue #580, opened 2015-07-01
Add method hasIndexByName


laurin1:
So if I want to check to see if an index exists before I attempt to drop it, I can't because this code checks against column names and not key names. That's fine if all of your index are a single column and named the same as the column, but I rarely do that.

If this is expected behavior, then can we add a method called hasIndexKeyName() or something to that effect? I can write the PR for it, but want some feedback first.

laurin1:
Same issue with hasForeignKey()

laurin1:
So there is a dropIndexByName, but not a hasIndexByName. Strange.

laurin1:
Ok, I've forked and written the code for all adapters for this new method. I have never submitted a pull request on Github. So I need to install all database systems (I have MySQL already) and write the unit tests for each of these methods and then submit the pull request?

robmorgan:
Hi @laurin1, yes as you are extending `AdapterInterface` you must provide implementations for all supported DB adapters including Sql Server. If you do not have access to Sql Server you can always test your code using tools like Appveyor.

laurin1:
Will it work with SQL Server Express?

robmorgan:
Yes it will if you install the Sqlsrv PDO driver.

dereuromark:
Looks like this can be closed?

PR #653, opened 2015-10-02
Added method hasIndexByName #580

laurin1:
I added unit tests for this method, but not sure how extensive these need to be.


robmorgan:
thanks @laurin1. we just need to update the docs now.

laurin1:
I don't really understand the docs structure. I search and I don't see where hasIndex is even documented. I can do both of them, but I just need to know where these should go?

In fact, many of these commands are not in the docs - are we only documenting "generally" or do we want every method documented?

Issue #657, opened 2015-10-14
How to make id field unsigned?


Serhioromano:
It always create `id` field as signed. But I am not going to have negative values there so I would like to use unsigned. I do not understand how to do that.

This is what I am doing now but I am not sure it is very elegant way

``` php
$this->table('user', ['id' => FALSE, 'primary_key' => 'id'])
    ->addColumn('id', 'integer', ['null' => TRUE, 'signed' => FALSE, 'identity' => TRUE,  'limit' => MysqlAdapter::INT_REGULAR])
```

jimdoescode:
I agree, it's kind of weird to have signed auto-incrementing id fields. I changed the id column after the create to make it work.

```php
$users = $this->table('users');
$users->addColumn('email', 'string', ['length' => 255])
            ->addColumn('password', 'string', ['length' => 255])
            ->addColumn('address', 'string', ['length' => 255])
            ->addTimestamps()
            ->addIndex('email', ['unique' => true])
            ->create();
$users->changeColumn('id', 'integer', ['signed' => false, 'identity' => true])->update();
```

Serhioromano:
@jimdoescode this looks like more elegant and clean way to do that. But I still believe there have to be an option.

wesleywesley:
Still this is not added. Shouldn't even be an option, but the default.

kristianheljas:
Agreed, just stated using phinx and discovered same problem.

I also agree with @wesleywesley to make id field unsigned by default.

Currently resorting with @jimdoescode hack, which works but shouldn't be nessecary.

piotr-cz:
IMHO it should be unsigned by default.
Unsigned ID doubles the upper limit and is recommended to use in MySQL official [docs](http://dev.mysql.com/doc/refman/5.1/en/example-auto-increment.html)

lukewatts:
I've implemented this in my fork here: https://github.com/lukewatts/phinx/

Just be aware I've also changed the default collation to utf8_unicode_ci in this fork, because it's better for foreign language support. 
To set it back to utf8_general_ci change this line: 
https://github.com/lukewatts/phinx/blob/master/src/Phinx/Db/Adapter/MysqlAdapter.php#L208

The lines to change to set the ids to unsigned by default are: 
https://github.com/lukewatts/phinx/blob/master/src/Phinx/Db/Adapter/MysqlAdapter.php#L219
[...]

PR #658, opened 2015-10-17
Allow to pass table option signed for primary key

Serhioromano:
Discussed in #657.

Primary field by default have to be unsigned because for autoincrement this is only that makes sense. But just in case there is parameter that you can pass to change that.


shadowhand:

If you do recreate this PR, please add tests.

PR #664, opened 2015-10-22
Allow status -v to display start and end times.

rquadling:
Currently, the timestamp shown by ``phinx status`` relates to the creation of the migration. If you want to know when the migration was actually run, you have to jump into the DB.

With this PR, ``phinx status -v`` will show the Started and Finished times from the phinxlog.


robmorgan:
I'll take this as soon as I'm back from holidays. 

Cheers,

Rob

robmorgan:
I've decided rather than add all these extra methods and verbosity bloat, it might be better to just add the start and end times by default

shadowhand:
@robmorgan what's the status on adding start and end times?

rquadling:
I'll be rebasing on the 4th or 5th of Jan.

PR #673, opened 2015-11-08
Redshift Adapter

kyrozetera:
Closes https://github.com/robmorgan/phinx/issues/592

This adds a new Redshift adapter specifically adding support for the `diststyle`, `distkey`, and `sortkey` table options. As well as a bit more validation on Postgres features that Redshift doesn't support.


shadowhand:

PR #681, opened 2015-11-17
Add $version variable to be passed into templates

sebmak:
We see a need to use the $version in the templates to create better backups and restores.


robmorgan:
fixed in 250a3d98ac63e35752db48b249328af6768bd46a

PR #685, opened 2015-11-20
Fixed rollbacks

daniel-gomes-sociomantic:
(as discussed in https://github.com/robmorgan/phinx/issues/628)

When rollbacking migrations, their order is now determined by when they were executed (ie. their Start Time) as opposed to when they were created (ie. their version name).
    
This means:
* With no parameters, the rollback command will rollback the last executed migration.
* With the -d (date) parameter, the migration start times are now used to determine the target version to rollback to. In other words, all migrations which were executed after that date will be rollbacked.
* With the -t (target version) parameter, all migrations which were executed after that version was executed will be rollbacked.

In terms of testing:
* The `testRollbacksByDate` test was changed to be more "unit"y, ie. it now asserts the rollback method is called with the right version (instead of asserting the right version is rollbacked). A few test cases were also added to it (via its provider).
* The `testRollbacksByVersion` was added.

A few notes:
* I based my work on https://github.com/robmorgan/phinx/pull/664 as I needed access to the Migration Start Times, I hope that's fine (the comment on the PR suggests it will be accepted anyway).
* I understand my approach is quite a game changer, as it completely changes the concept of "previous migration" and how the "-d" parameter works, but I couldn't see any other way to solve this.
* Seeing as how the execution times are the "norm" in this approach, the Status command should probably also be changed to order the Up migrations by execution time instead of name, but I couldn't decide on what ordering made sense with regards to up, down and missing migrations, so I didn't change it.

### DIFF
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -74,9 +74,10 @@ interface AdapterInterface
     /**
      * Get all migrated version numbers.
      *
+     * @param bool $fullRow
      * @return array
      */
-    public function getVersions();
+    public function getVersions($fullRow = false);
 
     /**
      * Set adapter configuration options.
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -74,10 +74,10 @@ interface AdapterInterface
     /**
      * Get all migrated version numbers.
      *
-     * @param bool $fullRow
+     * @param bool $fullVersion Get the entire version row, not just the version number.
      * @return array
      */
-    public function getVersions($fullRow = false);
+    public function getVersions($fullVersion = false);
 
     /**
      * Set adapter configuration options.
--- a/src/Phinx/Db/Adapter/AdapterWrapper.php
+++ b/src/Phinx/Db/Adapter/AdapterWrapper.php
@@ -184,9 +184,9 @@ abstract class AdapterWrapper implements AdapterInterface, WrapperInterface
     /**
      * {@inheritdoc}
      */
-    public function getVersions()
+    public function getVersions($fullRow = false)
     {
-        return $this->getAdapter()->getVersions();
+        return $this->getAdapter()->getVersions($fullRow);
     }
 
     /**
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -354,11 +354,11 @@ abstract class PdoAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
-    public function getVersions($fullRow = false)
+    public function getVersions($fullVersion = false)
     {
         $result = array();
         $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY version ASC', $this->getSchemaTableName()));
-        if ($fullRow) {
+        if ($fullVersion) {
             foreach($rows as $v) {
                 $result[$v['version']] = $v;
             }
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -354,15 +354,24 @@ abstract class PdoAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
-    public function getVersions()
+    public function getVersions($fullRow = false)
     {
+        $result = array();
         $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY version ASC', $this->getSchemaTableName()));
-        return array_map(
-            function ($v) {
-                return $v['version'];
-            },
-            $rows
-        );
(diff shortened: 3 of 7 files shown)


rquadling:
A very BIG +:100:

Poor Rob. So many outstanding patches!

I certainly hope this gets accepted. I may need to rework my breakpoint support issue / PR https://github.com/robmorgan/phinx/issues/494 / https://github.com/robmorgan/phinx/pull/497 to work with this feature, though things may just work.

With regard to order, how about ``--order|-o <c|created|e|executed>`` with the default (for the time being for backward compatibility), ``c|created``, but at some stage in the future, the ordering will be ``e|executed`` as that is really the most sensible order. I would keep with the ascending date order as that is also sensible.

voidcontext:
+1 on this... Usually reverting migration by their application order has more sense.

daniel-gomes-sociomantic:
> With regard to order, how about --order|-o <c|created|e|executed> with the default (for the time being for backward compatibility), c|created, but at some stage in the future, the ordering will be e|executed as that is really the most sensible order. I would keep with the ascending date order as that is also sensible.

I'm not sure if you meant this for the Status or for the Rollback command, but if it's for the Rollback, it makes total sense to me in order to keep that backwards compatibility and allow for "unusual" cases. Actually, this could be added to the Status command as well.

rquadling:
I think for status, the order could be exposed to the user, allowing them to keep the default broken order (i.e. migration creation date)  or have the execution date (I would prefer execution date as creation date is useless to me).

But, for BC, the default order is by creation and, at some stage in the future, it could be changed to be by execution.

When the default changes, BC can still be maintained by the user if they so wish.

A win/win I think.

daniel-gomes-sociomantic:
agreed. But maybe it would make sense (also for BC) to do the same for the rollbacking as well?

rquadling:
Personally, I think Rollback is broken/bugged because of the ordering being migration creation date.

So, to fix it, it would have to go to execution order.

But that would be a significant BC.

__MAYBE__ add a config entry (default_version_ordering or something like that), which, if not set, matches the current behaviour.

Can ALSO be overridden at the command line (like default template/class - though still waiting for https://github.com/robmorgan/phinx/pull/642 to be accepted on overrides).

So, for those that do nothing (no config/no command line), things work as is.
For those who use the command line, they can temporarily use the new mechanics.
For those who set the config, they use the new mechanics all the time unless they decide to override it at the command line.

daniel-gomes-sociomantic:
> Personally, I think Rollback is broken/bugged because of the ordering being migration creation date.
> 
> So, to fix it, it would have to go to execution order.
> 
> But that would be a significant BC.

@robmorgan can we get your opinion here? Should the rollback command sort the migrations by execution time by default, or should that be enabled via a `--order|-o <c|created|e|executed>` flag?

Another thing, shall I rebase this onto https://github.com/robmorgan/phinx/pull/699 (after @rquadling's changes)? If so, I would also set the PR to merge with `0.5.x-dev`.

rquadling:
@daniel-gomes-sociomantic  I'm just finishing the testing on #699 as there are some failing unit tests

shadowhand:

I would really like to see this added to Phinx as an option. Our current implementation of migrations would benefit greatly from this approach.

rquadling:
I think for the current version, all changes to how rollbacks are processed has to be optional, sure. But I think, in a future version, along with a considerable notice period to existing users, the default should be to rollback based upon execution time.

daniel-gomes-sociomantic:
> I would really like to see this added to Phinx as an option. Our current implementation of migrations would benefit greatly from this approach.

So I'll add (as suggested by @rquadling ) a `--order|-o <c|created|e|executed>` parameter with:

* `-c|created`: rollbacks are done by ordering the migrations by creation time (the default, ie. the current behaviour).
* `-e|executed`: rollbacks are done by ordering the migrations by start time.

Does that look fine @shadowhand?

daniel-gomes-sociomantic:
also @rquadling please lemme know when you rebase https://github.com/robmorgan/phinx/pull/664 so I can base this on your contribution

rquadling:
As requested by Rob (https://github.com/robmorgan/phinx/pull/699#issuecomment-161630374), we aren't going to use a ``-v`` option. Instead rebase upon https://github.com/robmorgan/phinx/pull/699

daniel-gomes-sociomantic:
@rquadling so your `feature/VerboseStatus` branch, correct? (just double checking)

rquadling:
@daniel-gomes-sociomantic That's the one.

daniel-gomes-sociomantic:
For simplicity, I'll reduce the possible values to `creation` and `execution`. I updated my original comment.

daniel-gomes-sociomantic:
@shadowhand actually I don't quite like to use free-text values for options which have a well-known (and pre-defined) set of valid values like our new `--order` option... As I suppose there won't ever be more than two ways to order the rollbacks, maybe we can instead use a `-s|--start-time` switch to enable rollbacks by start time?

daniel-gomes-sociomantic:
replaced with https://github.com/robmorgan/phinx/pull/745

PR #689, opened 2015-11-24
tests should be green on php7

AlexDpy:


AlexDpy:
All tests are now OK.
Shall we remove php7 from the "allow_failures" section ?

PR #695, opened 2015-11-30
Cleanup code / CS

dereuromark:
Cleanup after https://github.com/cakephp/migrations/pull/161

- Remove wrong returns when void
- Make stricter comparisons-
- No break on default in switch
- Consolidate functions and remove aliases in favor of actual function names


robmorgan:
thx

dereuromark:
I also highly recommend to
- remove yoda conditions and instead sniff on conditional assignment in general (prevents 100% instead of a small subset of cases)
- remove inline alignment and other commit noise generators like that.

I got sniffers/fixers for that in case you are interested in that cleanup. :)

PR #699, opened 2015-12-03
Feature/verbose status

rquadling:
Tests also force disable ANSI output. On Windows, ANSI support is enabled if ANSICON is loaded. And, seemingly, for Windows 10, the CMD /D /K option no longer disabled the autoloading of ANSICON, so all shells have it, therefore need to disable it in the tests. Argh!!!

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -217,11 +217,12 @@ class Manager
         });
 
         if (count($availableMigrations) > 0) {
-            $migration = min($availableMigrations);
             if (is_null($earlierVersion)) {
                 $this->getOutput()->writeln('Rolling back all migrations');
+                $migration = 0;
             } else {
                 $this->getOutput()->writeln('Rolling back to version ' . $earlierVersion);
+                $migration = $earlierVersion;
             }
             $this->rollback($environment, $migration);
         }
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -101,8 +101,8 @@ class Manager
             // TODO - rewrite using Symfony Table Helper as we already have this library
             // included and it will fix formatting issues (e.g drawing the lines)
             $output->writeln('');
-            $output->writeln(' Status  Migration ID    Migration Name ');
-            $output->writeln('-----------------------------------------');
+            $output->writeln(' Status  Migration ID    Started              Finished             Migration Name ');
+            $output->writeln('----------------------------------------------------------------------------------');
 
             $env = $this->getEnvironment($environment);
             $versions = $env->getVersionLog();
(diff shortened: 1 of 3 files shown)


robmorgan:
Hi @rquadling, I want to make this the default on the status command, regardless of the verbosity flag. 

1. Can you remove the parameters to the 'getVersions()' methods?
2. Can you make the start & end times always displayed?
3. Cleanup tests after those changes

Cheers,

Rob

rquadling:
Sure.

shadowhand:
Was this replaced by #712?

rquadling:
No. Not completely. My usual issue of adding too much to a PR/commit. Sorry. All my outstanding PRs need to be rebased. Will do that on Jan 4th or 5th.

rquadling:
Ah. Because testing fails on Windows+ANSICON, the decoration fix is present in this PR. Or at least was as the rebase should have taken care of it.

daniel-gomes-sociomantic:
@rquadling your PR seems to have a bug when running a second migration after having run a first one.

To reproduce:

1) Create a new Migration and run status:

```
 Status  Migration ID    Migration Name 
-----------------------------------------
   down  20151117182005  FirstMigration
```

2) Migrate (which works fine):

```
== 20151117182005 FirstMigration: migrating
== 20151117182005 FirstMigration: migrated 0.0721s
```

3) Add another migration and run status:

```
Status  Migration ID    Migration Name 
-----------------------------------------
     up  20151117182005  FirstMigration
   down  20160116122005  SomeOtherMigration
```

4) Try migrating:

```
== 20151117182005 FirstMigration: migrating

[PDOException]                                                                                            
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '20151117182005' for key 'PRIMARY'                                                                                          
```

Basically it tries to execute the `FirstMigration`, even though it was already executed before and is up (as the status confirms).

The same scenario works fine in the current phinx `master` branch (5efc456).

rquadling:
Hmmm. Did you run unit tests against your live database? I did that a couple of times and wondered why I had an empty phinxlog in my DB!!!

daniel-gomes-sociomantic:
I have a small test application where I manually tried, running on an empty database. The `phinxlog` table is also empty before the test, but that's expected (I double-checked manually the table in between steps).

rquadling:
Well this looks buggered! Same issue locally. I wonder if I've missed a call to ``getVersions()`` in recent code and that expects the old style integer version and is getting the full version info.

Hold on until I find'n'fix it.

daniel-gomes-sociomantic:
Great, cheers! I would also suggest a test to prevent this from happening in the future again :-)

rquadling:
Issue SEEMS to be in ``\Phinx\Migration\Manager::migrate()`` ... ``$versions = array_keys($env->getVersions());`` should be the line. Stupid me!!! And the same in ``\Phinx\Migration\Manager\Environment::getCurrentVersion()``.

Really sorry about this!!!

This is why I had getVersions() (untouched) and getFullVersions() (with full breakdown).

Then moving to getVersions() being full versions ...

Fix breaks testCurrentVersion() which is looking at the old result format.

Fixing that too.

rquadling:
I'm tempted to reintroduce the $fullVersion parameter as that worked well for me.

@robmorgan Ideas?

Obviously, I buggered this up, but now we have 2 common usages of getVersions(). One (and most commonly use case) is just the version IDs, and the other is the full version data.

I think it is sensible having a switch rather than always getting all and the then immediately filtering them in the calling code.

daniel-gomes-sociomantic:
makes sense to me.

Anyway, I've finished my Rollback feature but can't move forward until this is finalized and hopefully even merged, so some attention from the nice owner/collaborators would be nice :-)

rquadling:
Going to refactor based upon work done by Andy Coates in https://github.com/robmorgan/phinx/pull/744. Much less intrusive changes.

daniel-gomes-sociomantic:
hey @rquadling I think I found another bug:

start with these 2 migrations:

```
 Status  Migration ID    Started              Finished             Migration Name 
     up  20151117182005  2016-01-11 13:36:13  2016-01-11 13:36:13  FirstMigration
     up  20160116122005  2016-01-11 13:36:13  2016-01-11 13:36:13  SomeOtherMigration
```

Running a rollback to a date before both of the migrations only rollbacks one of the migrations:

```
> phinx rollback -d 20140101

 == 20160116122005 SomeOtherMigration: reverting
 == 20160116122005 SomeOtherMigration: reverted 0.0904s

 > phinx status

 Status  Migration ID    Started              Finished             Migration Name 
     up  20151117182005  2016-01-11 13:36:13  2016-01-11 13:36:13  FirstMigration
   down  20160116122005                                            SomeOtherMigration
```

This can be easily solved by adding a `$migration = 0` line after the `$this->getOutput()->writeln('Rolling back all migrations');` one (to do a rollback to version 0, ie. rollback all versions)

(by the way, the same bug occurs in the master branch so I guess it wasn't introduced by you, but since you altered the way the rollback is done, I reckon you might as well fix it :-) )

rquadling:
Yep. Upgraded unit tests to cover rolling back to a date, where the date is is:
1. Before the first migration - 2 migrations rolled back.
2. The first migration - 1 migration rolled back.
3. Between 2 migrations - 1 migration rolled back.
4. The last migration - no migrations rolled back.
5. After the last migration - no migrations rolled back.

rquadling:
Rebased to 5.2. What is required/outstanding, stopping this from being merged?

rquadling:
Thank you.

Issue #706, opened 2015-12-05
Define timestamp columns via arguments


samizdam:
Now it's hardcoded as 'created_at' and 'updated_at' at https://github.com/robmorgan/phinx/blob/0.5.x-dev/src/Phinx/Db/Table.php#L557 
`
Table::addTimestamps($createdAtColumnName = 'created_at', $updatedAtColumnName = 'updated_at')
`
more flexible.

dereuromark:
Are you able to provide a pull request with your suggested changes?

Issue #576, opened 2015-06-23
Add option to customize timestamps field names


topikito:
When calling method `addTimestamps()`, the fields are named by default (hardcoded). Would be ideal to have a way to customize the names of these fields to match each projects own conventions.

As a secondary nice to have, I'd say to specify the type of that field, as you may use native MySQL timestamp or perhaps want to use integer to match php `time()`.

danielfh:
+1

samizdam:
Hm, when I try find this feature, I don't see this issue and create #706  and implement with #707 
So, +1)

dereuromark:
Seems implemented according to 2nd link.

PR #707, opened 2015-12-05
Add arguments for timestamps columns names.

samizdam:
Implement #576 && #706


samizdam:
I don't sure, but this feature will be applied only after phar rebuilding? 
Do I must push with pull request binary file? 
Contributing FAQ not provide some info about it...

shadowhand:
@samizdam please add tests.

samizdam:
@shadowhand  this method is already covered. 
I'm update my branch with upstream now. 

table-coverage

robmorgan:
I dont really think we need this feature. If people want to change types etc or column names I generally tell them just to use `addColumn()` instead. I'd rather leave this method simple like Rails.

shadowhand:
I don't see any problem with having defaults that can be modified rather than hard coded. Seems like a reasonable approach to me.

rquadling:
I've implemented this feature, along with appropriate guarding of null (you cannot have a column called null - who'd have thought that!) and unit tests.

PR #712, opened 2015-12-11
Stop decorating output. This affects Windows users running ANSICON.

rquadling:
Will probably also affect ConEMU and other tools that add ANSI colouring automatically (i.e. without a --ansi option) that cannot (seemingly) be disabled when running phpunit.

I based this (as requested) on the tag v0.5.0 and not 0.5.x-dev. But seeing the "Can't automatically merge".

No idea how to handle that.

### DIFF
--- a/tests/Phinx/Console/Command/InitTest.php
+++ b/tests/Phinx/Console/Command/InitTest.php
@@ -31,6 +31,8 @@ class InitTest extends \PHPUnit_Framework_TestCase
         $commandTester->execute(array(
             'command' => $command->getName(),
             'path' => sys_get_temp_dir()
+        ), array(
+            'decorated' => false
         ));
 
         $this->assertRegExp(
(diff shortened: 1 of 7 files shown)


robmorgan:
@rquadling could you please rebase this with `0.5.x-dev` again? We need to fix the conflicts from recent status command changes.

rquadling:
Failures don't seem to be related to the patch.

PHP 5.4 isn't tested with Symfony 2.8.

Line 29 of .travis.yml says PHP 5.3, but compose.json line 20 requires PHP >= 5.4.

I'm not a travis expert, so the above may be incorrect.

shadowhand:
Thanks! In the future please follow [these rules](http://chris.beams.io/posts/git-commit/) when making PRs for this project. Setting up a tool like [static-review](https://github.com/sjparkinson/static-review) can help you verify your commits.

PR #719, opened 2015-12-15
added priority seeds

visavi:
added to the priority of the files
enough to register `public $priority = 50;`
before
public function run()

Regarding array_reverse to 627 line, in theory you could do without it, and change a little higher 
`return ($a->priority > $b->priority) ? -1 : 1;`

but then lost the natural sorting by file name

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -639,8 +639,14 @@ class Manager
                 }
             }
 
-            ksort($seeds);
-            $this->setSeeds($seeds);
+            uasort($seeds, function ($a, $b) {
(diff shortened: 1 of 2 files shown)


shadowhand:
Please rebase and add unit tests.

robmorgan:
I'm currently on the fence over whether to add this feature or take the call method instead. I.e seeders can call other seeders

robmorgan:
I might add a seperate issue for `call()` support so we can gather feedback from the community.

antonienko:
I've tried a different approach in [this pull request](https://github.com/robmorgan/phinx/pull/790)

PR #725, opened 2015-12-19
Add a documentation example for addForeignKey() with array params

chris48s:
I had to read the source to work out how to do this, so it would be helpful to include an example in the docs.


shadowhand:
Thanks!

PR #727, opened 2015-12-21
vendor/bin contains executable/runnables.

rquadling:
``vendor/bin`` contains executable/runnables. Using ``php`` to launch them is incorrect.


robmorgan:
thanks

PR #732, opened 2015-12-24
PHP 5.3 is no longer supported by phinx.

rquadling:
As composer.json says PHP must be >= 5.4, no point in testing against PHP 5.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ before_script:
     - psql -c 'create database phinx_testing;' -U postgres
 language: php
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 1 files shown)


shadowhand:
I've included this change with #733.

shadowhand:
Actually I'm just going to merge this. And rebase my patch. Thanks for contributing!

rquadling:
Travis is using PHPUnit V5 for PHP5.6 but I can't seem to get that dependency.

Killing off composer.lock and deleting vendor, still gives me PHPUnit 4.8.24

Any ideas? I'm trying to find the cause of the failing tests which have nothing to do with the code being amended. No failure locally, just on Travis.

PR #733, opened 2015-12-30
Clean up composer dependencies and travis config

shadowhand:
Having a `composer.lock` makes it harder stay current with dependencies,
so we remove it and prevent it from being added with `.gitignore`. This
was also locking in versions that caused Travis builds to fail.

Instead of explicitly testing Symfony 2.8 components, enable a "lowest
version" test on PHP 5.4 to prove that dependency versions are usable.

Refs #732


robmorgan:
thanks @shadowhand. This is what I should of done. Turns out its the same approach other projects are taking.

PR #734, opened 2016-01-04
Stop decorating output

rquadling:
Stop Symfony Console from decorating output when running unit tests and testing for specific content in the output.


shadowhand:
Thanks!

PR #736, opened 2016-01-05
Capturing migrations with GLOB_BRACE

wandersonwhcr:
I just enable `GLOB_BRACE` to capture migrations in multiple directories.

Example:

```php
<?php
return [
    'paths' => [
        'migrations' => '%%PHINX_CONFIG_DIR%%/module/Module{A,B}/migrations',
    ],
];
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -75,6 +75,14 @@ You can also use the ``%%PHINX_CONFIG_DIR%%`` token in your path.
     paths:
         migrations: %%PHINX_CONFIG_DIR%%/your/relative/path
 
+Migrations are captured with ``glob``, so you can define a pattern for multiple
+directories.
+
+.. code-block:: yaml
+
+    paths:
+        migrations: %%PHINX_CONFIG_DIR%%/module/*/{data,scripts}/migrations
+
 Custom Migration Base
 ---------------------
 
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -515,7 +515,7 @@ class Manager
     {
         if (null === $this->migrations) {
             $config = $this->getConfig();
-            $phpFiles = glob($config->getMigrationPath() . DIRECTORY_SEPARATOR . '*.php');
+            $phpFiles = glob($config->getMigrationPath() . DIRECTORY_SEPARATOR . '*.php', GLOB_BRACE);
 
             // filter the files to only get the ones that match our naming scheme
             $fileNames = array();


robmorgan:
ideally we should document this or allow the migrations string to be an array (i.e. support multiple paths).

wandersonwhcr:
I just added some lines in docs about `glob` pattern usage. :+1:

wandersonwhcr:
I don't know but this code may break `phinx create`. I'll check it.

shadowhand:
:+1: this is a simple change that makes the current config syntax more powerful. I like it.

wandersonwhcr:
@shadowhand do you test the `phinx create` funcionality with this? :confused:

rquadling:
This completely breaks ``phinx create``.

Using a migration path of ``$config['paths']['migrations'] = 'D:/modules/*/{data,schema}/migrations';`` results in ...

image

wandersonwhcr:
@rquadling I'm trying to fix this right now. I'll select the first directory captured from glob.

damn :cry:

rquadling:
I'd ask for the docs to be reverted until the feature is ready.

rquadling:
Also, wildcard directories is supported via ``glob()`` without the ``GLOB_BRACE`` option. Ideally, this should be prevented until the feature is ready - though, if the feature is ready and accepted soon, it would be just wasted churn.

wandersonwhcr:
@rquadling thanks. I just warned about this at comment above.

@shadowhand plz, can you revert as @rquadling said?

wandersonwhcr:
@rquadling do you remember if there is some other `phinx` features where `glob` can break?

wandersonwhcr:
Damn! I'll create a patch to **disable** `glob`. There's some weird behaviors that I found.

rquadling:
``glob`` is used a couple of times. In all instances, a wildcard in the path would be expanded and result in multiple paths when, potentially, a single path is expected.

Maybe glob should only be used with a pre-validated path.

wandersonwhcr:
Currently, `glob` is used 3 times:

```
$ grep -rin glob src
src/Phinx/Util/Util.php:73:        $phpFiles = glob($path . DIRECTORY_SEPARATOR . '*.php');
src/Phinx/Migration/Manager.php:518:            $phpFiles = glob($config->getMigrationPath() . DIRECTORY_SEPARATOR . '*.php', GLOB_BRACE);
src/Phinx/Migration/Manager.php:602:            $phpFiles = glob($config->getSeedPath() . DIRECTORY_SEPARATOR . '*.php');
```

But if we add `glob` to capture all paths at `src/Phinx/Console/Command/Create.php:95` and `glob` doesn't expand (there's no directories for pattern), I don't know how to fix that, because there's no directory to create at line 97.

PR #741, opened 2016-01-07
Add postgres network types

mvrhov:


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -746,6 +750,9 @@ public function getSqlType($type, $limit = null)
             case static::PHINX_TYPE_JSON:
             case static::PHINX_TYPE_JSONB:
             case static::PHINX_TYPE_UUID:
+            case static::PG_TYPE_CIDR:
+            case static::PG_TYPE_INET:
+            case static::PG_TYPE_MACADDR:
```
shadowhand: We really try to avoid these whenever possible. The JSON(B) types are default types, even though not all adapters support them. I'd rather see the same done here.

mvrhov:
Time to merge?

Hadrien45:
+1

mvrhov:
Can we have this merged after a year please? (And I'll have another one shortly adding new types)
It's really annoying that simple PRs are not merged almost immediately

robmorgan:
@mvrhov Thanks for your contribution. If I'm not mistaken these are still Postgres specific types. We also need to see further community support before we accept this one.

mvrhov:
Yes those are PG specific. What community support? These types are standard PG types.

robmorgan:
@mvrhov personally I use some of these custom types in a project however we won't support them in Phinx if only two of our users are requesting them. We are also more likely to support types that are not specific to one particular database system. I'm happy to keep this PR open though in case it becomes more popular.

mvrhov:
This is really unfortunate. Because this means that there is a lot of private forks because PR are open a long time.

MarcusSchwarz:
Well, here is my vote for anything psql-related

JayPHP:
Would be good to do something with this PR - @lorenzo are we still going with the voting process?

lorenzo:
@JayPHP no, I think it is good to have the new types in regardless of the votes

lorenzo:
@mamchenkov do you mind adding a dummy commit to your branch to trigger new builds in the CI systems?

lorenzo:
Actually, never mind. I'm merging it

mvrhov:
Thanks

lorenzo:
Thanks for contributing @mvrhov

PR #744, opened 2016-01-08
Show names of missing migrations in status log

### LINKED ISSUE #543: Possibility of including a 'name' column in phinxlog
One change that would make life easier when working on a project with lots of branches and migrations would be to include the name of the migration in the phinxlog table.

I appreciate that in an ideal world we'd always remember to rollback any migrations relating to a branch of code before switching branches, but in a busy development environment, this isn't always realistic!

This results in often spending a fair bit of time trying to track down which branch contains a migration which is still "up" but missing:
List of missing migrations
If this list gave away at least a migration name, it would make tracking them down much easier.

I'd be happy to provide a PR for this enhancement if meets approval.

coatesap:
Resolves #543 

To reduce the time it takes to locate which branch of code contains a missing migration, all migration names are now stored in the phinxlog table. These are now used as a fall-back reference when the status command is run.

**Example**
This image shows how even though this branch doesn't currently contain a previously migrated migration, the name still shows up in the status log:
migration-screenshot

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -183,9 +183,9 @@ abstract class PdoAdapter implements AdapterInterface
             $this->createSchemaTable();
         } else {
             $table = new Table($this->getSchemaTableName(), array(), $this);
-            if (!$table->hasColumn('name')) {
+            if (!$table->hasColumn('migration_name')) {
                 $table
-                    ->addColumn('name', 'string', array('limit' => 100))
+                    ->addColumn('migration_name', 'string', array('limit' => 100))
                     ->save();
             }
         }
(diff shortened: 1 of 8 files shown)


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -425,11 +444,13 @@ public function createSchemaTable()
                 $table->addColumn('version', 'biginteger', array('limit' => 14))
                       ->addColumn('start_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
                       ->addColumn('end_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
+                      ->addColumn('migration_name', 'string', array('limit' => 100))
```
shadowhand: I'd like to see this column placed after `version` instead of at the end of the table.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -354,13 +361,23 @@ public function insert(Table $table, $row)
...
+    {
+        $result = array();
         $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY version ASC', $this->getSchemaTableName()));
-        return array_map(
-            function ($v) {
-                return $v['version'];
-            },
-            $rows
-        );
+        foreach($rows as $version) {
```
robmorgan: need a space after `foreach` for PSR

src/Phinx/Db/Adapter/PdoAdapter.php
coatesap: Makes sense. I'll do that shortly.

coatesap:
@shadowhand - The new 'migration_name' column has now been moved directly after 'version', as requested.

robmorgan:
@coatesap thanks.. very well written code!

PR #745, opened 2016-01-11
Added option to allow for Rollbacks by start time

daniel-gomes-sociomantic:
When rollbacking, it is now possible to use a new `-s|--start-time` option to order the migrations by start time (ie. execution time) as opposed to the default behaviour (where they are ordered by creation time).

This means:
    
- With only the `-s|--start-time` option, the rollback command will rollback the last executed migration.
- With the `-s|--start-time` and `-d (date)` options, the migration start times will be used to determine the target version to rollback to. In other words, all migrations which were executed after that date will be rollbacked.
- With the `-s|--start-time` and `-t (target version)` options, all migrations which were executed after the target version was executed will be rollbacked.

Instead of using the `-s|--start-time` option in the Rollback command, the `rollback_order` configuration can also be used with the `start-time` value to order migrations by execution time when rolling back.

In terms of testing:

- More test cases were added.

Other notes:
- I based my work on #699 as I needed access to the Migration Start Times.
- Instead of altering the the original `rollback` and `rollbackToDateTime` methods to implement the new option, I chose to leave them intact and code the rollbacking by start time in two new methods (one for rollbacking to target date and one to target version). Even though this might mean some duplication of logic, I did this because it makes the resulting code much simpler and also easier to change in the future, should it be wanted that the current "rollback by creation date" behaviour (the default) be removed.
- The `-s|--start-time` option should probably also be added to the Status command.


rquadling:
A couple of thoughts - take them or leave them.
1. Having ``--target`` and ``--date`` seems redundant. ``--target`` is completely within ``--date``.
2. Could ``rollback`` and ``rollbackToDateTime`` be merged?
3. The ordering could be done as part of ``getVersionLog(string $orderBy = 'version')`` where $orderBy can be ``'version'`` or ``'start_time'``. This option may need propagating through ``getVersions()`` also.

daniel-gomes-sociomantic:
> Having --target and --date seems redundant. --target is completely within --date.

These options behave differently, as I explained in the description: assuming that `-s|--start-time` is being used, the `-d (date)` option will rollback all migrations which were executed **after that target date** whereas the `-t (target version)` will rollback all migrations which were executed **after the target version was executed**, so I wouldn't say they are redundant. Having said that, I'm sure about the necessity to rollback to a certain target version while ordering the migrations by start time.

> Could rollback and rollbackToDateTime be merged?

Maybe, but I chose not to touch these methods at all (the changes here to those methods come from the commits in your PR :-) ).

> The ordering could be done as part of getVersionLog(string $orderBy = 'version') where $orderBy can be 'version' or 'start_time'. This option may need propagating through getVersions() also.

Good catch, it does simplify the code so I'll add it together with the order direction as parameters.
As for the `getVersions`, my changes don't use it so I won't be touching it.

daniel-gomes-sociomantic:
anything I can do regarding the Scrutinizer failure? It doesn't seem to fail on account of the code

rquadling:
I don't know what is erroring with Scrutinizer. It just doesn't work sometimes. It seems.

daniel-gomes-sociomantic:
Was also thinking that it might be worth adding a configuration file parameter to change the default behaviour from order by creation time (the default) to order by start time so people like me won't have to always use the `-s|--start-time` parameter

rquadling:
Yep. Add a @todo to reverse the setting in a major release, so that, by default, rollback is in execution order, not migration creation order.

daniel-gomes-sociomantic:
Just added configuration option for default rollback order and updated the original PR description to reflect this

daniel-gomes-sociomantic:
rebased onto `0.5.x-dev` and squashed the commits around a bit to make it simpler (after the post-rebase changes).

I saw [another](https://github.com/robmorgan/phinx/pull/699) PR getting rebased and merged into `v0.5.2` so initially I rebased onto that branch, but it said it had conflicts with the base branch so I re-rebased onto `0.5.x-dev`, hope that's what's needed (not quite sure what's the correct approach here, my [other](https://github.com/robmorgan/phinx/pull/746) PR is currently based on `v0.5.2`).

rquadling:
I rebase onto the branch head, not a tag.

daniel-gomes-sociomantic:
ups, will re-rebase my other PR :-)

daniel-gomes-sociomantic:
I find it very discouraging that after 3 months no action has been taken on this PR. At the very least, I would expect a comment on what can/should be improved, or whether or not there is any plan to merge this

rquadling:
Yep. Even some bug fixes, with full tests, ... ⌚ 👀

JCMais:
Why is that PR being ignored?

rquadling:
Hi. As some may know I've recently been granted collaborator access to this project. As such, things take time to be added. I'm still getting my 'be responsible and don't deploy everything instantly' hat in place. A key aspect is that any signature changes require a minor version bump for SemVer. Without that, we could deploy changes that breaks someone's use case. And we can't do that.

So. Nothing is ignored. I'll be taking PRs with milestones as I see them.

As this is a volunteer project, I won't be setting (any more) timescales, but I am active and things will get merged into dev branches, allowing everyone a better chance to use and find problems.

Thank you for your patience.

rquadling:
Please rebase against 0.6.x-dev.

rquadling:
A suggestion for `phinx status`` Could a highlight be added to the column that is used for the ordering.

Maybe the `rollback_order` in the config could be `version_order`, allowing for the ordering to be used logically for `phinx status` also.

For the version ordering, if the key to the result was the ID or the execution date, then the single rollback COULD be used AND only a `--target` would be needed. Especially if the logic was merged, allowing something like `--target 201604` . rollback to the beginning of April 2016. Either by version ID or by execution date, depending upon the setting of `version_order` (currently `rollback_order` within this branch).

I think this branch is pretty important and it will go out, but I do want to explore every option in keeping the interface as consistent as possible.

Essentially, as I see it, once I've set the `version_order` I don't really care about what I'm targetting.

If possible, it would be nice to be able to say `--target yesterday` or `--target 3pm` sort of thing and have things be worked out appropriately. Especially if you've released and deployed and need to rollback quickly.

In conjunction with the breakpoint feature, `--target 0` is very useful now.

So. This feature is important, but I feel it still requires some tweaking to not have the mechanism exposed in the way it is. Once the ordering is set, I've made my intent clear. `--target` should operate on that setting appropriately without the need for addition flags.

daniel-gomes-sociomantic:
replaced by https://github.com/robmorgan/phinx/pull/901

Issue #567, opened 2015-06-12
Add a verbose dry-run feature


stephenlawrence:
All of our database changes must be approved by our DB guys. In order to assist with this it would be helpful if we could run a "dry-run" command that would output the SQL that would have been executed during the migration run.

elliotwms:
That sounds interesting. Perhaps a command-line argument like `--pretend`

robmorgan:
something like this will be included in the future. We need to log all SQL statements first then add a command to simulate.

robmorgan:
no timeline yet though

daniel-gomes-sociomantic:
+1

rquadling:
+1 for --dry-run. Matches the same expectation with many other Symfony console based applications.

daniel-gomes-sociomantic:
implemented it in https://github.com/robmorgan/phinx/pull/746

berarma:
+1

This feature would be a big timer saver. Right now debugging non-trivial migrations is a serious pain.

Rob, what's your take on this? Would it help that we tested Daniel's pull request?

dereuromark:
Are you able to provide a pull request with your suggested changes?

berarma:
There's already a PR that got closed (#746) because another one doing the same (#1127) was already merged. I think this issue can be closed.

PR #746, opened 2016-01-12
Implemented Dry Runs

daniel-gomes-sociomantic:
A new `--dry-run` option was added to the Migrate and Rollback commands, which causes the executed/rollbacked Migrations Code to be outputted instead of executed.

Technically speaking, I implemented it in a way that the new option is passed from the Migrate and Rollback Commands to the `Manager` class, whose `executeMigration` method was updated to print a different text header for the migrations being executed/rollbacked when dry-running. The parameter is then passed on to the `Environment` class' `executeMigration` method which enables dry-running in the `PDOAdapter`. Finally, when a statement is executed via the `PDOAdapter.execute` method, if dry-running was enabled it simply prints the SQL code and returns, instead of actually executing it via the underlying database connection.

For #567 and based on https://github.com/wheniwork/phinx/pull/1

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -57,6 +57,10 @@ The <info>migrate</info> command runs all available migrations, optionally up to
 <info>phinx migrate -e development -d 20110103</info>
 <info>phinx migrate -e development -v</info>
 
+The <info>--dry-run</info> option will output the SQL code of the migration(s) which would be executed:
+
+<info>phinx migrate -e development --dry-run</info>
+
 EOT
              );
     }
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -47,7 +47,7 @@ class Migrate extends AbstractCommand
              ->setDescription('Migrate the database')
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
              ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')
-             ->addOption('--dry-run', '-r', InputOption::VALUE_NONE, 'Output SQL rather than executing it')
+             ->addOption('--dry-run', null, InputOption::VALUE_NONE, 'Output SQL rather than executing it')
              ->setHelp(
 <<<EOT
 The <info>migrate</info> command runs all available migrations, optionally up to a specific version
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -47,6 +47,7 @@ class Migrate extends AbstractCommand
              ->setDescription('Migrate the database')
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
              ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')
+             ->addOption('--dry-run', '-r', InputOption::VALUE_NONE, 'Output SQL rather than executing it')
              ->setHelp(
 <<<EOT
 The <info>migrate</info> command runs all available migrations, optionally up to a specific version
@@ -74,6 +75,7 @@ EOT
         $version     = $input->getOption('target');
         $environment = $input->getOption('environment');
         $date        = $input->getOption('date');
+        $dryRun      = $input->getOption('dry-run');
 
         if (null === $environment) {
             $environment = $this->getConfig()->getDefaultEnvironment();
@@ -104,13 +106,17 @@ EOT
         if (isset($envOptions['table_suffix'])) {
             $output->writeln('<info>using table suffix</info> ' . $envOptions['table_suffix']);
         }
+        
+        if ($dryRun) {
+            $output->writeln('<info>Doing dry run</info>');
+        }
 
         // run the migrations
         $start = microtime(true);
         if (null !== $date) {
-            $this->getManager()->migrateToDateTime($environment, new \DateTime($date));
+            $this->getManager()->migrateToDateTime($environment, new \DateTime($date), $dryRun);
         } else {
-            $this->getManager()->migrate($environment, $version);
+            $this->getManager()->migrate($environment, $version, $dryRun);
         }
         $end = microtime(true);
 
--- a/src/Phinx/Console/Command/Rollback.php
+++ b/src/Phinx/Console/Command/Rollback.php
@@ -49,7 +49,7 @@ class Rollback extends AbstractCommand
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to rollback to')
              ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to rollback to')
              ->addOption('--force', '-f', InputOption::VALUE_NONE, 'Force rollback to ignore breakpoints')
-             ->addOption('--dry-run', '-r', InputOption::VALUE_NONE, 'Output SQL rather than executing it')
+             ->addOption('--dry-run', null, InputOption::VALUE_NONE, 'Output SQL rather than executing it')
              ->setHelp(
 <<<EOT
 The <info>rollback</info> command reverts the last migration, or optionally up to a specific version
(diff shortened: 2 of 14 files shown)


rquadling:
Could the option be just ``--dry-run`` without a short version?

rquadling:
With regard to alternative templates (``--template``, ``--class``), getting hold of the dry run option is a bit difficult.

The ``MigrationInterface`` and ``AbstractMigration`` may benefit from having an ``isDryRun()`` method, allowing third party subclasses to know if they should be dry running.

daniel-gomes-sociomantic:
> Could the option be just --dry-run without a short version?

sure, fine by me (`-r` can indeed be a bit confusing, but `-d` is already taken hehe).

> With regard to alternative templates (--template, --class), getting hold of the dry run option is a bit difficult.

I'm not familiar with those options, but they seem to be available only for the Create command, where the dry-run option doesn't come into play.

> The MigrationInterface and AbstractMigration may benefit from having an isDryRun() method, allowing third party subclasses to know if they should be dry running.

This can be easily done, I'll add it.

daniel-gomes-sociomantic:
removed the `-r` shortcut and updated the PR description accordingly

BallisticPain:
+1

What is left here?

thinkjson:
+1

rquadling:
>I'm not familiar with those options, but they seem to be available only for the Create command, where >the dry-run option doesn't come into play.

No, those options are used to create a migration template. The template may use phinx classes to execute the migration. Therefore knowing if the migration is dry running is the requirement. And having that available via the ``AbstractInterface`` allows that to operate successfully.

Which is exactly what you've done! :smile:

timhoeppner:
Anything I can help with to get this merged? Really could use this.

rquadling:
@timhoeppner Cloning @robmorgan and have the clone do our bidding/merging/releasing?

daniel-gomes-sociomantic:
> @timhoeppner Cloning @robmorgan and have the clone do our bidding/merging/releasing?

haha +1

timhoeppner:
Yes that's how I'm testing the PR.. I don't really see the humor in doing that. If there's anything I can do to help this make it into the next release I'm happy to do so.

daniel-gomes-sociomantic:
to whom it may concern, I'm right now working on adding dry-runs for seeding (ie. for inserting data rows) and adding tests for the whole thing

daniel-gomes-sociomantic:
done!

daniel-gomes-sociomantic:
rebased to `v0.5.2`

daniel-gomes-sociomantic:
rebased onto the correct `0.5.x-dev` branch

elazar:
@daniel-gomes-sociomantic Looks like this needs to be rebased again.

@robmorgan Any feedback on what it would take to get this merged?

beniamin:
Any ideea when this pull request will be merged and ready for use?

elazar:
@robmorgan Can I ask why this was closed?

robmorgan:
@elazar I removed the `0.5.x-dev` branch and it closed alot of PRs so I'm re-opening them again against `master` where we have moved development.

daniel-gomes-sociomantic:
@robmorgan the current code is probably not mergeable with `master`. I remember it also had some problems, but since I got no feedback I didn't feel like looking into them.

Is this something you want in phinx? Does it make sense I spend time looking at it?

robmorgan:
@daniel-gomes-sociomantic yes there are a few things I would change, but generally I think its a good idea. There is another PR I need to revisit which outputs the full SQL code as migrations are actually running. These two features are both similar so I'd like to see if we can combine them somehow.

robmorgan:
Ok I've now figured out in my mind how this will work and benefit most users:

1. We will improve the verbosity command to output the full SQL code as migrations are executing.
2. The dry run feature will simply be a flag that determines whether or not to actually do the execution.

Current question is around the syntax. Should adding the `--dry-run` option automatically increase the verbosity which in turn outputs the SQL? That's what I'm currently debating. It might look like:

```
$ php vendor/bin/phinx migrate -e development --dry-run
```

or 

```
$ php vendor/bin/phinx migrate -e development -vv --dry-run
```

timhoeppner:
IMHO, I think:
`$ php vendor/bin/phinx migrate -e development --dry-run`

should automatically increase the verbosity because I wouldn't bother doing a "dry run" unless I can perform my own evaluation of the SQL to be executed.

robmorgan:
@timhoeppner yes personally I agree it would be pretty pointless, but maybe some people just want to see the plan of the migrations it intends to run. e.g:

```
$ php vendor/bin/phinx migrate -e development -t 20170210111834 --dry-run

Phinx by Rob Morgan - https://phinx.org. 0.7.0

using config file ./phinx.yml
using config parser yaml
using migration paths 
 - /var/www/html/db/migrations
using seed paths 
using environment development
using adapter mysql
using database magento
warning this is a dry run

 == 20170210111701 UserTest: would migrate
 == 20170210111834 CreateTest: would migrate

All Done. Took 0.1737s
```

I'll see if theres any interest for something like that.

daniel-gomes-sociomantic:
> IMHO, I think:
> $ php vendor/bin/phinx migrate -e development --dry-run
> 
> should automatically increase the verbosity because I wouldn't bother doing a "dry run" unless I can perform my own evaluation of the SQL to be executed.

same here.

Another topic, @robmorgan I remember that the approach I followed had some problems with data migrations, ie. migrations that only operate on data and not on structure. 

For instance, if you have some code like:

```php
$rows = $this->query('SELECT * FROM ...');

foreach ($rows as $row) {
    $this->execute("UPDATE ... WHERE id = {$row['id']}");
}
```

Then it isn't really possible to dry-run this, since the `UPDATE` queries to be executed depend on the result of the `SELECT`. 

At the time this led me to think that `SELECT` queries should **not** be dry-run (ie.should be executed even if dry-running), but this might have implications I didn't consider (although the "don't change anything" premise of a dry-run would still ensured).

robmorgan:
@daniel-gomes-sociomantic ahh very good point, but unfortuntely that throws a rather large spanner into the works. We can either modify the `query()` method to return empty arrays in a dry run or accept `SELECT` queries, but this starts to get really ugly as we'd have to parse the strings being passed in.

robmorgan:
Maybe Rails doesn't do this for this particular reason.

daniel-gomes-sociomantic:
> We can either modify the query() method to return empty arrays

that wouldn't really help with the example I wrote, the user would still not be able to see how the `UPDATES` would look like if the migration is really executed. For this, the only way I see it is by allowing the SELECT queries to run which, personally, I think is the best solution. But maybe that's just me hehe. 

Would it be too much to add "allow select queries" option when dry-running (and perhaps warning the user if dry-running and a SELECT query is detected)?

> but this starts to get really ugly as we'd have to parse the strings being passed in.

Does phinx allow for non-SELECT queries on `$this->query()`calls?

robmorgan:
@daniel-gomes-sociomantic 

| Does phinx allow for non-SELECT queries on $this->query() calls?

It's doesn't block them. Thats the problem.

daniel-gomes-sociomantic:
Right, so the only way might be to indeed parse all query strings to see if they are `SELECT` queries and (assuming a dry-run):

* If they aren't, echo the query but do not execute it.
* If they are, echo the query and execute it (this could be dependent on an extra option as I mentioned earlier though).

If at some point Phinx starts enforcing that the queries executed by `$this->query()` must be `SELECT` queries, this can be improved by always allowing those queries to run and dry-running the remaining ones.

daniel-gomes-sociomantic:
(in the meantime, rebased to master so its up-to-date)

daniel-gomes-sociomantic:
> Right, so the only way might be to indeed parse all query strings to see if they are SELECT queries and (assuming a dry-run):
> 
> If they aren't, echo the query but do not execute it.
> If they are, echo the query and execute it (this could be dependent on an extra option as I mentioned earlier though).
> If at some point Phinx starts enforcing that the queries executed by $this->query() must be SELECT queries, this can be improved by always allowing those queries to run and dry-running the remaining ones.

How does this sound @robmorgan ?

JayPHP:
https://github.com/cakephp/phinx/pull/1127/files has since been merged. @daniel-gomes-sociomantic I'm not sure why this one got missed. Is there anything you would like to add to #1127 ?

lorenzo:
I'm missing pull requests that are more than one year old. I'm open to improvements to those that got merged already :)

lorenzo:
btw, I don't think having `--dry-run` from seeds is necessary. The use case is not very common, as it is for potentially destructive operations like migrations.

daniel-gomes-sociomantic:
Nop, seems like that other PR does exactly the same as this, so all good! I'm just happy there's such an option now :-)

PR #748, opened 2016-01-13
Assign $output before initialising the migration

rquadling:
Allows the option of the initialisation phase of a migration to output
something meaningful, with appropriate styling options, if pre-conditions
for the migration aren't met in some way.

### DIFF
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -68,7 +68,7 @@ abstract class AbstractMigration implements MigrationInterface
     final public function __construct($version, OutputInterface $output = null)
     {
         $this->version = $version;
-        $this->output = $output;
+        $this->setOutput($output);
 
         $this->init();
     }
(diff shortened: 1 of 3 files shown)


rquadling:
I suppose this is for my abstract migration subclasses. We use different styling and assign them during the init() phase. But there is no $output to interact with. Hadn't quite mentally joined up the fact that $output was set, but after init() was called. This PR will allow the subclass to have access to $output within the init() phase.

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -62,11 +62,14 @@
...
      *
-     * @param int $version Migration Version
+     * @param int                  $version Migration Version
+     * @param OutputInterface|null $output
      */
-    final public function __construct($version)
+    final public function __construct($version, OutputInterface $output = null)
     {
         $this->version = $version;
+        $this->output = $output;
```
shadowhand: In order to not make this a breaking change, this needs to call `$this->setOutput($output)`.

rquadling:
@shadowhand Changes made as requested.

rquadling:
Rebased to 5.2. Anything else outstanding for this to be merged?

rquadling:
What are we waiting on for this?

PR #757, opened 2016-01-20
Break into tables documentation into logical sections

coatesap:
The 'tables' section had become difficult to navigate and find information quickly, so I've broken it down into 4 discrete sections:
 - Tables
 - Columns
 - Indexes
 - Foreign Keys


shadowhand:
Makes sense to me! Thanks!

PR #762, opened 2016-01-25
General concepts only in navigation-level headings

coatesap:
New users of Phinx were previously confronted with 'The AbstractMigration Class' as a navigation-level heading, which is really an implementation detail, rather than a key area of knowledge that they will be looking for.


robmorgan:
@coatesap thx, i agree the doco needs a bit of a rework. currently one of many things on my todo list.

robmorgan:
before I merge this I need to check if there are any references floating around.

rquadling:
@robmorgan Did you find any references?

PR #765, opened 2016-01-25
Revert "Capturing migrations with GLOB_BRACE"

rquadling:
Reverts robmorgan/phinx#736

Support for multiple migration paths isn't ready yet.


rquadling:
I assume as this hasn't been merged and that unless you use braces in the migration path, there is no issue, this PR can be deleted.

@shadowhand, @robmorgan, @wandersonwhcr - comments?

Issue #250, opened 2014-05-14
Unsigned primary keys


johnblythe:
Is there a way to cause the autogen'd primary key ('id') to be unsigned?

adayth:
Alternative method overriding Phinx classes.
https://gist.github.com/f0ead7bd210e3968208e.git

paul-at-start:
This is a feature I miss. Doesn't necessarily break the app, but it hurts inside knowing that all my auto-increment id's will be signed for no practical reason.

steefaan:
@robmorgan I would love to see this feature. Sure @twoixter solution works great (thanks!) but it would be nicer to have the possibility to configure it directly.

caseycs:
Just posted a small update for @twoixter solution: https://gist.github.com/adayth/f0ead7bd210e3968208e#gistcomment-1585602

bubach:
Who came up with the brilliant idea to set primary keys as signed in the first place, that's one of the reasons I'm starting to think raw SQL files for migrations wasn't such a bad idea after all... Overall Phinx feeling -> This is just retarded.

robmorgan:
Hi @bubach that was me and it's why I no longer work on this project anymore. You might want to check out this instead: https://flywaydb.org/ 👍

PR #768, opened 2016-01-27
Allow to pass table option signed for primary key

DimazzzZ:
Just new pull request for #658 made by @Serhioromano. Very useful feature.

Are additional tests required?

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -215,6 +215,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $column = new Column();
             $column->setName('id')
                    ->setType('integer')
+                   ->setSigned(isset($options['signed']) ? $options['signed'] : false)
                    ->setIdentity(true);
 
             array_unshift($columns, $column);
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -215,7 +215,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $column = new Column();
             $column->setName('id')
                    ->setType('integer')
-                   ->setSigned(isset($options['signed']) ? $options['signed'] : false)
+                   ->setSigned(isset($options['signed']) ? $options['signed'] : true)
                    ->setIdentity(true);
 
             array_unshift($columns, $column);
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -455,6 +455,37 @@ To simply change the name of the primary key, we need to override the default ``
             }
         }
 
+By default ``primary_key`` is ``signed``.
+To simply set it to unsigned just pass ``signed`` option with ``false`` value:
+
+.. code-block:: php
+
+        <?php
+
+        use Phinx\Migration\AbstractMigration;
+
+        class MyNewMigration extends AbstractMigration
+        {
+            /**
+             * Migrate Up.
+             */
+            public function up()
+            {
+                $table = $this->table('followers', array('signed' => false));
+                $table->addColumn('follower_id', 'integer')
+                      ->addColumn('created', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
+                      ->save();
+            }
+
+            /**
+             * Migrate Down.
+             */
+            public function down()
+            {
+
+            }
+        }
+
 Valid Column Types
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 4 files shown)


DimazzzZ:
Please wait I'll add some fixes.

DimazzzZ:
Please check my fixes in tests, are they correct? The main idea is that table column that is used in FK must be unsigned as referenced table's pk is.

rquadling:
Will these changes still allow me to create a signed primary key? I think the tests are now only looking for unsigned primary keys.

DimazzzZ:
@rquadling yes they allow. For example:

```php
// Default (unsigned)
$this->table('test')->addColumn('name', 'string')->create();

// Custom with signed PK
$this->table('test', ['signed' => true])->addColumn('name', 'string')->create();
```

I fixed existing tests because of changing default behavior, so I think we need additional tests with signed pk.

rquadling:
@DimazzzZ The extra tests to make sure signed PK is still feasible. I'd be wary about changing the default though. Consider all those migrations out there already. Rolling them back and then forward again could lead to issues. Ones that come under the heading of "It used to work and we've not changed anything!".

DimazzzZ:
I understand your concerns about these changes but how many and how often developers make signed PK? Or even other ids? IMHO most of them using hacks like @twoixter's from #250 (thanks him btw). 

My position is "table auto increment pk must be unsigned" since `AUTO_INCREMENT` starts at zero by default and increments in the positive direction.

I'll add more tests as soon as posible.

twoixter:
Hi @DimazzzZ , couldn't agree more! +1, or better +1,000 to this PR. ;-)

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -215,6 +215,7 @@ public function createTable(Table $table)
             $column = new Column();
             $column->setName('id')
                    ->setType('integer')
+                   ->setSigned(isset($options['signed']) ? $options['signed'] : false)
```
shadowhand: This represents a backwards compatibility break as implemented. The way to do it without breaking BC would be to keep the current behavior (signed by default) and respect the `signed` option to create unsigned PKs.

DimazzzZ: > The way to do it without breaking BC would be to keep the current behavior (signed by default) and respect the signed option to create unsigned PKs.

I thought about it and that was a second option... On the one hand it breaks a backwards compatibility but on the other hand is `signed` pk is right behavior or we just want to save backwards compatibility forever? We can inform users via changelog about new pk behavior as it was done with PHP 5.3 support, by the way nobody forbids to download an older version of Phinx.

Also we may add new property to phinx.yml like `signed_pk: false` or something.

Let's discuss.

rquadling:
Minor patch upgrades must not break BC. This is important as people rely on the semantic version numbers.

http://semver.org/

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

So, at best, introduce this as a patch, but allow the user to configure their setup to use unsigned PK, but the default must be signed.

DimazzzZ:
@rquadling okay this is a good reason. I'll set it as `signed` for now. 

May we set this behavior as default in the major 1.0 version?

DimazzzZ:
Added fix as a minor patch with `signed` default behavior and new tests. Check this out please.

Who can restart Scrutinizer task or just merge the PK?

rquadling:
We have to wait for merges. Takes a while sometimes.

shadowhand:
Looks good to me. @robmorgan do you have any objections to this?

shadowhand:
Can you add this option to the docs?

DimazzzZ:
I was in hospital, sorry. The docs has been updated.

rquadling:
Hope all is OK.

DimazzzZ:
@rquadling all is ok, thanks!

Any updates on this PR?

rquadling:
Still waiting for Rob.

DimazzzZ:
Version 0.6.x released but PR still opened...  Should I re-create PR for 0.6.x branch? (

rquadling:
Hi. Please do.

magnus-eriksson:
Is there any ETA on this patch?

Serhioromano:
Or my goodness! For one line change more than 100 lines of testing. That is what I call testing obsession. Technologies are good, but sometimes they just overkill.

@DimazzzZ thank you for this PR. Hope it gets accepted.

robmorgan:
I can accept this if the default is `unsigned` however I think the docs need to be updated to say its MySQL specific.

DimazzzZ:
@robmorgan so I update for `unsigned` default behavior? As @rquadling says it may cause backward compatibility problems for < v.0.6.*.

dmarklund:
Any updates to this PR?

jakejohns:
any word here?

magnus-eriksson:
Version 0.8 is now out and this is still open? Any ETA?

robmorgan:
I tweaked the docs a little bit. I was on the fence about renaming the `signed` table option to `unsigned_primary_key`, but I've left it alone for now. I'll ship it later today.

PR #770, opened 2016-01-28
Continue to produce output at higher verbosities

tristanpemble:
When using the `-vv` and `-vvv` flags, there was no output from the command.


robmorgan:
thx I was able to replicate it!

PR #772, opened 2016-02-01
Fix example usages of save()

sbol-coolblue:
To revert a migration on the go
You will need to let Phinx know
The way you want your table to behave
So please, avoid using the method "save()"


robmorgan:
@sbol-coolblue nice poem, however inside `up()` and `down()` methods you can use `save()` if you desire.

sbol-coolblue:
Thanks for the feedback :) Makes sense, closing the PR.

PR #778, opened 2016-02-10
add bit type support for MysqlAdapter

kajisha:
Add `bit` type support for MysqlAdapter.


mvrhov:
IMO it would be fine to also extend to [PostgreSQL bit type](http://www.postgresql.org/docs/9.4/static/datatype-bit.html)

PR #788, opened 2016-02-24
Added change events

### LINKED ISSUE #626: ability to run up/down alongside with change
while working on a project that brought up #625 another thing is that i have to enable/disable an postgres extension while running the migration.

i'm not used to `change()` (didn't exist in the previous tools i've used) i've started figuring out what my migration actually needs to do and one thing it needs to do is enable an extension. which is absolutely easy using `execute` to issue a query.

reversing the migration fails because the extension is already enabled (obviously). extending the `CREATE EXTENSION` query with a `IF NOT EXISTS` would work, although i dislike that is does not actually disable the extension but only hide the underlying error it would generate.

Looking at `Migration\Manager\Environment#executeMigration` i see the following (compressed for brevity):

```
public function executeMigration(MigrationInterface $migration, $direction = MigrationInterface::UP)
{
    $direction = ($direction == MigrationInterface::UP) ? MigrationInterface::UP : MigrationInterface::DOWN;

    if (method_exists($migration, MigrationInterface::CHANGE)) {
        if ($direction == MigrationInterface::DOWN) {
            $migration->change();
            $proxyAdapter->executeInvertedCommands();
        } else {
            $migration->change();
        }
    } else {
        $migration->{$direction}();
    }
}
```

which goes either for `up`/`down` **or** `change` - which is in agreement with the documentation:

> Please be aware that when a `change` method exists Phinx will automatically ignore the `up` and `down` methods. If you need to use these methods it is recommended to create a separate migration file.

only that having another migration is not really fixing my problem.

to conclude the question/idea: enabling `up`/`down` to run alongside `change` might not be the best idea (given that it did not work that way until now and would be a BC) - i was thinking about a hook that would be executed (in case it exists) _before_ the migration is going up and _after_ the migration is going down.

any thoughts on that? or is there already something else that people are using and that i'm simply overlooking ;o

mvrhov:
Added onBefore(Up|Down)/After(Up|Down) functions that are called in the process of change migration.

This fixes #626

### DIFF
--- a/src/Phinx/Migration/Migration.template.php.dist
+++ b/src/Phinx/Migration/Migration.template.php.dist
@@ -24,8 +24,10 @@ class $className extends $baseClassName
      *
      * Remember to call "create()" or "update()" and NOT "save()" when working
      * with the Table class.
+     *
+     * @param string $direction Is one of MigrationInterface::UP or MigrationInterface::DOWN
      */
-    public function change()
+    public function change($direction)
     {
 
     }
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -56,7 +56,7 @@ Phinx automatically creates a skeleton migration file with a single method:
              * Remember to call "create()" or "update()" and NOT "save()" when working
              * with the Table class.
              */
-            public function change()
+            public function change($direction)
             {
 
             }
@@ -94,7 +94,7 @@ down automatically for you. For example:
              *
              * Uncomment this method if you would like to use it.
              */
-            public function change()
+            public function change($direction)
             {
                 // create the table
                 $table = $this->table('user_logins');
@@ -128,7 +128,7 @@ recommended to create a separate migration file.
 
 .. note::
 
-    When creating or updating tables inside a ``change()`` method you must use
+    When creating or updating tables inside a ``change($direction)`` method you must use
     the Table ``create()`` and ``update()`` methods. Phinx cannot automatically
     determine whether a ``save()`` call is creating a new table or modifying an
     existing one.
@@ -298,7 +298,7 @@ insert methods in your migrations.
 
 .. note::
 
-    You cannot use the insert methods inside a `change()` method. Please use the
+    You cannot use the insert methods inside a `change($direction)` method. Please use the
     `up()` and `down()` methods.
 
 Working With Tables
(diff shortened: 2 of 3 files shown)


mvrhov:
I have nothing against that. And that would be a sensible approach, however this means that everyone would need to update their migrations.

rquadling:
Nope. As `change()` is not a defined method. It is called if it exists. If it doesn't exist, it can't be called and also cannot be defined in an interface anywhere.

mvrhov:
I probably missed something :P

src/Phinx/Migration/Migration.template.php.dist
```diff
@@ -24,8 +24,10 @@ class $className extends $baseClassName
      *
      * Remember to call "create()" or "update()" and NOT "save()" when working
      * with the Table class.
+     *
+     * @param string $direction
      */
```
rquadling: I'd add some description here along the lines that the value will be one of the constants I previously highlighted.

mvrhov:
Also the in the `Inserting Data`  section there is a note that is now redundant?

src/Phinx/Migration/Migration.template.php.dist
```diff
@@ -24,8 +24,11 @@ class $className extends $baseClassName
      *
      * Remember to call "create()" or "update()" and NOT "save()" when working
      * with the Table class.
+     *
+     * $direction value is one of MigrationInterface::UP or MigrationInterface::DOWN
+     * @param string $direction
```
rquadling: `* @param string $direction Is one of ...` - inline the description.

rquadling:
Inserting Data in a `change()` method would still be considered a no-no. OOI, why would you not stick with `up()` and `down()` if you wanted to control anything regarding direction? `change()` really is only about reversible processing.

rquadling:
Whilst the changes are valid, I'm still not sure why `change($direction)` over `up()` and `down()`.

And due to the fact that, internally, the table manipulation methods used in a migration's `change()` method are 'reversed' automatically based upon direction, I think there is a big scope for significant fubar here.

As things stand, the commits are sound, but the actual usage, I'm not so sure about. Convince me (and anyone else who is reading this and wondering why).

mvrhov:
Because change knows how to reverse changes itself. However there are some things that one would like to do and cannot be reversed automatically. e.g create some stored procedures, triggers or some other things.

Why do I need to write up/down for everything and not just for things phinx doesn't know how to reverse?

rquadling:
For views, stored procedures, triggers, etc. we will be building repeatables - migrations that can be run again and again specifically for these sort of changes.

But I can see a level of use for this feature.

mvrhov:
PostgreSQL doesn't have if not exists or similar for trigger.

PR #794, opened 2016-02-29
Allow for environment-dependent migrations

camcima:
This pull request makes the environment object available when writing a migration, so you can account for different requirements you might have for each specific environment.


daniel-gomes-sociomantic:
we'd also like to know the environment name from inside a migration. 

For instance, when deploying a tighter business rule with a unique index, we'd like to remove any invalid/duplicated data in the staging environment (since it's only for tests and so isn't critical), but let the index adding fail in the production environment if there is invalid data so that we are forced to solve the duplicated rows manually, thus ensuring it's all done correctly.

daniel-gomes-sociomantic:
another use case is a migration that drops a column: its `down` method could simply re-add the column for test environments (without worrying about the loss of data), whereas on Production this would not be allowed

daniel-gomes-sociomantic:
@camcima you should probably rebase your PR onto the new default branch

rquadling:
Absolutely good idea. Can you rebase on 0.7.x-dev please!

rquadling:
I would recommend simply injecting the environment into the AbstractMigration constructor.

As 0.7.x-dev isn't released yet, this would be an ideal time to do this.

camcima:
The `Environment` is now being injected into the `AbstractMigration`.

daniel-gomes-sociomantic:
nice and easy :+1:

rquadling:
With this approach, you won't have access to the environment until execution (i.e. up/down). As init() is the first point of contact, the environment needs to be injected as part of the constructor.

Just like the input and output.

https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Migration/Manager.php#L603.

So, the AbstractMigration::__constructor() receives the environment and stores it (using a setter()).

And then, in the migrations init() method, you'll have access to the environment before up/down/change.

robmorgan:
Looking at this PR now, I'll all for making the `Environment` available inside your migrations. I know some people have very specific use-cases. However in general I strongly advise against environment specific migrations. I push my teams to achieve dev/staging/prod parity where ever possible. That way we are guaranteed that whatever happens on staging will most likely happen on production. Its one of the key tenants of the 12 factor (https://12factor.net) app.

rquadling:
Hopefully the use case is more to do with needing something from the environment rather than branching due to the something from the environment.

I'll tag this for me to look at at the weekend.

Thank you for your patience.

dereuromark:
What is the status of this enhancement? Do we want to merge this?

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -82,7 +82,7 @@
     /**
      * Class Constructor.
      *
-     * @param int $version Migration Version
+     * @param float $version Migration Version
```
dereuromark: not int?

camcima: It is declared as float in the phpdoc for the class property. Maybe because of size limitation for int in windows 32-bit platforms.

WallTearer:
Any progress on this feature?
Having a possibility to determine environment inside a migration command would be really helpful to insert/update environment-dependant data.

lorenzo:
@WallTearer you can already do that. Use `$this-> getEnvironment()` to know the name of the environment inside a migrations.

I'm closing this as it is quite old, there are conflicts and seems to no be progressing.

WallTearer:
@lorenzo thanks for a quick reply. I see now it was added in the recent version 0.10.0.

PR #799, opened 2016-03-02
Added support for the "seed:run" command in TextWrapper

rikbruil:
When I was trying to get our database migrations and seeds executed from within our own web app, I noticed support for executing seeds was lacking in the default TextWrapper.

It wasn't very hard to extend the TextWrapper and write our own implementation for it, but I think it would be nice if Phinx supported this out of the box (since I can imagine more people wanting to fill their local dev environment with data after migrating). 

Tests for the original TextWrapper class were missing, so I did not write tests for this new method. It follows the conventions of this class as far as I can tell. I was doubting if I should also extend the app/web.php example file.

Feedback is welcome :)

### DIFF
--- a/src/Phinx/Wrapper/TextWrapper.php
+++ b/src/Phinx/Wrapper/TextWrapper.php
@@ -147,6 +147,25 @@ class TextWrapper
     }
 
     /**
+     * Returns the output from running the "seed:run" command.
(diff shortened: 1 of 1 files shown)


robmorgan:
closing in favor of: https://github.com/robmorgan/phinx/pull/921.  `web.php` should be extended separately.

PR #808, opened 2016-03-12
add column without process conment in postgres adapter

fastcgi:
If someone use postgres, and add column to a table, phinx forgeted to process the comment in `$options` array.

``` php
        $table = $this->table('trades');
        $table->addColumn(
            'total_amount', 
            PostgresAdapter::PHINX_TYPE_DECIMAL, 
            ['precision' => 10, 'scale' => 2, 'comment' => 'the trade\'s total amount' ]
        )->update();
```


JayPHP:
Hi @fastcgi, could you fix the conflicts and create an issue relating to this PR please?
This will also need some additional tests.

PR #810, opened 2016-03-16
Adding support for mysql adapter to use default collation and engine

danielasplund:
This commit makes the mysql adapter to use default collation and engine settings from master configfile, Instead of the fallback of `InnoDB` and `utf8_general_ci` wich were default if no table options were given or present.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -207,6 +207,15 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             'engine' => 'InnoDB',
             'collation' => 'utf8_general_ci'
         );
+
+        $options = $this->getOptions();
+        if (isset($options['engine'])) {
+            $defaultOptions['engine'] = $options['engine'];
+        }
+        if (isset($options['collation'])) {
+            $defaultOptions['collation'] = $options['collation'];
+        }
+
         $options = array_merge($defaultOptions, $table->getOptions());
 
         // Add the default primary key


priidikvaikla:
What else needs to be done so that this pull request would get merged?

HillLiu:
👍

lvht:
Could you @shadowhand review this PR?

robmorgan:
by the looks of things unit tests - I haven't checked but i doubt there is a test wrapping that code.

glensc:
i looked the code and figured phinx doesn't suppor this. so i added overriden `table` method to my migration. and as `__construct` did not have adapter initialized, setup the class variables in `change` method.


```php
...

class InitDatabase extends AbstractMigration
{
    /**
     * MySQL Engine
     * @var $engine
     */
    private $engine;

    /**
     * MySQL Charset
     * @var $string
     */
    private $charset;

    /**
     * MySQL Collation
     * @var $string
     */
    private $collation;

    private function initOptions()
    {
        // extract options from phinx.php config
        $options = $this->getAdapter()->getOptions();
        $this->charset = $options['charset'];
        $this->collation = $options['collation'];
        $this->engine = $options['engine'];
    }

    public function change()
    {
        $this->initOptions();

        $this->table('api_token', ['id' => 'apt_id'])
            ->addColumn('apt_usr_id', 'integer', ['length' => 10])
...
        ->create();
...
    }

    /**
     * {@inheritdoc}
     */
    public function table($tableName, $options = [])
    {
        $options['engine'] = $this->engine;
        $options['charset'] = $this->charset;
        $options['collation'] = $this->collation;

        return parent::table($tableName, $options);
    }
```

glensc:
@danielasplund can you add `charset` as well, as my above example does?

PR #812, opened 2016-03-28
Fix date format so it works when mssql is not configured in us-english.

hermesti:
We cannot run migrations due to problems with date format in the phinxlog table. We use mssql with spanish date format: "dmy". Phinx always use the format 'Y-m-d H:i:s' which is not recognized by mssql with this configuration. Our change sets a date format in the SqlServerAdapter that is always recognized by mssql, regardless language configuration.

Environment:
PHP 5.4.45
SQL Server 2012

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -1169,6 +1169,10 @@ SQL;
      * {@inheritdoc}
      */
     public function migrated(MigrationInterface $migration, $direction, $startTime, $endTime) {
+	    // Fix date format so it works when mssql is not configured in us-english.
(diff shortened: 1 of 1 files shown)


src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1169,6 +1169,10 @@ public function getColumnTypes()
      * {@inheritdoc}
      */
     public function migrated(MigrationInterface $migration, $direction, $startTime, $endTime) {
+	    // Fix date format so it works when mssql is not configured in us-english.
```
robmorgan: @hermesti could you fix the comment indentation here?

JayPHP:
Closing this due to age, conflicts and no response from author. Please open a new Issue if you would like this re-opened.

PR #817, opened 2016-04-08
Added Foreign Key table name escaping (Postgres)

flavioheleno:
It avoids problems when the foreign key's referenced table has a special name, such as user.


flavioheleno:
I've messed up the PR, I'll fix and add it again, sorry!

PR #821, opened 2016-04-13
Provide the ability to run multiple seeds at once (in the given order)

enygma:
This PR adds in a simple update to allow for running multiple seeds at once.
Currently the seed command allows you to specify which you'd like to run with the `-s` command line flag like:
```
vendor/bin/phinx seed:run -s UserCreate
```

With this PR, multiple seeds can be defined here using a comma-separated list:
```
vendor/bin/phinx seed:run -s UserCreate,ProductCreate
```

The seeds are then run in that order (as opposed to the normal `seed:run` that just does them in the order `glob` finds them in).

### DIFF
--- a/src/Phinx/Console/Command/SeedRun.php
+++ b/src/Phinx/Console/Command/SeedRun.php
@@ -71,6 +71,7 @@ EOT
 
         $seed        = $input->getOption('seed');
         $environment = $input->getOption('environment');
+        $seedSet     = (strstr($seed, ',') !== false) ? explode(',', $seed) : [$seed];
 
         if (null === $environment) {
             $environment = $this->getConfig()->getDefaultEnvironment();
@@ -104,7 +105,9 @@ EOT
 
         // run the seed(ers)
         $start = microtime(true);
-        $this->getManager()->seed($environment, $seed);
+        foreach($seedSet as $seed) {
+            $this->getManager()->seed($environment, trim($seed));
+        }
         $end = microtime(true);
 
         $output->writeln('');


enygma:
FYI: It doesn't look like the failures in the Travis-CI build above have anything to do with the functionality added by this PR.

src/Phinx/Console/Command/SeedRun.php
```diff
@@ -104,7 +105,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         // run the seed(ers)
         $start = microtime(true);
-        $this->getManager()->seed($environment, $seed);
+        foreach($seedSet as $seed) {
+            $this->getManager()->seed($environment, trim($seed));
```
bartosz-maciaszek: That `trim()` causes a problem when a developer wants to run all seeds. Previously it was possible by not passing any argument to the command (`$seed` was `null` and it was passed like this to `seed()` method).

At the moment, `trim(null)` causes that an empty string is passed to that method causing:

```
[InvalidArgumentException]
The seed class "" does not exist
```

enygma: Are you using the `-s` option to try to run all seeds or something odd? I don't see an error when I just run `phinx seed:run` on the latest `dev-master`.

bartosz-maciaszek: Yeah, simple `seed:run`. The bug is in `0.5.x-dev` branch, not in `master`. Follow your changes to see how it works when `$seed` is `null`. See https://github.com/robmorgan/phinx/issues/838 and [that patch](https://github.com/robmorgan/phinx/pull/835/files) for reference.

rquadling:
Wouldn't it have been better to use a multiple option?

`` ->addOption('--seed', '-s', InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'What is the name of the seeder?')``

That way, the ``$input->getOption('seed');`` is always an array, even for a single seed.

rquadling:
As I accidentally released this ( new to releasing others code - so sorry about that!!! ), I'll get the bug fix in ASAP. Once the test for #835 is amended to test the current situation appropriately.

PR #826, opened 2016-04-20
FIX: PdoAdapter: getVersionLog() does not quote table_name

qyanu:
fixes issue https://github.com/robmorgan/phinx/issues/825


rquadling:
Same goes for SqlServerAdapter::migrated(), PostgresAdapter::migrated(), PdoAdapter::createSchemaTable(), PdoAdapter::migrated().

I've also got some more on one of my own PRs (rollback support), so when you've done these lot, I'll merge and update mine.

Can you generate a config/unit tests for a failing case that now works with this patch.

robmorgan:
yes please add a unit test with this PR

PR #829, opened 2016-04-25
Fix loading the migration/seed files in some linux systems

onnimonni:
I found out that `GLOB_BRACE` is not available in solaris or alpine linux.

I get this error in my alpine php docker container:

```
Use of undefined constant GLOB_BRACE - assumed 'GLOB_BRACE'
```

I also googled that this can be problem in solaris too:
https://lodge.statamic.com/questions/580-getting-notice-use-of-undefined-constant

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -516,7 +516,7 @@ class Manager
     {
         if (null === $this->migrations) {
             $config = $this->getConfig();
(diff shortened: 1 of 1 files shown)


robmorgan:
@onnimonni thanks, 👍  closing in favour of: https://github.com/robmorgan/phinx/commit/28d9afb01175704062277c5365f610a08fed21d4

PR #834, opened 2016-05-03
Select from multiple migration paths for creation

### LINKED ISSUE #833: Cannot create migrations with paths using glob patterns
Given a `migrations` path with a glob pattern new migrations will be created in a directory literally named after the pattern. Thus:

```
paths:
  migrations: foo/*/migrations
```

Executing `phinx create FooBar` will create a new migration in the path `foo/*/migrations/20160503134337_foo_bar.php` literally.

Not sure how to tackle this properly but maybe ask the user where to put the migration if a glob pattern is used?

mbrodala:
This asks the user where to create migrations if multiple
valid migration paths where found.

Fixes #833

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -107,7 +107,6 @@ class Create extends AbstractCommand
         $paths = glob($this->getConfig()->getMigrationPath(), defined('GLOB_BRACE') ? GLOB_BRACE : 0);
 
         if (count($paths) > 1) {
-
           $helper = $this->getHelper('question');
           $question = $this->getMigrationDirectorySelectionQuestion($paths);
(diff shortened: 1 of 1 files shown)


mbrodala:
@robmorgan Any specific reason why this wasn't accepted?

robmorgan:
@mbrodala I closed the `0.5.x-dev` branch and it closed a lot of open PRs. I'm reopening them for consideration against the `0.7` branch which is now `master`.

mbrodala:
OK, I'll rebase.

mbrodala:
Hm, seems like this was already added with 28d9afb, thus there's nothing to do here anymore. :-)

robmorgan:
@mbrodala ahh funny that ;)

PR #835, opened 2016-05-04
"[InvalidArgumentException] The seed class "" does not exist" when trying to run all the available seeds

bartosz-maciaszek:
[This pull request](https://github.com/robmorgan/phinx/pull/821) introduces a bug that makes it impossible to run all the existing seed(ers). This one fixes it.

Surprisingly enough, the previous fix, according to the [Version History](https://github.com/robmorgan/phinx#version-history) has been included in version `0.5.4` which was never released (it's not available in [packagist](https://packagist.org/packages/robmorgan/phinx)).

### DIFF
--- a/src/Phinx/Console/Command/SeedRun.php
+++ b/src/Phinx/Console/Command/SeedRun.php
@@ -71,7 +71,6 @@ EOT
 
         $seed        = $input->getOption('seed');
         $environment = $input->getOption('environment');
-        $seedSet     = (strstr($seed, ',') !== false) ? explode(',', $seed) : [$seed];
 
         if (null === $environment) {
             $environment = $this->getConfig()->getDefaultEnvironment();
@@ -103,11 +102,18 @@ EOT
             $output->writeln('<info>using table suffix</info> ' . $envOptions['table_suffix']);
         }
 
-        // run the seed(ers)
         $start = microtime(true);
-        foreach($seedSet as $seed) {
-            $this->getManager()->seed($environment, trim($seed));
+
+        if (null === $seed) {
+            // run all the seed(ers)
+            $this->getManager()->seed($environment);
+        } else {
+            // run seed(ers) specified in a comma-separated list of classes
+            foreach (explode(',', $seed) as $seed) {
+                $this->getManager()->seed($environment, trim($seed));
+            }
         }
+
         $end = microtime(true);
 
         $output->writeln('');
--- a/tests/Phinx/Console/Command/SeedRunTest.php
+++ b/tests/Phinx/Console/Command/SeedRunTest.php
@@ -102,4 +102,38 @@ class SeedRunTest extends \PHPUnit_Framework_TestCase
         $commandTester->execute(array('command' => $command->getName()), array('decorated' => false));
         $this->assertRegExp('/using database development/', $commandTester->getDisplay());
     }
+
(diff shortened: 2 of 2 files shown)


ket4yii:
I have the same bug. I've opened #838 for it.

bartosz-maciaszek:
Please merge this pull request, since the mentioned bug is now merged and tagged as `v0.5.4`.

bartosz-maciaszek:
There you go. The unit tests have been added.

rquadling:
The test is inappropriate for the bug. The test should be more along the lines that the output should NOT contain the error being reported for specific input, not that the output contains something unrelated to the bug.

Another way at looking at this, if the bug wasn't fixed, the test would still pass, giving us a false positive.

A test named ``executingMultipleSeedersDoesNotGeneratesMissingSeedException()`` is much more descriptive of the intent of the test.

If you remove the fix in ``src/Phinx/Console/Command/SeedRun.php``, and rerun the new test, you'll see what I mean. The tests will pass, even though the bug hasn't been fixed.

bartosz-maciaszek:
Yeah, you're right. Have amended the mock in the `testExecute` to check the `seed()` method arguments and make sure an empty string, which caused the bug in the first place, is not passed.

When it comes to checking the output, I think it doesn't really matter as we have mocks that check how the `seed()` method is being called.

Have a look if you're happy with it.

rquadling:
That's better. Now have a failing test prior to fix!

What do you think of moving from a CSL of seeds to using Symfony's approach of 

```php
->addOption('--seed', '-s', InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'What is the name of the seeder?')
```
produces 
```
 -s, --seed[=SEED]                  What is the name of the seeder? (multiple values allowed)
```

```php
$seedSet = $input->getOption('seed');
```
will be an array automatically, and without any need to explode on `,`. 

Remove the explode() and we've got the same functionality without any need for more code.

I prefer this approach as ``--help`` shows the multiple values allowed.

bartosz-maciaszek:
Makes sense to me. However it'd change the syntax for input options from `-s One,Two,Three` (alternatively `--seed One,Two,Three`) to `-s One -s Two -s Three` (or `--seed One --seed Two --seed Three`). Doesn't bother me to be honest ;) Not sure if Symfony Console allows any other syntax.

BTW, are you sure it should be `InputOption::VALUE_OPTIONAL` and not `InputOption::VALUE_REQUIRED`?

rquadling:
The syntax is more standard for all Symfony Console based apps, so makes sense to follow it.

BallisticPain:
Is this in the new tagged 0.5.5? I just ran into this today.

bartosz-maciaszek:
I've pushed the changes. Since this PR is already merged, should I submit another one?

rquadling:
All should be in 0.5.5. Doc/tests/code all want -s one -s two -s three rather than -s one,two,three

rquadling:
I'd consider this issue closed.

OK with that?

bartosz-maciaszek:
No, `-s one -s two` thingy has not been merged yet. You merged and tagged before I pushed the latest changes ;)

bartosz-maciaszek:
Oh, looks like you've done it yourself in the meantime 👍

PR #843, opened 2016-05-19
Allowing configuration to be loaded form environment variables

aalexgabi:
User story: We have one database per customer but one single software version deployed. We need to automate the migration of these bases. For this I made a multi-threaded script that executes a shell command for each database. The shell command gets database credentials from environment variables set up by the script. Generating the phinx configuration for each database in shell is cumbersome and not user friendly. So I added support for loading configuration from environment variables in phinx. Another approach would have been to add command line options but given the way phinx is designed (based on the assumption that we have a configuration file and not a configuration source), I would have needed to change a lot of things.

This is how it works now:
```
db-looper PHINX_DB_NAME=$DB_NAME PHINX_DB_USER=$DB_USER PHINX_PATHS_MIGRATIONS=. phinx -p env migrate
```

Feel free to change it if you think it's not coherent with design/patterns/naming conventions etc.

I would like to be able to add unit tests tests for this but unfortunately I have very limited time.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -79,6 +79,43 @@ class Config implements ConfigInterface
     }
 
     /**
+     * Create a new instance of the config class using environment variables.
+     *
+     * @throws \RuntimeException
+     * @return Config
+     */
+    public static function fromEnv()
+    {
+        /** Database name is required */
+        if(getenv('PHINX_DB_NAME') === false) {
+            throw new \RuntimeException('Cannot get database name from PHINX_DB_NAME environment variable');
+        }
+
+        /** @noinspection PhpIncludeInspection */
+        $configArray = array(
+            'paths' => array(
+                'migrations' => getenv('PHINX_PATHS_MIGRATIONS') ? : getcwd(),
+                'seeds' => getenv('PHINX_PATHS_SEEDS') ? : getcwd()
+            ),
+            'environments' => array(
+                'default_migration_table' => 'phinxlog',
+                'default_database' => 'environment',
+                'environment' => array(
+                    'adapter' => getenv('PHINX_DB_ADAPTER') ? : 'mysql',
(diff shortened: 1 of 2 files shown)


rquadling:
The solution we use is to use phinx.php rather than phinx.yml. And then our config looks like ...

```
                'production' =>
                    [
                        'adapter' => 'mysql',
                        'host' => $_ENV['DB_PRODUCTION_HOST'],
                        'name' => $_ENV['DB_PRODUCTION_DATABASE'],
                        'user' => $_ENV['DB_PRODUCTION_USERNAME'],
                        'pass' => $_ENV['DB_PRODUCTION_PASSWORD'],
                        'port' => 3306,
                        'charset' => 'utf8',
                        'collation' => 'utf8_unicode_ci',
                    ],
```

aalexgabi:
Good idea! I'll try it. This PR renders the `.php` file optional.

So what's your take on merging this?

rquadling:
We just swapped from using .yml to .php. Phinx supports both without changes.

rquadling:
I can see it as useful for those using env-vars without a config. But if you are using the env parser, you cannot use the same setup on development as you would need additional config; for example migration_base_class, templates, aliases.

rquadling:
I'm not in favour of merging this. Using a config file is what we use (YAML, PHP and JSON formats supported). If you want to pull from environment variables, then the PHP based config supports that already.

There is a PR being formulated that will allow you to define the config file location via an environment variable, allowing more control.

If there is nothing else, then I will close this PR tomorrow.

aalexgabi:
Sold. You are right this is not complete and needs to be maintained manually. I used the `.php` file as you suggested and it works like a charm :+1: 

It didn't occur to me at that time that I can do that :sob: 

> There is a PR being formulated that will allow you to define the config file location via an environment variable, allowing more control.

Not useful in my case because there are currently 3099 databases and they are created/deleted for each client all the time so I would need to update the config files very frequently.

PR #846, opened 2016-05-27
add use of limit to datetime, timestamp and time column types

rchrdkn:
add use of limit to datetime, timestamp and time column types for specifying fractional precision.

http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html


mvrhov:
Why the limit field.. If it's named a precision in a documentation.
Please also add support for postgresql.. the precision is specified the same as in mysql `field(p)`

PR #850, opened 2016-05-31
Introduce support for query parameter binding.

courtney-miles:
This is a refreshed attempt to introduce support for binding query parameters.  Original PR was #486.

I'm open to suggestion about this interface.  I'm a little unsure about the array of QueryBind objects:

``` php
$this->query(
    'SELECT * FROM users WHERE user_id BETWEEN :minId AND :maxId',
    array(
        ':minId' => new QueryBind(10, QueryBind::TYPE_INT),
        ':maxId' => new QueryBind(15, QueryBind::TYPE_INT)
    )
);
```

I don't quite like that the `QueryBind` does not include the parameter name, which is instead determined by the array key.  But then, this works nicer with positional parameters, and the alternatives I can think of would make it a more complicated system for users.


courtney-miles:
The current build failure appears to be a pre-existing issue with the 0.5.x-dev branch.

cyrusboadway:
You have the patience of a saint.

arjoe:
I too think this would be a much welcomed feature. I also concur that the fact that that the QueryBind instance does not contain the parameter name seems odd. I would prefer to pass an array of QueryBind objects that contain (1) the parameter name, (2) parameter type, and (3) parameter value.

Koriit:
Why not just add `prepare` method to AdapterInterface to return PDOStatement object and allow users to call `bindParam`/`bindValue` themselves or to just execute the statement with parameters array?

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -301,17 +301,60 @@ public function disconnect()
...
+            $statement = $this->getConnection()->query($sql);
+        } else {
+            $statement = $this->executeParameterisedQuery($sql, $bindValues);
+        }
+
+        return $statement;
+    }
+
+    /**
+     * @param $sql
```
glensc: type string

tchalvak:
Binding is crucial.  But instead of writing a completely new API/approach, why not just wrap PDO's existing system, but exposing it within phinx!  Don't rewrite this wheel!

tchalvak:
Binding is crucial.  And reading through the discussion in: https://github.com/cakephp/phinx/pull/486 this looks like the most likely way that this will ever be available.

lorenzo:
@tchalvak we are considering adding a more robust and feature rich database abstraction layer to phinx in the next versions. So this feature is coming soon

lorenzo:
Closing this as I'm working on a more feature-rich solution to this and the code has too many conflicts now

lorenzo:
Thanks for your contribution anyways, it helped us moving forward

glensc:
please link to the new solution so interested parties can subscribe. thanks!

michaelze:
@lorenzo
Any update on this? Thr docs at https://book.cakephp.org/3.0/en/phinx.html don't seem to mention prepared statements...

PR #859, opened 2016-06-13
Feature/add support for input interface

rquadling:
Is based upon a previous outstanding #748 to inject the $output (and now the $input) into the migration/seed.

If this PR is accepted, then #748 will no longer be required.


rquadling:
Not really merged. Being an idiot I pushed to the wrong repo.

PR #866, opened 2016-06-16
Add support for SQLite enums using CHECK().

basilfx:
SQLite has dynamic types, and no support for ENUMs. This PR adds support for ENUMs using CHECK(). This allows us to at least ensure the data that is inserted is OK.

To give an idea of what it generates:

```sql
CREATE TABLE "node_rev" (
	`nodeID`	INTEGER  PRIMARY KEY AUTOINCREMENT,
	`nodeType`	ENUM CHECK(nodeType IN ( 'block' , 'relation' , 'parameter' , 'validation' , 'transformation' , 'model' , 'experiment' , 'image' , 'text' ))
)
```


robmorgan:
thx

PR #869, opened 2016-06-17
Config File Location Feature

ralphschindler:
Features added:
- ability to provide PHINX_CONFIG_DIR to scan
- ability to traverse one level deep looking for matching config

Thoughts on including this?

The motivation is to allow the config files to live outside of root, but also in a discoverable place.  The idea behind the environment variable is so that isolated development environments (virtual, docker compose), can specify the migration directory (should be be more than 1 level deep).

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -6,12 +6,17 @@ Configuration
 
 When you initialize your project using the :doc:`Init Command<commands>`, Phinx
 creates a default file called ``phinx.yml`` in the root of your project directory.
-This file uses the YAML data serialization format.
+This file uses the YAML data serialization format.  This configuration file can
+also exist in different formats, namely as ``phinx.php`` or ``phinx.json``.
 
 If a ``--configuration`` command line option is given, Phinx will load the
-specified file. Otherwise, it will attempt to find ``phinx.php``, ``phinx.json`` or
-``phinx.yml`` and load the first file found. See the :doc:`Commands <commands>`
-chapter for more information.
+specified file.  If the environment variable ``PHINX_CONFIG_DIR`` exists, phinx
+will attempt to file a configuration file within the specified directory.
+Otherwise phinx will attempt to find and load the first file found from the
+current working directory or the first level of directories under the current
+working directory.
+
+See the :doc:`Commands <commands>` chapter for more information.
 
 .. warning::
 
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -186,11 +186,23 @@ abstract class AbstractCommand extends Command
 
         $cwd = getcwd();
 
+        $locateDirs = [];
+
+        if ($envDir = getenv('PHINX_CONFIG_DIR')) {
+            $locateDirs[] = $envDir;
+        }
+
+        $locateDirs[] = $cwd . DIRECTORY_SEPARATOR;
+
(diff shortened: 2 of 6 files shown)


rquadling:
OOI, if PHINX_CONFIG_DIR is specificed, why look beyond that directory? If the file is further down, just set the PHINX_CONFIG_DIR appropriately?

For this to be accepted, we'll need the documentation updated as well as appropriate unit tests.
And rebased to the current dev branch (0.6.x-dev).

ralphschindler:
I can and will provide those things, wanted to start a discussion first to see if the concept/prototype would be well received.

I agree on stopping at the environment provided directory and not scanning cwd. I'll make that change.

Thanks for having a look.

ralphschindler:
tests added, docs added, base changed for PR. please review.

ralphschindler:
@rquadling bump :)

ralphschindler:
@robmorgan ping

rquadling:
Please rebase to 0.7.x

ralphschindler:
I just tried to rebase to 0.7.x-dev, unfortunately, the branch even without my commits doesn't run tests correctly.  There is another PR here that will fix the `getMock()` warnings.  Also, the phpunit.xml.dist file should likely disable integration tests by default (ie: skip things that actually need a database present).

Do you have a timeline for getting 0.7.0 released? I would like to coordinate getting it in shape, and getting my PR working against that branch.  Thanks!

robmorgan:
@ralphschindler I will merge with the existing `0.6.x-dev` branch a do a mini-release with this change and a few others.

robmorgan:
@ralphschindler I've having a problem running this change locally directly from master. It seems its loading the `phinx.php` application stub in `app/phinx.php` by mistake. Can you reproduce this locally?

1. Clone `master`
2. Run `php bin/phinx status`

```
[RuntimeException]                                                      
  PHP file '/Users/robbym/Code/phinx/app/phinx.php' must return an array
```

Unit tests pass though. I changed them slightly to work on Windows.

ralphschindler:
I don't fully understand what this workflow represents.  Is it a valid use case to clone the project and host migrations inside it? Running `php bin/phinx` seems to imply that you're treating the phinx project itself as the project hosting the migrations (is this valid?)

Finding `app/phinx.php` and expecting it to be a phinx configuration file seems to be the right behavior, no?  After all, the traditional use case would be running phinx from `vendor/bin/phinx` (in which phinx would never look inside vendor/ for a matching config file.)

Please advise.

(Alternatively, if you rename app/phinx.php to something like app/console.php, this would alleviate the immediate problem.)

(Additionally, this enhancement was included b/c the original source code had a @todo in it, just not 10 levels deep: https://github.com/robmorgan/phinx/pull/869/files#diff-c9eb06ef069427530fa3f9f516d6bbd0L190)

robmorgan:
@ralphschindler its how most of us actually do Phinx development.

Yeah I'll have to consider if we rename the class. Originally for some crazy idea I can't remember I thought about traversing up, but deeper obviously makes much more sense.

Can you also help me to fix the tests on Windows? I managed to get 2/3 working, see latest build for an example: https://ci.appveyor.com/project/robmorgan/phinx/build/1.0.1515

I'm going to revert this PR whilst we make the changes so I can ship a mini-release.

ralphschindler:
Do you have a channel you chat in about dev stuffs?

rquadling:
@ralphschindler We have now merged all branches into master, so can you rebase onto master please.

PR #876, opened 2016-06-21
Use the correct boolean value for "phinxlog.breakpoint" column in Postgres

mariorez:
This small fix avoid the Exception message:
`SQLSTATE[42804]: Datatype mismatch: 7 ERROR:  column "breakpoint" is of type boolean but expression is of type integer`

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -460,7 +460,7 @@ abstract class PdoAdapter implements AdapterInterface
     {
         $this->query(
             sprintf(
-                'UPDATE %s SET breakpoint = CASE breakpoint WHEN 1 THEN 0 ELSE 1 END WHERE version = \'%s\';',
+                'UPDATE %s SET breakpoint = CASE breakpoint WHEN true THEN false ELSE true END WHERE version = \'%s\';',
                 $this->getSchemaTableName(),
                 $migration->getversion()
             )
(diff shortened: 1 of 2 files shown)


rquadling:
Seems that using '0' and '1' should work, but 0 and 1 don't.

I suspect that `phinx breakpoint -e development` also fails? I've not got a postgres installation locally. Can check tomorrow.

mariorez:
@rquadling You are right.
I "pushed" another commit that fix this issue for `phinx breakpoint` command.
And also test it (with success) in Postgres and MySQL :)

rquadling:
Thank you. I'll be able to check this tomorrow.

rquadling:
`true`/`false` doesn't work for sqlite, but `'0'`/`'1'` seems to work OK for sqlite/mysql/sqlsrv/postgres.

rquadling:
It looks like `'1'` and `'0'` works with sqlite/mysql/mssql, but not pgsql.
It looks like `true` and `false` works with mysql/pgsql, but not mssq/sqlite.

A ``castToBool()`` method solves all of this and allows for standardisation of the breakpoint related queries into the PDOAdapter.

Take a look at https://github.com/rquadling/phinx/tree/fix_breakpoint and see what you think.

mariorez:
I really prefer your implementation approach to resolve the "boolean" problem.
It solves this issue now and may be helpful for other future needs.
I suggest to close my PR and open a new one with your branch:
https://github.com/rquadling/phinx/tree/fix_breakpoint

Congratulations Richard. Well done :+1:

rquadling:
Can we get away with a patch release as this does introduce an API change (new method in adaptor interface).

The method could be removed from the interface and the PDOAdapter variant exists and can be overridden in a concrete adaptor if needed - like it is for postgres.

That preserves API and doesn't inhibit anything moving forward.

rquadling:
Closing this PR in favour of #877

rquadling:
New release v0.6.2

mariorez:
Thanks... Good Job :+1:

PR #879, opened 2016-06-23
Add update and updateData methods

### LINKED ISSUE #862: Need to be able to "updateData" when Migrating Existing Tables...
I'm working on a PR for this, but basically we need the ability to migrate data in existing tables and rows. So we need a way to "update" a record.

Unfortunately I had this written on a colleague's system via paired programming, but we lost it when he did a composer update ... DOH!

Also @robmorgan, is this something you would consider if I provide all the necessary pieces in the PR? I'm going to submit regardless, just wanted to know.

Thanks!

BallisticPain:
Added update and updateData methods. Also added a shortcut method for updating.

@rquadling @robmorgan 
I'm sure there may be changes needed. Please let me know and I'll implement ASAP.

Thanks for the opportunity to work with you.

Fixes #862

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -28,6 +28,7 @@
  */
 namespace Phinx\Db\Adapter;
 
+use InvalidArgumentException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Output\NullOutput;
@@ -387,6 +388,50 @@ abstract class PdoAdapter implements AdapterInterface
 
         $stmt = $this->getConnection()->prepare($sql);
         $stmt->execute(array_values($row));
+        $this->endCommandTimer();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function update(Table $table, $row, $whereColumns)
+    {
+        $this->startCommandTimer();
+        $this->writeCommand('update', array($table->getName()));
+
+        if (is_string($whereColumns)) {
+            $whereColumns = [$whereColumns];
+        }
+        if (!is_array($whereColumns)) {
+            throw new InvalidArgumentException;
+        }
+
+        $whereValues = [];
+        $whereParams = [];
+        foreach ($whereColumns as $column) {
+            // Store value for where clause
+            $whereValues[$column] = $row[$column];
+            // Set param name and quote column
+            $whereParams[] = $this->quoteColumnName($column) . "=?";
+            // Remove where values as they shouldn't be updated
+            unset($row[$column]);
+        }
+
+        $columns = array_keys($row);
+        $sql = sprintf(
+            "UPDATE %s SET ",
+            $this->quoteTableName($table->getName())
+        );
+        $setColumns = [];
+        foreach ($columns as $col) {
+            $setColumns[] = $this->quoteColumnName($col) . '=?';
+        }
+        $sql .= implode(', ', $setColumns);
+
+        $sql .= " WHERE " . implode(' AND ', $whereParams);
+
+        $stmt = $this->getConnection()->prepare($sql);
+        $stmt->execute(array_merge(array_values($row), array_values($whereValues)));
         $this->endCommandTimer();
     }
 
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -394,27 +394,25 @@ abstract class PdoAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
-    public function update(Table $table, $row, $whereColumns)
+    public function update(Table $table, $row, $whereParams)
     {
         $this->startCommandTimer();
         $this->writeCommand('update', array($table->getName()));
 
-        if (is_string($whereColumns)) {
-            $whereColumns = [$whereColumns];
+        if (!is_array($whereParams)) {
+            throw new InvalidArgumentException('$whereParams must be an array type, "' . gettype($whereParams) . '" was given.');
         }
-        if (!is_array($whereColumns)) {
-            throw new InvalidArgumentException;
-        }
+        // Get column names to validate $whereParams
+        $tableColumns = $table->getColumnNames();
 
-        $whereValues = [];
-        $whereParams = [];
-        foreach ($whereColumns as $column) {
-            // Store value for where clause
-            $whereValues[$column] = $row[$column];
-            // Set param name and quote column
-            $whereParams[] = $this->quoteColumnName($column) . "=?";
-            // Remove where values as they shouldn't be updated
-            unset($row[$column]);
+        $whereColumns = [];
+        foreach ($whereParams as $column => $value) {
+            // Validate column name is valid for this table
+            if (!in_array($column, $tableColumns)) {
+                throw new InvalidArgumentException('$whereParams keys must contain valid column names for this table: ' . $table->getName());
+            }
+            // Quote column and prep unnamed param
+            $whereColumns[] = $this->quoteColumnName($column) . "=?";
         }
 
         $columns = array_keys($row);
@@ -428,10 +426,10 @@ abstract class PdoAdapter implements AdapterInterface
         }
         $sql .= implode(', ', $setColumns);
 
-        $sql .= " WHERE " . implode(' AND ', $whereParams);
+        $sql .= " WHERE " . implode(' AND ', $whereColumns);
 
         $stmt = $this->getConnection()->prepare($sql);
-        $stmt->execute(array_merge(array_values($row), array_values($whereValues)));
+        $stmt->execute(array_merge(array_values($row), array_values($whereParams)));
         $this->endCommandTimer();
     }
 
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -298,6 +298,70 @@ insert methods in your migrations.
     You cannot use the insert methods inside a `change()` method. Please use the
     `up()` and `down()` methods.
 
+Updating Data
+-------------
+
+Phinx makes it easy to update data in your tables. Whilst this feature is
+intended for the :doc:`seed feature <seeding>`, you are also free to use the
+update methods in your migrations. Using this in migrations is especially useful when you are working with
+an existing database.
+
+.. code-block:: php
+
+        <?php
+
+        use Phinx\Migration\AbstractMigration;
+
+        class UpdateStatus extends AbstractMigration
+        {
+            /**
+             * Migrate Up.
+             */
+            public function up()
+            {
+                // updating only one row
+                $singleRow = [
+                    'id'    => 1,
+                    'name'  => 'In Progression'
+                ]
+
+                $table = $this->table('status');
+                $table->setData([$singleRow]); // setData expects an array of rows (a row is an array of fields)
+                $table->updateData('id');
+
+                // updating multiple rows
+                $rows = [
+                    [
+                      'id'    => 2,
+                      'name'  => 'Stopping'
+                    ],
+                    [
+                      'id'    => 3,
+                      'name'  => 'Queueing'
+                    ]
+                ];
+
+                // this is a handy shortcut
+                $this->update('status', $rows, ['id']);
+            }
+
+            /**
+             * Migrate Down.
+             */
+            public function down()
+            {
+                $this->execute('DELETE FROM status');
+            }
+        }
+
+.. note::
+
+    1. The `$whereColumns` **cannot** be updated. However, they **must** be included
+    in the `$row` to properly update.
+
+    2. You cannot use the update methods inside a `change()` method. Please use the
+    `up()` and `down()` methods.
+
 Working With Tables
 -------------------
 
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -303,8 +303,8 @@ Updating Data
 
 Phinx makes it easy to update data in your tables. Whilst this feature is
 intended for the :doc:`seed feature <seeding>`, you are also free to use the
-update methods in your migrations. Using this in migrations is especially useful when you are working with
-an existing database.
+update methods in your migrations. Using this in migrations is especially useful
+when you are working with an existing database.
 
 .. code-block:: php
 
@@ -327,7 +327,7 @@ an existing database.
 
                 $table = $this->table('status');
                 $table->setData([$singleRow]); // setData expects an array of rows (a row is an array of fields)
-                $table->updateData('id');
+                $table->updateData(['id' => 1]);
 
                 // updating multiple rows
                 $rows = [
@@ -342,7 +342,7 @@ an existing database.
                 ];
 
                 // this is a handy shortcut
-                $this->update('status', $rows, ['id']);
+                $this->update('status', $rows, ['id' => 1]);
             }
 
             /**
@@ -356,8 +356,8 @@ an existing database.
 
 .. note::
 
-    1. The `$whereColumns` **cannot** be updated. However, they **must** be included
-    in the `$row` to properly update.
+    1. The `$whereParams` is an associative array. Ex. `$whereParams = ['columnName' => 'filterValue'];`.
+    This provides the ability to update a value that you are also filtering by.
 
     2. You cannot use the update methods inside a `change()` method. Please use the
     `up()` and `down()` methods.
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -286,10 +286,10 @@ interface AdapterInterface
      * 
      * @param Table $table where to update data
      * @param array $row of data to be updated
-     * @param array|string $whereColumns specifying which row to update
+     * @param array $whereParams filtering which row to update
      * @return void
      */
-    public function update(Table $table, $row, $whereColumns);
+    public function update(Table $table, $row, $whereParams);
 
     /**
      * Quotes a table name for use in a query.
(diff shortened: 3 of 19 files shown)


BallisticPain:
1 failure is due to `ClassName::CLASS` on PHP 5.4.
1 failure is due to `createMock` (as getMock is deprecated) on PHP 5.5.

I'm guessing I need to fix that to work with 5.4 and 5.5?

Thanks,
Jarvis

BallisticPain:
So I guess I now know why you're using getMock. I was going to submit a quick PR to replace those with createMock. Not now. :-P

rquadling:
PHP 5.4 is still the minimum requirement for Phinx.

BallisticPain:
Found a bug in my own further development of my legacy system. Fixed it.

BallisticPain:
@rquadling Any thoughts / ideas on what I have attempted here? Any improvements?

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -393,6 +393,36 @@ public function insert(Table $table, $row)
...
+        $sql = sprintf(
+            "UPDATE %s SET ",
+            $this->quoteTableName($table->getName())
+        );
+        $setColumns = [];
+        foreach ($columns as $col) {
+            $setColumns[] = $this->quoteColumnName($col) . '=?';
+        }
+        $sql .= implode(', ', $setColumns);
+        $sql .= " WHERE " . $primaryKeyName . "='" . $pKey . "'";
```
rquadling: 1. Does $primaryKeyName needs to be escaped appropriately?
2. Should $pKey be re-added to the end of $row once the update part of the query has been generated? No need to quote a string and if the string contains a `'` ... 💣

rquadling:
Added a few comments inline.

This is looking good though.

BallisticPain:
Awesome. I'll work on this on Monday...
Thanks for your feedback @rquadling.

BallisticPain:
Well it's late on Monday, but I made it in today! Yay.

BallisticPain:
@rquadling 
Does my latest commit fix your comments? BTW good catches on the SQL injection items. I can't believe I missed them... thank goodness for code reviews!

rquadling:
I've got a few other things I want to see ready by the weekend - though that may travel to Monday/Tuesday as daughter's birthday at the weekend.

rquadling:
Oh. Can you squash the commits? Makes things cleaner for merging.

rquadling:
A thought. The supplying of 'id' as the primary key name ... I have just checked one of our servers. Databases comprising over 60,000 columns. Databases generated by different developers over time 15 years. Just under half the tables have an 'id' column as the primary key. Different application server (over 280,000 columns, 5,000 tables) no column called 'id' at all. You'll probably find as much variation in this as I.

I would suggest though, that most people would have to supply a `$primaryKeyName` anyway. So, dropping the default would enforce the requirement of supplying a `$primaryKeyName`. 

The only place we supply a value is the default for the referenced columns for `\Phinx\Db\Table::addForeignKey()` .. and I think that suffers the same issue is that in most cases adding a foreign key will require column names that are rarely going to be 'id'.

Changing that, at this stage isn't a possibility, but something that I'll look into. Defaults are great if they are useful. If not, and you forget to override, unintended consequences ensue. Whereas a mandatory parameter is caught much earlier on.

BallisticPain:
Happy Birthday to your daughter!

I'll squash the commits.

You're correct, I did that because by default phinx adds an 'id' column to each table unless otherwise specified. I have no issues with making it non-default. I've found two main paradigms to be common `id` is one, and the other is `tablename_id`. My recent endeavor is working with a legacy system and they used `K` most of the time... lol

Since it won't have a default, I'll check it for null or non-string values and throw an `InvalidArgumentException`. Though really I can just check for non-string values. Sitting here thinking about that scenario though... I don't account for tables with multiple primary key columns... should I handle this scenario as well? Which would then require an array or a string right? I see more test writing... :)

I'll work on the above, and await your input @rquadling, after which I'll squash the commits.

Thanks for your help!

rquadling:
The problem is, if you think it, you have to code it.

Multiple column primary keys is of course necessary, but also, considering this is a generic update handler, it may NOT be even primary or even indexed!

"Update the paid flag on orders where the account is x and the paid flag is not true" for example.

`$this->table('orders')->update(['paid'=>true], ['account'=>'x', 'paid'=>false]);`

You then get into the issue of how to add `where orderDate < '2016-01-01'`.

Until we have a full ORM behind Phinx, I think we are going to have to offer update() as an option for simple updates. Multiple keys is fine, but more complex where clauses (including OR or IN, etc.) is beyond this. But you can be pretty sure more will be wanted on this.

BallisticPain:
@rquadling 
LOL, yes sir. I assumed as much, and because of that I didn't want to think past the multiple fields. ;)

Do I need to allow string or array or just require an array whether it's a single value or not?

So are ORM(s) being contemplated? I guess once you start getting more complex like that it does start to make more sense as they have already solved those problems.

rquadling:
I've not been looking into an ORM specifically. I know Rob Morgan was, but I think time constraints have limited his ongoing involvement (which is why I'm here!).

Obviously Doctrine would be a strong contender, but I suspect there are others equally as useful out there. Of course, rolling our own is a possibility, but then at what stage do we become more than just a migration tool with ORM features and not an ORM with migration support? I don't know the answer to that.

I think, for the time being, a single primary key is probably enough. Pretty much every pattern after that gets that level of complexity which needs a better discussion.

BallisticPain:
Well I fixed it easily enough to `whereColumns` and no default, however now when I run tests I get `Invalid parameter number: mixed named and positional parameters`. Previously I used named and positional parameters with the Primary key stuff, and it wasn't complaining. So I guess I can revert back to that if you're saying I don't need to account for multiple WHERE columns.

@rquadling Interesting on the ORM front. Maybe it would be better to have some "Interfaces" that could then be used for anyone to use their favorite ORM. I'm still using Doctrine Fixtures for a portion of my "demo" data instead of seeds as it was easier. :)

rquadling:
For the Where clause columns, it may be simpler to follow the same approach that you've used for the main.

With regard to appropriate interfaces, then create a separate PR. Let's see what grabs the userbase.

BallisticPain:
@rquadling 
Okay never got this committed on Friday... so I hope I'm not too late. I'm awaiting the testing confirmation, but I have squashed the commits...

Hope you have a great weekend!

BallisticPain:
@rquadling 
I'm stuck a bit on these failing tests. I believe it should be calling `update` once, not `updateData` as the `PdoAdapter` doesn't have an `updateData` (which is why it's likely failing). However, if I change it to update as I expect I get another error all together saying that the `Table` object doesn't match the `Table` object.

This is the other error as seen in 43941b9:
```
1) Test\Phinx\Db\Adapter\TablePrefixAdapterTest::testUpdateData
Expectation failed for method name is equal to <string:update> when invoked 1 time(s)
Parameter 0 for invocation Phinx\Db\Adapter\PdoAdapter::update(Phinx\Db\Table Object (...), Array (...), Array (...)) does not match expected value.
```

BallisticPain:
I should say that the tests do not act the same on my system so it's hard to test without doing a ton of commits changing things.

BallisticPain:
I think I have realized why it wasn't working, but I can't tell you why it was working for some of the versions... :) New commit / squash coming soon.

BallisticPain:
@rquadling 
Okay I think it's finally ready!! Sorry for the delay.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -393,6 +394,50 @@ public function insert(Table $table, $row)
...
+
+        $whereValues = [];
+        $whereParams = [];
+        foreach ($whereColumns as $column) {
+            // Store value for where clause
+            $whereValues[$column] = $row[$column];
+            // Set param name and quote column
+            $whereParams[] = $this->quoteColumnName($column) . "=?";
+            // Remove where values as they shouldn't be updated
+            unset($row[$column]);
```
rquadling:
```sql
UPDATE table SET key = value1 WHERE key = value2;
```

The where clause has no bearing on the columns being updated. The unset() would remove this capability.

rquadling: Ahhh.. I'm REALLY torn about how to handle the where clause. I initially thought it may be necessary to ONLY allow primary key values to keep things simple. But then that is next to useless for the majority of data changes where you want to filter on other columns. I think I'm running in circles on this.

BallisticPain: You're correct, I was still stuck in the you're going to use a Primary Key to update a row, and you're not going to update the Primary Key...

I understand running circles on it. Is there a way we could even force/enforce primary key selection?

rquadling:
Mysql : http://stackoverflow.com/questions/893874/mysql-determine-tables-primary-key-dynamically
sqlite : http://stackoverflow.com/questions/10472103/sqlite-query-to-find-primary-keys
mssql : http://stackoverflow.com/questions/3930338/sql-server-get-table-primary-key-using-sql-query
postgres : http://stackoverflow.com/questions/1214576/how-do-i-get-the-primary-keys-of-a-table-from-postgres-via-plpgsql

Using INFORMATION_SCHEMA seems to be a good way - need to manually test that with tables with compound primary keys.

That may work on all DBs.

BallisticPain:
Is this an indication to head in that direction? If so, I could re-add my unset as you wouldn't be updating primary key columns...

rquadling:
If we think `update()` can only work with filtering by the primary key, then extracting that info is required to enforce this rule.

As for the `unset()`, you could be updating a primary key. Especially if you have cascading foreign key constraints.

BallisticPain:
@rquadling 
You're right. I didn't think about cascading keys. We can filter using other fields. I will go through another vamp of the methods to change `whereColumns` to `whereParams` so that an array can be provided as `$whereParams = ['columnName' => 'filterValue'];` This way they can update the columns in the `$row`, but also filter by the same columns with their original values.

We'll get this thing done ... eventually! Thanks for all of your input and guidance!

BallisticPain:
@rquadling 
I've noticed the use of `$var = array();` versus `$var = [];`. Is this something I should follow? I personally usually choose the `[]` option, and have double checked it's valid for PHP 5.4. However, I want to follow the guidelines that y'all would prefer. (I'm getting close to my next commit).

Thanks!

BallisticPain:
@rquadling 
Any issue with me adding a `$table->getColumnNames()` since I don't need column objects and it would save me a loop in my `update` method? I'm going to add it, and remove if you say no. ;-)

BallisticPain:
@rquadling 
So the more I think about this the less I'm thinking the whereParams will work... Let me see if I can articulate this...

With multiple rows of data to update, you cannot specify a set of whereParams for each `$row` of update... I started seeing this when I started modifying my tests, my `$whereParams` would suddenly modify 2 rows at once, and then again the second time (not useful). I think I need the ability (at times) to be able to specify a `$whereParams` per `$row`... That being the case... either we don't need the `updateData` and the use of `setData` which means you would have to do it through the adapter. (likely the simplest solution) Or, we would allow them to specify a `whereParams` key in the `$row`, and parse that out into the parameter on the `update` method.

Now I await your further input. :)

BallisticPain:
I decided to go ahead and commit what I have thus far to see if you have any qualms with it, other than the above questions/comments.

rquadling:
As PHP 5.4 is the minimum, [] is OK for arrays.

rquadling:
Is there any reason to not use INFORMATION_SCHEMA all the time? Simpler and consistent across the majority of DBs.

rquadling:
As we already have \Phinx\Db\Adapter\AdapterInterface::getColumns()`, I'd recommend reusing it.

BallisticPain:
I was originally re-using it, but then I had to loop through that and grab just the names... Are you opposed to adding the `getColumnNames` or just suggesting? ;) You say the word and I'll add my loop back and get rid of the `getColumnNames`, and then we just need to decide about the `whereParams` ... 

My thought is that if `$row` contains a key called `phinx_whereParams` or something like that, we use that instead of what might be a "default" whereParams supplied to the `update` method. I wouldn't necessarily have to remove the `updateData` method, but not sure if it makes much sense to have it.

rquadling:
How about having a new method that calls the old method and filters accordingly. That way, if anything changes regarding column, we don't have 2 retrieval methods to amend.

BallisticPain:
@rquadling 
I can do that... Had thought of that at first, but wasn't sure if it would be faster having less columns in the select statement, but probably not enough to warrant the possible future maintenance. I'll get on that.

What about the whereParams per row?

rquadling:
If you have multiple rows of data to update, surely you would supply the where clause with the appropriate data being amended.

`UPDATE table SET col1=val1 WHERE col2=val2;`

That may affect 1 row or many rows.

Can you supply some examples of the issue? I'm not sure I see it.

If you are only supplying data (col1=val1), than that applies to ALL rows, and would be errored by the DB for FK or duplicate key issues.

BallisticPain:
So I may be over thinking the issue... I'm trying to really think of a good example, but I guess if you want different where parameters for *some* of the rows of data, but you have processed ALL of the data... Like what I'm doing ... I'm going through and generating UUIDs for each and every record so I really need it to update each and every row, but I'm doing that 1000 rows at a time.

That make more sense?

rquadling:
The columns in the update's SET clause do not need to be present in the WHERE clause.

The more I think about this, unless we have appropriate clause builder routines, this feature is going to not be worth the effort, considering you can easily write queries to capture the exact intent.

Take a single column and a single value. Many different possibilities.

a = b
a IN (b)
a LIKE 'b%'
a LIKE '%b'
a LIKE '%b%'

and then their NOT variations. So, for a single nvp, potentially 10 different representations. How to express that simply and not use SQL.

With more than 1 nvp, are you AND or OR?

And this scales to silly numbers very quickly when you introduce more pairs. 


I think, without a very clear and defined goal, I'm not there is an easy way out. Maybe introducing an optional dependency on a query builder would be the best option here.

Use Phinx and you get row update based solely on primary key (no primary key, no updating).

If you want to use more complex queries, then a builder is required.

So either we pick one or we abstract from several a common interface and allow the user to decide which one they wish to use.

What do you think?

BallisticPain:
You are correct the columns in the SET clause do not need to be in the WHERE clause. My example was the need to update each row with a separate value, and the only way I can see of doing that is allowing a special key in the $row data to allow for that.

I fear ultimately you are correct that we are only dealing with a subset of users and to fully get everyone we will want some sort of a query builder. However, I also believe we would service a large portion of that subset with the simple "AND" where type. Also instead of a query builder, we could add a third column to the `whereParams` that would allow you to specify the operator... I know that's getting close to a query builder... which a QB would simplify the process quite a lot.

I think it would be worth looking at a few in order to determine if we want to have a common interface that can then have "adapters" to each one... or we through the same process can pick one.

rquadling:
Is it really worth the effort?

Considering that if a user is already using an ORM, then they can use it in Phinx already.

Step 1 - Create their own migration base class from `\Phinx\Migration\AbstractMigration`.
```php
<?php
namespace MyApp\Migrations;
abstract class AbstractMigration extends \Phinx\Migration\AbstractMigration
{
  /**
   * @var DB $db
   */
  protected $db;
  public function init()
  {
    // Instantiate their DB here, preferably from a container so
    // they get a reference to it for subsequent migrations.
    $this->db = DI::make('DB');
  }
}
```

Step 2 - Set the `migration_base_class` config option to `MyApp\Migrations\AbstractMigration`.

Step 3 - Use `$this->db` in the migrations, utilising all the models and repos you have in play.

BallisticPain:
@rquadling 
Just got back from vacation and going through all of this again. I was unaware that I could integrate my ORM by making my own abstract migration... That's an interesting concept, and if that's the case, then you are probably right this will end up being unnecessary though it sounds as though @robmorgan was still considering some of the functionality I have mentioned here...

I'm going to explore this a bit... Thanks for suggesting it!

robmorgan:
@BallisticPain no I'd like to unofficially make Phinx an ORM considering many people already add it to their projects and then add an additional ORM to make update/delete queries etc. Thus they could avoid multiple PDO connections, db config, Composer dependencies etc. As much as I hate adding new features, this is one I think is necessary.

rquadling:
@robmorgan, some ideas/suggestions/thoughts. None of this is a particular direction that I would go in, so very happy to be directed by what you feel is appropriate.

So, some thoughts.

I think the option of swapping out phinx's adapters to use an external ORM would be more efficient. If (for example) my app is Doctrine based, with all the beauty that it provides, it would make sense to be able to "just use it" for my migrations also. If I were to use a different DBAL/ORM, could we provide Phinx2xxxx wrappers?

As there is a good amount of abstraction within Phinx already, is this a potential possibility?

And, more importantly, easier to achieve and support?

There is another factor here. Migrations are not "core" to a product (sure, they are core to THIS product). As a sitting-on-the-edge utility, the amount of work a user has to go through may be a little more than they need for some features in their main product.

Having Phinx work the way Phinx works (for better or worse) is OK I think. So rather than integrating/supporting an external ORM, we do our own.

I would say that that would be both time consuming to develop/test/maintain and be non DRY. There are already many well developed/tested/maintained DBAL/ORMs out there. Can't we "just" pick one?

BallisticPain:
@rquadling Good points...
@robmorgan Look forward to your response(s)...

My personal opinion is that if Phinx could create adapters or even the interface for adapters and maybe host some of the primary adapters and allow others to "plugin" to the ORM/DBAL systems that are out there it will keep y'all agnostic of the other tools in someone's toolbox, but give that extra bit of functionality without having to add anything additional to Phinx core.

I like the idea of Phinx sticking to it's core utility which is migrations. Though I definitely could see some value in expansion. If y'all believe you can provide additional or better value to the existing ORM/DBAL's then I say go for it.

Thanks for letting me be a part of this conversation and attempting to contribute to your project. Proud to be working with y'all.

BallisticPain:
@rquadling @robmorgan 
Thank you both VERY much for your time on my PR and this project. It seems that at this time, this feature is too complex or too limited to implement currently. I don't have anytime left to continue it's integration. I'm closing this PR. Hopefully it will be useful to those who come after. I must revert my branch back to what's working for our project in order to move forward with my project and also keep up with the latest changes here.

PR #880, opened 2016-06-27
Fix - removeColumn in change() instead of up()

skorpa:
Method removeColumn() is not supported in change() method according to http://docs.phinx.org/en/latest/migrations.html#dropping-a-column.


rquadling:
Thank you for this. Can you rebase it on the current dev branch (0.6.x-dev).

PR #883, opened 2016-06-29
Fix - removeColumn in change() instead of up()

skorpa:
Method removeColumn() is not supported in change() method according to http://docs.phinx.org/en/latest/migrations.html#the-change-method.


rquadling:
Thank you.

PR #886, opened 2016-07-04
Remove outdated comment

coatesap:
`save()` now calls `update()` or `create()` behind the scenes, and hence should be the preferred method.


coatesap:
Retracting due to discussion in https://github.com/robmorgan/phinx/issues/851#issuecomment-223882883 as historic use of `$this->table('new_table')` instead of `$this->createTable('new_table')`.

PR #888, opened 2016-07-05
AdapterFactory: fix typo in error message

qyanu:
I don't think I'm wrong. But who knows?


rquadling:
Unfortunately, you are wrong.

rquadling:
In English "Subject must do action", "Subject must exhibit behaviour". Passive (I think).

To have "Subject must be doing action", "Subject must be exhibiting behaviour" is not how we speak or write.

rquadling:
OMG!!! Lesson to self. Don't code when asleep.

PR #889, opened 2016-07-05
AdapterInterface: also provide method to quote string values

qyanu:
strictly speaking, a concrete migration must not assume, that
the used adapter is subclass of PdoAdapter (although currently
always is), thus using $this->getAdapter()->getConnection()->quote(..)
is a clutch.
Instead, AdapterInterface should provide the proposed quoteStringValue(...)
function for quoting string values.

Although I think this will not be used often, i would
  a) consider AdapterInterface incomplete without it,
  b) will it indeed be necessary in exceptional cases.

### DIFF
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -298,6 +298,14 @@ interface AdapterInterface
     public function quoteColumnName($columnName);
 
     /**
+     * Quotes a string value for use in an sql expression.
+     *
+     * @param string $stringValue the string
+     * @return string
+     */
+    public function quoteStringValue($stringValue);
+
+    /**
      * Checks to see if a table exists.
      *
      * @param string $tableName Table Name
--- a/src/Phinx/Db/Adapter/AdapterWrapper.php
+++ b/src/Phinx/Db/Adapter/AdapterWrapper.php
@@ -324,6 +324,14 @@ abstract class AdapterWrapper implements AdapterInterface, WrapperInterface
     /**
      * {@inheritdoc}
      */
+    public function quoteStringValue($stringValue)
+    {
+        return $this->getAdapter()->quoteStringValue($stringValue);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function hasTable($tableName)
     {
         return $this->getAdapter()->hasTable($tableName);
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -333,6 +333,16 @@ abstract class PdoAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
+    public function quoteStringValue($stringValue)
+    {
+        // Note: for example with postgres and mysql, correctly quoting
+        // a string value depends on server and/or connection settings.
+        return $this->getConnection()->quote($stringValue);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function execute($sql)
     {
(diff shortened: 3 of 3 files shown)


qyanu:
i wanted to write a test for the string quoting, but this turned out to be quote difficult.
take, for example postgresql: depending on the value of `standard_conforming_strings`, there are different correct outcomes (using php-syntax):

```
# normal stuff
assert($a->quoteStringValue("text")==="'text'");
assert($a->quoteStringValue("te'xt")==="'te''xt'");

# attacker wants to break free
assert($a->quoteStringValue("text\\")==="'text\\'"); # standard_conforming_strings=on
assert($a->quoteStringValue("text\\")==="'text\\\\'"); # standard_conforming_strings=off
```

with mysql, there's a similar setting, with use of `NO_BACKSLASH_ESCAPES`

rquadling:
I like this. As it is an interface change, this will require a new branch. Tagged for 0.7.0

rquadling:
With regard to the tests, it would seem that some sniffing of the flags need to be performed. Not ideal.

qyanu:
I could at least write the test for mysql; with variable sniffing. This would then test whatever the tester has configured at the time.
Additionally, a seach by daylight revealed that the mysql variable `sql_mode` is marked as `Scope: Global,Session` (see https://dev.mysql.com/doc/refman/5.5/en/server-system-variables.html#sysvar_sql_mode )
So maybe the test could set the variable to one value, then test first case, then set the variable to other value and test second case.

Writing the mysql test is on my todo-list and won't be forgotten. However, it might take a month to come around.

rquadling:
That's OK.

rquadling:
For mysql, setting the mode for the session as one of the values supplied by a dataProvider would be ideal.

From https://www.postgresql.org/docs/9.5/static/release-8-1.html#AEN158577

> Add E'' syntax so eventually ordinary strings can treat backslashes literally (Bruce)

> Currently PostgreSQL processes a backslash in a string literal as introducing a special escape sequence, e.g. \n or \010. While this allows easy entry of special values, it is nonstandard and makes porting of applications from other databases more difficult. For this reason, the PostgreSQL project is planning to remove the special meaning of backslashes in strings. For backward compatibility and for users who want special backslash processing, a new string syntax has been created. This new string syntax is formed by writing an E immediately preceding the single quote that starts the string, e.g. E'hi\n'. While this release does not change the handling of backslashes in strings, it does add new configuration parameters to help users migrate applications for future releases:

>   standard_conforming_strings — does this release treat backslashes literally in ordinary strings?

>   escape_string_warning — warn about backslashes in ordinary (non-E) strings


This is for a release date of 2005-11-08.

I wonder if it would be OK to say that we only do things the modern way for postgres? Add appropriate documentation indicating this limitation to the new method.

rquadling:

JayPHP:
@qyanu Looking to close this due to age and lack of tests. If you would like this kept open, please let me know and raise a new issue.

qyanu:
@JayPHP I found the bug, i reported the bug. i provided a proposal-patch to fix the bug.
If that does not meet your guidelines, it is your decision to throw away or merge anyway.

Frankly, i don't have time to do anything further.
(Sorry if that seems arrogant, it isn't meant to be. but that's the reality of my day).

JayPHP:
@qyanu Understood, it's just waiting on tests

> Writing the mysql test is on my todo-list and won't be forgotten. However, it might take a month to come around.

I'll leave it open for a day or two to see if anyone's interested in taking this up.

dereuromark:
@ ping at anyone who wants to contribute or help getting this merged.

JayPHP:
Happy to re-open if someone would like to pick it up

PR #899, opened 2016-07-11
Add missing quotes. Fixed #887

### LINKED ISSUE #887: MySQL enum values with special characters
Hello

I'm trying to store some special values in a MySQL enum field like ' (single quote) and \ (backslash). Here is an example:

``` php
$this->table("test")->changeColumn('code', 'enum', array('values' => array('1','22','33','\'','"','\\','/',',',',,')))->update();`
```

**Error message**

`[PDOException]
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that
   corresponds to your MariaDB server version for the right syntax to use near '"', '\', '/', ',', ',,') NULL COMMENT
   '4' AFTER`deleted`' at line 1
`  
**Query**

``` sql
ALTER TABLE `test` CHANGE `code` `code` ENUM('1', '22', '33', ''', '"', '\', '/', ',', ',,') NULL AFTER `deleted`
```

**Expected query:**

``` sql
ALTER TABLE `test` CHANGE `code` `code` ENUM('1', '22', '33', '\'', '"', '\\', '/', ',', ',,') NULL AFTER `deleted`
```

To make the query working I've added a backslash before ' and \.
' => \'
\ => \\
I think it's an small issue with a missing value escaping. How could I fix it?

Version: 0.6.2

odan:


dereuromark:
This sounds like an important issue to be fixed. Anything missing to get this finished?

spronkey:
This still appears to be a problem, I'm having an issue with 'values' => ['Don\'t Know', 'Yes', 'No'] in an enum definition returning a syntax error where the ' in Don't Know is not escaped.

PR #901, opened 2016-07-13
Rollbacks by start time

daniel-gomes-sociomantic:
A new Configuration option named `version_order` was added. Its possible values are:

* `creation-time` (the default): causes the Rollback and Status commands to order the executed migrations by their creation times (aka Migration IDs or Names).
* `start-time`: causes the Rollback and Status commands to order the executed migrations by their start times (aka Execution Times).

This means that when invoking the Rollback command:
    
- With a `start-time` version order, the last executed migration will be rollbacked.
- With a `start-time` version order and a `-d (date)` option, the migration start times will be used to determine the target version to rollback to. In other words, all migrations which were executed after that date will be rollbacked.
- With a `start-time` version order and a `-t (target version)` option, all migrations which were executed after the target version was executed will be rollbacked.

In terms of testing:

- More test cases were added.
- General test improvements were done (like adding the expected arguments to the underlying Manager calls in `RollbackTest` and `StatusTest`.

Other notes:
- The old `rollback` and `rollbackToDateTime` methods were merged into a single `rollback` method that handles all cases and version orders.
- This PR replaces https://github.com/robmorgan/phinx/pull/745.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -127,12 +127,31 @@ Specifying 0 as the target version will revert all migrations.
 
         $ phinx rollback -e development -t 0
 
+To rollback all migrations to a specific date then use the ``--date``
+parameter or ``-d`` for short.
+
+.. code-block:: bash
+
+        $ phinx rollback -e development -d 2012
+        $ phinx rollback -e development -d 201201
+        $ phinx rollback -e development -d 20120103
+        $ phinx rollback -e development -d 2012010312
+        $ phinx rollback -e development -d 201201031205
+        $ phinx rollback -e development -d 20120103120530
+
 If a breakpoint is set, blocking further rollbacks, you can override the
 breakpoint using the ``--force`` parameter or ``-f`` for short.
 
 .. code-block:: bash
 
         $ phinx rollback -e development -t 0 -f
+
+.. note::
+
+        When rolling back, Phinx orders the executed migrations using 
+        the order specified in the ``version_order`` option of your 
+        ``phinx.yml`` file.
+        Please see the :doc:`Configuration <configuration>` chapter for more information.
 
 The Status Command
 ------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -280,3 +280,12 @@ The aliased classes will still be required to implement the ``Phinx\Migration\Cr
     aliases:
         permission: \Namespace\Migrations\PermissionMigrationTemplateGenerator
         view: \Namespace\Migrations\ViewMigrationTemplateGenerator
+
+Version Order
+------
+
+When rolling back or printing the status of migrations, Phinx orders the executed migrations according to the 
+``version_order`` option, which can have the following values:
+
+* ``creation-time`` (the default): migrations are ordered by their creation time, which is also part of their filename.
+* ``start-time``: migrations are ordered by their start time, also known as execution time.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -287,5 +287,5 @@ Version Order
 When rolling back or printing the status of migrations, Phinx orders the executed migrations according to the 
 ``version_order`` option, which can have the following values:
 
-* ``creation-time`` (the default): migrations are ordered by their creation time, which is also part of their filename.
-* ``start-time``: migrations are ordered by their start time, also known as execution time.
+* ``creation`` (the default): migrations are ordered by their creation time, which is also part of their filename.
+* ``execution``: migrations are ordered by their execution time, also known as start time.
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -39,6 +39,16 @@ use Symfony\Component\Yaml\Yaml;
 class Config implements ConfigInterface
 {
     /**
+     * The value that identifies a version order by creation time.
+     */
+    const VERSION_ORDER_CREATION_TIME = 'creation-time';
+
+    /**
+     * The value that identifies a version order by start time.
+     */
+    const VERSION_ORDER_START_TIME = 'start-time';
+
+    /**
      * @var array
      */
     private $values = array();
@@ -55,6 +65,8 @@ class Config implements ConfigInterface
     {
         $this->configFilePath = $configFilePath;
         $this->values = $this->replaceTokens($configArray);
+
+        $this->setVersionOrder();
     }
 
     /**
@@ -292,6 +304,58 @@ class Config implements ConfigInterface
 
         return $this->values['templates']['class'];
      }
+
+    /**
+     * Get the version order.
+     *
+     * @return string
+     */
+    public function getVersionOrder()
+    {
+        return $this->values['version_order'];
+    }
+
+    /**
+     * Set the version order.
+     * @param string $versionOrder The Version Order to set.
+     */
+    public function setVersionOrder($versionOrder = null)
+    {
+        if (!isset($versionOrder)) {
+            $versionOrder = isset ( $this->values['version_order'] ) ? $this->values['version_order'] : null;
+        }
+
+        if (!isset($versionOrder)) {
+            // if the configuration value is missing we use the default version order (creation time)
+            $this->values['version_order'] = self::VERSION_ORDER_CREATION_TIME;
+            return;
+        }
+
+        $validVersionOrders = array(self::VERSION_ORDER_CREATION_TIME, self::VERSION_ORDER_START_TIME);
+
+        if (!in_array($versionOrder, $validVersionOrders)) {
+            throw new \RuntimeException(sprintf(
+                'Invalid version_order configuration option: %s. Valid values: %s',
+                $versionOrder, implode (' or ', $validVersionOrders)
+            ));
+        }
+
+        $this->values['version_order'] = $versionOrder;
+    }
+    
+    /**
+     * Is version order creation time?
+     *
+     * @return boolean
+     */
+    public function isVersionOrderCreationTime()
+    {
+        $versionOrder = $this->getVersionOrder();
+
+        return $versionOrder == self::VERSION_ORDER_CREATION_TIME;
+    }
+
+    
 
     /**
      * Replace tokens in the specified array.
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -41,12 +41,12 @@ class Config implements ConfigInterface
     /**
      * The value that identifies a version order by creation time.
      */
-    const VERSION_ORDER_CREATION_TIME = 'creation-time';
+    const VERSION_ORDER_CREATION_TIME = 'creation';
 
     /**
-     * The value that identifies a version order by start time.
+     * The value that identifies a version order by execution time.
      */
-    const VERSION_ORDER_START_TIME = 'start-time';
+    const VERSION_ORDER_EXECUTION_TIME = 'execution';
 
     /**
      * @var array
@@ -331,7 +331,7 @@ class Config implements ConfigInterface
             return;
         }
 
-        $validVersionOrders = array(self::VERSION_ORDER_CREATION_TIME, self::VERSION_ORDER_START_TIME);
+        $validVersionOrders = array(self::VERSION_ORDER_CREATION_TIME, self::VERSION_ORDER_EXECUTION_TIME);
 
         if (!in_array($versionOrder, $validVersionOrders)) {
             throw new \RuntimeException(sprintf(
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -65,8 +65,6 @@ class Config implements ConfigInterface
     {
         $this->configFilePath = $configFilePath;
         $this->values = $this->replaceTokens($configArray);
-
-        $this->setVersionOrder();
     }
 
     /**
@@ -312,37 +310,13 @@ class Config implements ConfigInterface
      */
     public function getVersionOrder()
     {
+        if (!isset($this->values['version_order'])) {
+            return self::VERSION_ORDER_CREATION_TIME;
+        }
+
         return $this->values['version_order'];
     }
 
-    /**
-     * Set the version order.
-     * @param string $versionOrder The Version Order to set.
-     */
-    public function setVersionOrder($versionOrder = null)
-    {
-        if (!isset($versionOrder)) {
-            $versionOrder = isset ( $this->values['version_order'] ) ? $this->values['version_order'] : null;
-        }
-
-        if (!isset($versionOrder)) {
-            // if the configuration value is missing we use the default version order (creation time)
-            $this->values['version_order'] = self::VERSION_ORDER_CREATION_TIME;
-            return;
-        }
-
-        $validVersionOrders = array(self::VERSION_ORDER_CREATION_TIME, self::VERSION_ORDER_EXECUTION_TIME);
-
-        if (!in_array($versionOrder, $validVersionOrders)) {
-            throw new \RuntimeException(sprintf(
-                'Invalid version_order configuration option: %s. Valid values: %s',
-                $versionOrder, implode (' or ', $validVersionOrders)
-            ));
-        }
-
-        $this->values['version_order'] = $versionOrder;
-    }
-    
     /**
      * Is version order creation time?
      *
--- a/src/Phinx/Console/Command/Rollback.php
+++ b/src/Phinx/Console/Command/Rollback.php
@@ -109,7 +109,7 @@ EOT
         }
         
         $versionOrder = $this->getConfig()->getVersionOrder();
-        $output->writeln('<info>ordering by </info>' . $versionOrder);
+        $output->writeln('<info>ordering by </info>' . $versionOrder . " time");
 
         // rollback the specified environment
         if (null === $date) {
--- a/src/Phinx/Console/Command/Rollback.php
+++ b/src/Phinx/Console/Command/Rollback.php
@@ -137,8 +137,7 @@ EOT
     public function getTargetFromDate($date)
     {
         if (!preg_match('/^\d{4,14}$/', $date)) {
-            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
-                'part, but the missing part(s) must all be in the end of the date string)');
+            throw new \InvalidArgumentException('Invalid date. Format is YYYY[MM[DD[HH[II[SS]]]]].');
         }
 
         // what we need to append to the date according to the possible date string lengths
@@ -156,15 +155,13 @@ EOT
         }
 
         if ($target === null) {
-            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
-                'part, but the missing part(s) must all be in the end of the date string)');
+            throw new \InvalidArgumentException('Invalid date. Format is YYYY[MM[DD[HH[II[SS]]]]].');
         }
 
         $dateTime = \DateTime::createFromFormat('YmdHis', $target);
 
         if ($dateTime === false) {
-            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
-                'part, but the missing part(s) must all be in the end of the date string)');
+            throw new \InvalidArgumentException('Invalid date. Format is YYYY[MM[DD[HH[II[SS]]]]].');
         }
 
         return $dateTime->format('YmdHis');
--- a/src/Phinx/Console/Command/Rollback.php
+++ b/src/Phinx/Console/Command/Rollback.php
@@ -31,6 +31,7 @@ namespace Phinx\Console\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Phinx\Config\Config;
 
 class Rollback extends AbstractCommand
 {
@@ -61,6 +62,10 @@ The <info>rollback</info> command reverts the last migration, or optionally up t
 If you have a breakpoint set, then you can rollback to target 0 and the rollbacks will stop at the breakpoint.
 <info>phinx rollback -e development -t 0 </info>
 
+The <info>version_order</info> configuration option is used to determine the order of the migrations when rolling back.
+This can be used to allow the rolling back of the last executed migration instead of the last created one, or combined 
+with the <info>-d|--date</info> option to rollback to a certain date using the migration start times to order them.
+
 EOT
              );
     }
@@ -81,14 +86,16 @@ EOT
         $date        = $input->getOption('date');
         $force       = !!$input->getOption('force');
 
+        $config = $this->getConfig();
+
         if (null === $environment) {
-            $environment = $this->getConfig()->getDefaultEnvironment();
+            $environment = $config->getDefaultEnvironment();
             $output->writeln('<comment>warning</comment> no environment specified, defaulting to: ' . $environment);
         } else {
             $output->writeln('<info>using environment</info> ' . $environment);
         }
 
-        $envOptions = $this->getConfig()->getEnvironment($environment);
+        $envOptions = $config->getEnvironment($environment);
         if (isset($envOptions['adapter'])) {
             $output->writeln('<info>using adapter</info> ' . $envOptions['adapter']);
         }
@@ -100,17 +107,66 @@ EOT
         if (isset($envOptions['name'])) {
             $output->writeln('<info>using database</info> ' . $envOptions['name']);
         }
+        
+        $versionOrder = $this->getConfig()->getVersionOrder();
+        $output->writeln('<info>ordering by </info>' . $versionOrder);
 
         // rollback the specified environment
-        $start = microtime(true);
-        if (null !== $date) {
-            $this->getManager()->rollbackToDateTime($environment, new \DateTime($date), $force);
+        if (null === $date) {
+            $targetMustMatchVersion = true;
+            $target = $version;
         } else {
-            $this->getManager()->rollback($environment, $version, $force);
+            $targetMustMatchVersion = false;
+            $target = $this->getTargetFromDate($date);
         }
+
+        $start = microtime(true);
+        $this->getManager()->rollback($environment, $target, $force, $targetMustMatchVersion);
         $end = microtime(true);
 
         $output->writeln('');
         $output->writeln('<comment>All Done. Took ' . sprintf('%.4fs', $end - $start) . '</comment>');
+    }
+
+    /**
+     * Get Target from Date
+     *
+     * @param string $date The date to convert to a target.
+     * @return The target
+     */
+    public function getTargetFromDate($date)
+    {
+        if (!preg_match('/^\d{4,14}$/', $date)) {
+            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'part, but the missing part(s) must all be in the end of the date string)');
+        }
+
+        // what we need to append to the date according to the possible date string lengths
+        $dateStrlenToAppend = array(
+            14 => '',
+            12 => '00',
+            10 => '0000',
+            8 => '000000',
+            6 => '01000000',
+            4 => '0101000000',
+        );
+
+        if (isset($dateStrlenToAppend[strlen($date)])) {
+            $target = $date . $dateStrlenToAppend[strlen($date)];
+        }
+
+        if ($target === null) {
+            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'part, but the missing part(s) must all be in the end of the date string)');
+        }
+
+        $dateTime = \DateTime::createFromFormat('YmdHis', $target);
+
+        if ($dateTime === false) {
+            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'part, but the missing part(s) must all be in the end of the date string)');
+        }
+
+        return $dateTime->format('YmdHis');
     }
 }
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -411,7 +411,7 @@ abstract class PdoAdapter implements AdapterInterface
             case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
                 $orderBy = 'version ASC';
                 break;
-            case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
+            case \Phinx\Config\Config::VERSION_ORDER_EXECUTION_TIME:
                 $orderBy = 'start_time ASC, version ASC';
                 break;
         }
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -406,8 +406,18 @@ abstract class PdoAdapter implements AdapterInterface
     public function getVersionLog()
     {
         $result = array();
-        $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY version ASC', $this->getSchemaTableName()));
-        foreach ($rows as $version) {
+
+        switch ($this->options['version_order']) {
+            case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
+                $orderBy = 'version ASC';
+                break;
+            case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
+                $orderBy = 'start_time ASC, version ASC';
+                break;
+        }
+
+        $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY %s', $this->getSchemaTableName(), $orderBy));
+        foreach($rows as $version) {
             $result[$version['version']] = $version;
         }
 
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -414,6 +414,8 @@ abstract class PdoAdapter implements AdapterInterface
             case \Phinx\Config\Config::VERSION_ORDER_EXECUTION_TIME:
                 $orderBy = 'start_time ASC, version ASC';
                 break;
+            default:
+                throw new \RuntimeException('Invalid version_order configuration option');
         }
 
         $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY %s', $this->getSchemaTableName(), $orderBy));
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -115,7 +115,7 @@ class Manager
                 case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
                     $migrationIdAndStartedHeader = "<info>[Migration ID]</info>  Started            ";
                     break;
-                case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
+                case \Phinx\Config\Config::VERSION_ORDER_EXECUTION_TIME:
                     $migrationIdAndStartedHeader = "Migration ID    <info>[Started          ]</info>";
                     break;
             }
@@ -426,7 +426,7 @@ class Manager
         $sortedMigrations = array();
 
         foreach ($executedVersions as $versionCreationTime => &$executedVersion) {
-            // if we have a date (ie. the target must not match a version) and we are sorting by start time, we
+            // if we have a date (ie. the target must not match a version) and we are sorting by execution time, we
             // convert the version start time so we can compare directly with the target date
             if (!$this->getConfig()->isVersionOrderCreationTime() && !$targetMustMatchVersion) {
                 $dateTime = \DateTime::createFromFormat('Y-m-d H:i:s', $executedVersion['start_time']);
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -105,22 +105,84 @@ class Manager
         $migrations = array();
         $hasDownMigration = false;
         $hasMissingMigration = false;
-        if (count($this->getMigrations())) {
+        $migrations = $this->getMigrations();
+        if (count($migrations)) {
             // TODO - rewrite using Symfony Table Helper as we already have this library
             // included and it will fix formatting issues (e.g drawing the lines)
             $output->writeln('');
-            $output->writeln(' Status  Migration ID    Started              Finished             Migration Name ');
+
+            switch ($this->getConfig()->getVersionOrder()) {
+                case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
+                    $migrationIdAndStartedHeader = "<info>[Migration ID]</info>  Started            ";
+                    break;
+                case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
+                    $migrationIdAndStartedHeader = "Migration ID    <info>[Started          ]</info>";
+                    break;
+            }
+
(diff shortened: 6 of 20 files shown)


daniel-gomes-sociomantic:
@rquadling this is simply the rebase from `0.5.x-dev` to `0.6.x-dev`, where I had to adapt to the new breakpoint feature.

I still have to act on your https://github.com/robmorgan/phinx/pull/745#issuecomment-232322038

daniel-gomes-sociomantic:
regarding your comment:

> A suggestion for phinx status, could a highlight be added to the column that is used for the ordering.

ok, but I'm not sure how to highlight the column

> Maybe the rollback_order in the config could be version_order, allowing for the ordering to be used logically for phinx status also.

makes sense to me, shall I do it?

> For the version ordering, if the key to the version log retrieval was the ID or the execution date (in CCYYMMDDHHIISS format), then the existing rollback methods COULD be used AND only a --target would be needed. No need to differentiate between the mechanisms for ordering.

hmm so you're saying like, get the versions properly ordered by creation or execution time from `getVersionLog` and then the rollback itself would be the same in either case? Makes sense, I dunno why I didn't do it like that initially :-) I'll have a look at it, as currently the methods to rollback by creation (which already existed) and execution time (which I added) are quite different.

> Once the ordering is set, I've made my intent clear. --target should operate on that setting appropriately without the need for addition flags.

So no `-s/--start_time` command-line option, only the config option?

rquadling:
That sounds about perfect. No `-s/--start-time`.

For highlighting, https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Migration/Manager.php#L112 contains the column headings.

A simple variant could be to wrap the name with `<info>[...]</info>` or to use `=====` as underline for the column.

Probably though the `[...]` would be best.

daniel-gomes-sociomantic:
on it!

quick question: for the "version order" references, I believe it's much cleaner to have the valid values (strings `start-time` and `create-time`) defined somewhere as constant values so we can use them cleanly throughout the code, eg. doing:

```php
// bad
if ('start-time' === $versionOrder)
```

Vs

```php
// good
if (VERSION_ORDER_START_TIME === $versionOrder)
```

My question is, where shall I store these values? They are gonna be needed (I think) in the `Rollback` command and `Manager` classes.

daniel-gomes-sociomantic:
Another question: for passing the version order from the `Rollback` command class (which gets it from the configuration file and checks its valid) to the `Manager` class (which obviously needs it), shall I pass it simply as a `rollback` method argument, or is there a better way? Just thinking that that method already has 2 optional arguments (`$version` and `$force`), if we keep adding optional parameters it might make it hard to call it with only some options (eg. `->rollback('env', null, null, $versionOrder)`.

Maybe I can simply read it from the Config class? The only other thing I'd need to do is set the key to its default value (creation-time) in the `Rollback` command class if it is not set.

rquadling:
I would put the constants in Config, as they are configuration constants. With an appropriate getter(). Which will also include the validation and throw an exception if the value is incorrect/unknown. If not set then supply the default which would preserve the current status quo of creation date order.

So, then, in both status and rollback, something like `$versionOrder = $this->getConfig()->getVersionOrder();` maybe.

daniel-gomes-sociomantic:
perfect, looks good.

What about the rollback command's `-d/--date` parameter, does it still make sense to have it? From your comment:

> Especially if the logic was merged, allowing something like --target 201604 . rollback to the beginning of April 2016. Either by version ID or by execution date, depending upon the setting of version_order (currently rollback_order within this branch).

It looks like we don't need a `--date/-d` argument at all?

daniel-gomes-sociomantic:
there is a problem with:

> For the version ordering, if the key to the version log retrieval was the ID or the execution date (in CCYYMMDDHHIISS format), then the existing rollback methods COULD be used AND only a --target would be needed. No need to differentiate between the mechanisms for ordering.

which is that whereas it is impossible to have two versions with the same create time (right?), it might happen that two versions are executed in bulk at the same time and end up with the same start time, meaning we can't really index the versions/migrations by start time without ensuring we won't ignore entries with the same start time (ie. array key)...

daniel-gomes-sociomantic:
> it is impossible to have two versions with the same create time (right?),

I double-checked since I didn't remember seeing any checks for this in the code, but the PRIMARY key in the `phinxlog` table ensures that version names (ie. their creation times) are indeed unique

daniel-gomes-sociomantic:
> we can't really index the versions/migrations by start time without ensuring we won't ignore entries with the same start time (ie. array key)...

I could index the migrations (returned in `Manager.getMigrations()`) and the versions (returned in `PdoAdapter.getVersionLog()`) by start time if the array values are arrays of versions instead of single versions. 

This would allow for having several versions with the same start time, but it wouldn't be able to ensure any definitive version order _for the same start time_, which could lead to unexpected behaviour. Maybe I can sort the versions by start time _and_ creation time (aka ID)?

daniel-gomes-sociomantic:
just updated the initial PR description )copied from the old PR) with the new stuff from this PR

rquadling:
If they are ordered by `$versionOrderingColumn, 'version'`, then that should keep the order just fine.

With regard to the output for the status, try running your code with `--ansi`. Does this output anything visually? If not, I think we need an additional marker for the ordering.

rquadling:
Is there no way to merge `rollbackToDateTime()` with `rollbackToDateTimeByStartTime()`  and to merge `rollback()` with `rollbackByStartTime()`.

It just grates having 4 methods that all do the same work. In essence. I may need to take a closer look at the workings of these.

rquadling:
And I'm happy to be helping here. A lot of the features people are working on I will be actively using anyway.

daniel-gomes-sociomantic:
> If they are ordered by $versionOrderingColumn, 'version', then that should keep the order just fine.

on it.

> With regard to the output for the status, try running your code with --ansi. Does this output anything visually? If not, I think we need an additional marker for the ordering.

the output is the same with `--ansi` or without anything, the "order-by" column is shown in green while the remaining columns stay white.

> Is there no way to merge rollbackToDateTime() with rollbackToDateTimeByStartTime() and to merge rollback() with rollbackByStartTime().
> 
> It just grates having 4 methods that all do the same work. In essence. I may need to take a closer look at the workings of these.

I will try to merge them now - the only problem I had was the ordering, now that you replied I'm good to try it.

How about the `-d/--date` option? Stays or goes?

rquadling:
`--no-ansi` (DOH!)

As for `-d/--date` yeah. I can now see they are both needed. But I'd expand date to work as:

2016 -> 20160101000000 / Midnight Jan 1st 2016
201603 -> 20160301000000 // Midnight March 1st 2016
20160503 -> 20160503000000 // Midnight 3rd May 2016
2016070810 -> 20160708100000 // 10:00:00 on 8th July 2016
201609121345 -> 20160912134500 // 13:45:00 on 12th September 2016
20160912134555 -> 20160912134555 // 13:45:55 on 12th September 2016

Basically, allowing me to use a point in time between 2 runs of the migration.

It would be nice to have a --dry-run option that would show the status page with the current up/down and the new up/down, taking into account any breakpoints. https://github.com/robmorgan/phinx/issues/906

daniel-gomes-sociomantic:
> --no-ansi (DOH!)

Indeed, with `--no-ansi` the status is displayed all in white and you can't tell the order-by column (sorry, only now did I understand where you were going with that). How would we signalize the order-by column in that case? Maybe an asterisk `*` next to the column name? Don't really know what the standard is...

As for the `-d/--date` option, I think I get the need for it. 

Let me just summarize how I think the rollback mechanism would work, to make sure we are on the same page:

Premise: Depending on the configuration's version order, the existing versions (ie. the previously executed migrations) will be indexed and sorted by creation time (aka name) or start time (aka execution time).

Behaviour with a `-t/--target` option:
* Value 0: **all** migrations are rolled back.
* Version name of an existing (and previously executed) migration (eg. `20111018185412`): all the migrations **after** that one (in the sorted list from the premise) will be rolled back.
* Version name of a non-existing migration: an exception is thrown.

Behaviour with a `-d/--date` option:
* All the migrations **after** that date (in the sorted list from the premise) will be rolled back.

Is that it?

(note that this description is ignoring breakpoints, the implementation of which is quite simple: if at any point a migration is about to be rolled back and it has a breakpoint set for it, the mechanism stops right there without rolling back that migration and prints the `Breakpoint reached` error message)

rquadling:
Maybe changing the `-------` under the label to `=======` or add `[....]` around the label. See which looks good. Probably `<info>[....]</info>` would be my preference.



As for `-d|-t` ...

In both cases, date and target operate on the sorted results.

Depending upon the ordering, the status page would look like ...
```
By version

Status  [Migration ID]  Started
-------------------------------------------
    up  20160105000000  2016-01-05 00:00:00 - 1st deployment
    up  20160110000000  2016-01-12 00:00:00 - 3rd deployment
    up  20160115000000  2016-01-10 00:00:00 - 2nd deployment
```
or
```
By execution

Status  Migration ID    [Started          ]
-------------------------------------------
    up  20160105000000  2016-01-05 00:00:00 - 1st deployment
    up  20160115000000  2016-01-10 00:00:00 - 2nd deployment
    up  20160110000000  2016-01-12 00:00:00 - 3rd deployment
```



Now lets got through the some "targets" for rollback.

```
Target         Expectation
-------------- -----------------------------------------------
0              All migrations
2016           All migrations from 2016-01-01 00:00:00 onwards
201602         All migrations from 2016-02-01 00:00:00 onwards
20160304       All migrations from 2016-03-04 00:00:00 onwards
2016040506     All migrations from 2016-04-05 06:00:00 onwards
201605060708   All migrations from 2016-05-06 07:08:00 onwards
20160607080910 All migrations from 2016-06-07 08:09:10 onwards
```

For 'By version' supplying a target would rollback the following migrations.
```
Target   Rollbacks performed in the following order
-------- -------------
20160104 2nd, 3rd, 1st
20160105 2nd, 3rd
20160106 2nd, 3rd
20160107 2nd, 3rd
20160108 2nd, 3rd
20160109 2nd, 3rd
20160110 2nd
20160111 2nd
20160112 2nd
20160113 2nd
20160114 2nd
20160115 none
20160116 none
```

For 'By execution' supplying a target would rollback the following migrations.
```
Target   Rollbacks performed in the following order
-------- -------------
20160104 3rd, 2nd, 1st
20160105 3rd, 2nd
20160106 3rd, 2nd
20160107 3rd, 2nd
20160108 3rd, 2nd
20160109 3rd, 2nd
20160110 3rd
20160111 3rd
20160112 none
20160113 none
20160114 none
20160115 none
20160116 none
```

I see "target" and "date" being merged into a "time point", which has automatic expansion to a full datetime depending upon what parts are not present.

And that datetime is then used against the ordering. As the ordering is set at a config level, I don't see a need to differentiate once the migrations to rollback have been determined.

All migrations after the timepoint are rolled back, depending upon the sort order.

daniel-gomes-sociomantic:
> Maybe changing the ------- under the label to ======= or add [....] around the label. See which looks good. Probably <info>[....]</info> would be my preference.

easy peasy.

> I see "target" and "date" being merged into a "time point", which has automatic expansion to a full datetime depending upon what parts are not present.

I'm with you, but doesn't that mean that `-d/--date` and `-t/--target` are the same then? If both of them are translated into a "time point" that is then matched against the executed migrations (regardless of how they were ordered as per the current configuration) to decide which ones get rolled back, then I can't see any difference between the 2 options. I thought that the difference was that the `-v/--version` was "stricter" in the sense that you had to pass an exact version name (regardless of the configuration version order), but reading what you wrote, it doesn't look like it... I mean, in your example you yourself simply wrote:

> Now lets got through the some "targets" for rollback.

without distinguishing between a target version via `v/--version` or a target date via `-d/--date`.

rquadling:
It is just painful having -t 20160105121314 AND a -d "2016-01-05 12:13:14"

If someone needs a "strict" point, they will enter one. But it is also about enforcing a strict behaviour.

Fair enough.

So, because of the 'strict' nature of MUST having a matched ...

`-d` gets expanded to a full date time (the timepoint).
`-t` is already expanded and so is already a timepoint.

If the timepoint is from `-d` it doesn't have to exist as a date (version or execution, depending upon ordering).

If the timepoint is from `-t` it DOES have to match at least 1 datetime (version or execution, depending 
upon ordering).

How does that sound.

And of course, that may be exactly what we do already! Ha ha! (ahem - I'm off home - sorry for wasting everyone's time!!!).

rquadling:
I do feel that -d and -t are essentially the same. The one issue is the 'strictness' of -t. The requested target MUST exist.

So.

I think if we can support all the features with minimal code (i.e. not 4 methods that nearly all do the same thing, then I think that will be best. Keeping both -t and -d will be easier for end users as they already have that.

I think the code should do the following steps.

Get the migrations in the appropriate order. Keyed by version ID.
If -t, verify migration exists. If migration does exist, extract the appropriate timepoint (depending upon version ordering). If the migration does not exist, bail out.

If -d, expand to a full time point.

Rollback migrations in order until migration timepoint (depending upon the version ordering).

All handled by 1 method.

daniel-gomes-sociomantic:
> Get the migrations in the appropriate order. Keyed by version ID.
> If -t, verify migration exists. If migration does exist, extract the appropriate timepoint (depending upon version ordering). If the migration does not exist, bail out.
> 
> If -d, expand to a full time point.
> 
> Rollback migrations in order until migration timepoint (depending upon the version ordering).
> 
> All handled by 1 method.

@rquadling done :-) let me know what you think

daniel-gomes-sociomantic:
I see some failed tests in AppVeyor, although in Travis CI (and locally) everything passes. What spefifically is AppVeyor testing, how dos it differ from Travis and how can I run that environment locally?

rquadling:
AppVeyor runs Windows + MSSQL. I can re-run the tests locally and give some feedback. If you can grant me collaborator access to your phinx repo, I can push the fixes I find.

rquadling:
You've changed the tests and have embedded unix line endings in the regex which will fail on Windows.

```git
 -         $this->assertRegExp('/\s*up  20120103083300  2012-01-11 23:53:36  2012-01-11 23:53:37  *\*\* MISSING \*\*\n'.
 -             '\s*up  20120815145812  2012-01-16 18:35:40  2012-01-16 18:35:41  Example   *\*\* MISSING \*\*\n' . 
 -             '\s*down  20120111235330                                            TestMigration\n' .
 +         $this->assertRegExp('/\s*up  20120103083300  2012-01-11 23:53:36  2012-01-11 23:53:37  *\*\* MISSING \*\*'.PHP_EOL.
 +             '\s*up  20120815145812  2012-01-16 18:35:40  2012-01-16 18:35:41  Example   *\*\* MISSING \*\*' .PHP_EOL.
 +             '\s*down  20120111235330                                            TestMigration' .PHP_EOL.
```

The above is an example patch.

daniel-gomes-sociomantic:
indeed I have! will get rid of that

daniel-gomes-sociomantic:
:boom:

daniel-gomes-sociomantic:
I'll add the new features to the docs now

daniel-gomes-sociomantic:
Done. Didn't build the docs to see how they looked, hope it's all fine

docs/commands.rst
```diff
@@ -127,13 +127,28 @@ Specifying 0 as the target version will revert all migrations.
...
 If a breakpoint is set, blocking further rollbacks, you can override the
 breakpoint using the ``--force`` parameter or ``-f`` for short.
 
 .. code-block:: bash
 
         $ phinx rollback -e development -t 0 -f
 
+.. note::
+
+        When rolling back, Phinx orders the executed seed migrations using
```
rquadling: The use of the word "seed" here is probably confusing as migrations and seeds are separate entities.

Seeds are done in a different order and not affected by `version_order`.

daniel-gomes-sociomantic: Ups, bad copy/paste :-)

docs/configuration.rst
```diff
@@ -280,3 +280,12 @@ The aliased classes will still be required to implement the ``Phinx\Migration\Cr
...
         permission: \Namespace\Migrations\PermissionMigrationTemplateGenerator
         view: \Namespace\Migrations\ViewMigrationTemplateGenerator
+
+Version Order
+------
+
+When rolling back or printing the status of migrations, Phinx orders the executed migrations according to the 
+``version_order`` option, which can have the following values:
+
+* ``creation-time`` (the default): migrations are ordered by their creation time, which is also part of their names.
```
rquadling: Suggest talking about migration filename rather than name.

daniel-gomes-sociomantic: good point, I thought the same but then didn't fix it

docs/configuration.rst
```diff
@@ -280,3 +280,12 @@ The aliased classes will still be required to implement the ``Phinx\Migration\Cr
...
         permission: \Namespace\Migrations\PermissionMigrationTemplateGenerator
         view: \Namespace\Migrations\ViewMigrationTemplateGenerator
+
+Version Order
+------
+
+When rolling back or printing the status of migrations, Phinx orders the executed migrations according to the 
+``version_order`` option, which can have the following values:
+
+* ``creation-time`` (the default): migrations are ordered by their creation time, which is also part of their filenames.
```
rquadling: Migrations only have 1 filename. :-D

rquadling: Migrations only have a single filename.

src/Phinx/Config/Config.php
```diff
@@ -294,6 +304,52 @@ public function getTemplateClass()
...
     /**
+     * Get the version order.
+     *
+     * @return string
+     */
+    public function getVersionOrder()
+    {
+        if (!isset($this->values['version_order'])) {
+            // if the configuration value is missing, return the default version order (creation time)
+            return self::VERSION_ORDER_CREATION_TIME;
```
rquadling: It would be nicer for the setter to have the guard (validate appropriate parameter, use default if not supplied and set/save it accordingly) and the getter be dumb, considering this is only exposing a single property in the config.

The setter would need to be called as part of the config instantiation process if it not already (commenting inline without seeing all the other code).

src/Phinx/Config/Config.php
```diff
@@ -294,6 +304,52 @@ public function getTemplateClass()
...
+
+        return $versionOrder == self::VERSION_ORDER_CREATION_TIME;
+    }
+
+    /**
+     * Set the version order.
+     */
+    public function setVersionOrder($versionOrder)
+    {
+        $this->values['version_order'] = $versionOrder;
```
rquadling: Setter needs to guard against invalid values.

src/Phinx/Console/Command/Rollback.php
```diff
@@ -61,6 +62,10 @@ protected function configure()
 If you have a breakpoint set, then you can rollback to target 0 and the rollbacks will stop at the breakpoint.
 <info>phinx rollback -e development -t 0 </info>
 
+The <info>version_order</info> configuration option is used to determine the order of the migrations when rolling back.
+This can be used to allow rolling back the last executed migration instead of the last created one, or combined with the
```
rquadling: `rolling back ___of___ the last executed migration`

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -406,8 +406,18 @@ public function getVersions()
     public function getVersionLog()
     {
         $result = array();
-        $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY version ASC', $this->getSchemaTableName()));
-        foreach ($rows as $version) {
+
+        switch ($this->options['version_order']) {
+            case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
```
rquadling: Internally, I think it would be nicer to see `VERSION_ORDER_VERSION = 'version';` / `VERSION_ORDER_START_TIME = 'start_time` as these are the names of the columns in the phinxlog table.

Externally (in the config), the end user setup may be simpler to have `version_order` be one of `'execution'` or `'creation'`.

src/Phinx/Migration/Manager.php
```diff
@@ -368,61 +391,89 @@ public function executeSeed($name, SeedInterface $seed)
...
      *
      * @param string $environment Environment
-     * @param int $version
+     * @param int $target
      * @param bool $force
+     * @param bool $targetMustMatchVersionName
      * @return void
      */
-    public function rollback($environment, $version = null, $force = false)
+    public function rollback($environment, $target = null, $force = false, $targetMustMatchVersionName = true)
```
rquadling: `$targetMustMatchVersionName` - I would prefer `$targetMustMatchVersion`. The '`Name`' part doesn't fit with anything in the DB or other code.

src/Phinx/Migration/Manager.php
```diff
@@ -368,61 +391,89 @@ public function executeSeed($name, SeedInterface $seed)
...
     {
-        $migrations = $this->getMigrations();
-        $versionLog = $this->getEnvironment($environment)->getVersionLog();
-        $versions = array_keys($versionLog);
+        // note that the migrations are indexed by name (aka creation time) in ascending order
+        $projectMigrations = $this->getMigrations();
```
rquadling: Don't rename existing variables unless the intent has changed.

`$migrations = $this->getMigrations();` is fine.

src/Phinx/Migration/Manager.php
```diff
@@ -368,61 +391,89 @@ public function executeSeed($name, SeedInterface $seed)
...
 
-        ksort($migrations);
-        sort($versions);
+        // note that the version log are also indexed by name with the proper ascending order according to the version order
+        $executedMigrationVersions = $this->getEnvironment($environment)->getVersionLog();
```
rquadling: Inconsistent naming `$projectMigrations` vs `$executedMigrationVersions` ... very confusing when `$migrations` is right.

src/Phinx/Migration/Manager.php
```diff
@@ -555,7 +609,7 @@ public function setMigrations(array $migrations)
      * Gets an array of the database migrations.
      *
      * @throws \InvalidArgumentException
-     * @return AbstractMigration[]
+     * @return AbstractMigration[], indexed by migration name (aka creation time) and sorted in ascending order
```
rquadling: `* @return AbstractMigration[] Indexed by migration name (aka creation time) and sorted in ascending order`

The return type and description are separate entities.

src/Phinx/Migration/Manager/Environment.php
```diff
@@ -231,7 +231,8 @@ public function getVersions()
     }
 
     /**
-     * Get all migration log entries, indexed by version number.
+     * Get all migration log entries, indexed by version name and sorted ascendingly by the configuration's version 
+     * order option
```
rquadling: The configuration option is called `version_order`. Keep this name consistent.

rquadling:
A few more points about consistency. An important one about changing the name of pre-existing variables and introducing new terms in the names ($projectMigrations when $migrations seemed perfectly reasonable). If separate collections of migrations then maybe things like `$executedMigrations`, `$outstandingMigrations`, `$missingMigrations` ... all fine if used throughout the codebase.

src/Phinx/Config/Config.php
daniel-gomes-sociomantic: initially I only had the getter, which is why the logic ended up here, will move it to the setter

src/Phinx/Console/Command/Rollback.php
daniel-gomes-sociomantic: wouldn't that need a `the` before, as in: `to allow ___the___ rolling back ___of___ the last executed migration`?

src/Phinx/Db/Adapter/PdoAdapter.php
daniel-gomes-sociomantic: Won't that be confusing, having 2 names for the same thing? Since in the end we don't actually use the `version` and `start_time` column names directly (as you can see down there, it's more of a `version -> version ASC` and `start_time -> start_time ASC, version ASC` mapping), it seems to me like adding confusion for no real gain?

I do think it looks indeed nicer if I rename the valid config option values from `creation-time` and `start-time` to `creation` and `execution` like you said, makes it clearer for the end-user

daniel-gomes-sociomantic:
To be honest, I was getting **very** confused with names like `$migrations` and `$version`, as in, which one refers to the migration classes and which one refers to the rows in `phinxlog`. 

But maybe it's just me, so for the sake of consistency, I'll rename the variables.

In the meantime, I think I found a bug in the Status command, so hold off on the merge ;-)

src/Phinx/Migration/Manager.php
daniel-gomes-sociomantic: can do. The problem was that `version` sometimes refered to the creation time and sometimes to the array with all the version data (creation time, start time etc). But again for consistency, renaming

daniel-gomes-sociomantic: what I mean was, I used `*versionName` for several variable names, which should also be fixed, but it can lead to confusion with the fact that `version` is used for the version array-data. In this case, I guess I'll rename the variables to `*versionCreationTime`

src/Phinx/Migration/Manager.php
daniel-gomes-sociomantic: so where should the `Indexed by migration name (aka creation time) and sorted in ascending order` part go to? I thought that for clarity it's important that the structure of that's method's returned array is defined somewhere

src/Phinx/Migration/Manager/Environment.php
daniel-gomes-sociomantic: right, this was the "english name" of the configuration option, wasn't meant to be taken literally, but to avoid confusion I'll rename it to the actual configuration option

src/Phinx/Migration/Manager.php
daniel-gomes-sociomantic: right, into the description, of course ;-)

daniel-gomes-sociomantic:
bugs fixed, ready for more scrutiny!

src/Phinx/Console/Command/Rollback.php
```diff
@@ -100,17 +107,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $dateStrlenToAppend = array(
+            14 => '',
+            12 => '00',
+            10 => '0000',
+            8 => '000000',
+            6 => '01000000',
+            4 => '0101000000',
+        );
+
+        $format = null;
```
rquadling: Unused?

src/Phinx/Console/Command/Rollback.php
```diff
@@ -100,17 +107,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if ($target === null) {
+            throw new \InvalidArgumentException(sprintf('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'parts, but the missing part(s) must all be in the end of the date string', $date));
+        }
+
+        $dateTime = \DateTime::createFromFormat('YmdHis', $target);
+
+        if ($dateTime === false) {
+            throw new \InvalidArgumentException(sprintf('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'parts, but the missing part(s) must all be in the end of the date string', $date));
```
rquadling: What is the sprintf doing? Doesn't use $date. Ditto for line 164.

src/Phinx/Console/Command/Rollback.php
```diff
@@ -100,17 +107,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            12 => '00',
+            10 => '0000',
+            8 => '000000',
+            6 => '01000000',
+            4 => '0101000000',
+        );
+
+        $format = null;
+
+        foreach($dateStrlenToAppend as $dateStrLen => $append) {
```
rquadling:
```php
if (isset($dateStrlenToAppend[strlen($date)]){
  $target = $date . $dateStrlenToAppend[strlen($date)];
}
```
maybe?

src/Phinx/Console/Command/Rollback.php
```diff
@@ -100,17 +107,71 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
+        $dateTime = \DateTime::createFromFormat('YmdHis', $target);
+
+        if ($dateTime === false) {
+            throw new \InvalidArgumentException(sprintf('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'parts, but the missing part(s) must all be in the end of the date string', $date));
+        }
+
+        return $dateTime->format('YmdHis');
```
rquadling: Line 168 makes a string, can't call method on a string. As all unit tests are passing, it would seem this code is not called

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -406,8 +406,18 @@ public function getVersions()
...
+                $orderBy = 'version ASC';
+                break;
+            case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
+                $orderBy = 'start_time ASC, version ASC';
+                break;
+        }
```
rquadling: If the config has no guard (which would be consistent with the majority of the settings), then a `default:` clause is probably worth having here.

src/Phinx/Db/Adapter/PdoAdapter.php
rquadling: Move the default option to the bottom of the list and add a `default:` clause, so that if nothing is setup right, the default option will operate appropriately.

src/Phinx/Migration/Manager.php
```diff
@@ -105,22 +105,84 @@ public function printStatus($environment, $format = null)
...
-        if (count($this->getMigrations())) {
+        $migrations = $this->getMigrations();
+        if (count($migrations)) {
             // TODO - rewrite using Symfony Table Helper as we already have this library
             // included and it will fix formatting issues (e.g drawing the lines)
             $output->writeln('');
-            $output->writeln(' Status  Migration ID    Started              Finished             Migration Name ');
+
+            switch ($this->getConfig()->getVersionOrder()) {
+                case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
```
rquadling: Same about the default here as in the `getVersionLog()` comment above.

rquadling:
Rebase and squash as we are getting a lot of commits here. Everything is going in the right direction.

rquadling:
Rebase is important as you are missing tests (and maybe code) related to breakpoints.

src/Phinx/Console/Command/Rollback.php
daniel-gomes-sociomantic: indeed, much better

src/Phinx/Console/Command/Rollback.php
daniel-gomes-sociomantic: it used to ;-) removing `sprintf`

src/Phinx/Config/Config.php
```diff
@@ -294,6 +306,58 @@ public function getTemplateClass()
...
+     */
+    public function setVersionOrder($versionOrder = null)
+    {
+        if (!isset($versionOrder)) {
+            $versionOrder = isset ( $this->values['version_order'] ) ? $this->values['version_order'] : null;
+        }
+
+        if (!isset($versionOrder)) {
+            // if the configuration value is missing we use the default version order (creation time)
+            $this->values['version_order'] = self::VERSION_ORDER_CREATION_TIME;
```
rquadling: Oh dear. I led you the wrong way. I forgot that the config class is, essentially, read only, no setters() needed. All guards are in the getters().

And looking at the rest of config, it doesn't really guard against much!

I think the getter can be as simple as:
```php
return $this->values['version_order'] ?: self::VERSION_ORDER_CREATION_TIME;
```

src/Phinx/Console/Command/Rollback.php
rquadling: Add all the the's and ofs. :-D

src/Phinx/Config/Config.php
daniel-gomes-sociomantic: so I should remove the setter and replace the getter with that? Where should the guard go to? Having an invalid configuration version order (because of a typo, for instance) might lead to unexpected behaviour

rquadling: Not necessarily. If you add the `default:` clauses, the the system will use that for an incorrect value. I think there are a limited number of places where the value is used.

docs/commands.rst
```diff
@@ -127,13 +127,28 @@ Specifying 0 as the target version will revert all migrations.
 
         $ phinx rollback -e development -t 0
 
+To rollback all migrations to a specific date then use the ``--date``
+parameter or ``-d`` for short.
+
+.. code-block:: bash
+
+        $ phinx rollback -e development -d 20120103
```
rquadling: Extend this to show that you can (for example) rollback to 2016 by `-d 2016` or 3AM today by `-d 2016-07-19 03`.

Essentially, `Y[m[d[H[i[s]]]]]`.

I wonder about supporting `-d 3pm`, `-d yesterday 2pm`, `-d 2 hours ago`, etc.

And should format of the -d parameter be the same as -t

daniel-gomes-sociomantic:
rebased and squashed

docs/commands.rst
daniel-gomes-sociomantic: > Extend this to show that you can (for example) rollback to 2016 by -d 2016 or 3AM today by -d 2016-07-19 03.
> 
> Essentially, Y[m[d[H[i[s]]]]].

will do.



> I wonder about supporting -d 3pm, -d yesterday 2pm, -d 2 hours ago, etc.

wouldn't be too tricky, maybe as a future improvement?

> And should format of the -d parameter be the same as -t

what do you mean?

src/Phinx/Config/Config.php
daniel-gomes-sociomantic: so u mean, wherever in, say, the `Manager` class where I do a `getVersionOrder`, I do a `switch` clause with cases for creation and start time and default to raising an exception for an unknown order?

src/Phinx/Console/Command/Rollback.php
```diff
@@ -100,17 +107,66 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+     * Get Target from Date
+     *
+     * @param string $date The date to convert to a target.
+     * @return The target
+     */
+    public function getTargetFromDate($date)
+    {
+        if (!preg_match('/^\d{4,14}$/', $date)) {
+            throw new \InvalidArgumentException('Invalid date. Format is YYYYmmddHHiiss (with no mandatory '.
+                'part, but the missing part(s) must all be in the end of the date string)');
```
daniel-gomes-sociomantic: @rquadling maybe I make this message more nice and concise like:

    Invalid date. Format is Y[m[d[H[i[s]]]]].

?

src/Phinx/Db/Adapter/PdoAdapter.php
daniel-gomes-sociomantic: > Move the default option to the bottom of the list and add a default: clause, so that if nothing is setup right, the default option will operate appropriately.

to be discussed in https://github.com/robmorgan/phinx/pull/901#discussion_r71486423

as for the actual option string names, should I rename `creation-time` and `start-time` to `creation` and `execution` (respectively) or not?

src/Phinx/Config/Config.php
daniel-gomes-sociomantic: or it actually sounds like you're suggesting something like:

```php
switch($this->getConfig()->getVersionOrder()) {
    case \Phinx\Config\Config::VERSION_ORDER_START_TIME:
        // handle start time version order
        break;
    case \Phinx\Config\Config::VERSION_ORDER_CREATION_TIME:
    default:
        // handle creation time version order
}
```

The problem I see here is that an invalid version order (eg. a typo like `starttime`) would be invisibly handled as a creation-time, which could lead to unexpected behaviour (since phinx would rollback the "invalid" commit, at least in the eyes of the user). I believe we should guard against this, if not on the Config getter/setter, then at least in the switch's `default` clause, by raising an exception. If we do so, the only thing left to do is to return the default version order (creation-time) in the getter if none was set in the configuration file.

rquadling: I would use default to be the default. So the current behaviour is in creation time order, so that's the default behaviour.

docs/commands.rst
rquadling: > future improvement?
Yes.

Should -t be ONLY for specific creation timestamps?

src/Phinx/Console/Command/Rollback.php
rquadling: The format is an acceptable way of describing a parameter. It is not uncommon to see nested optional parts. Unusual to see many as this, probably, not not unusual enough to need anything more complex.

I'd recommend `Invalid value. Format is CCYY[MM[DD[HH[II[SS]]]]]`. I'd also like to have the --help page show a few examples as well as mention it is not JUST date.

-t is target (or timestamp)
-d is date (or datetime)

All a bit samey.

src/Phinx/Config/Config.php
daniel-gomes-sociomantic: > I would use default to be the default. So the current behaviour is in creation time order, so that's the default behaviour.

even if the user writes something else in the config file? that seems to go beyond the scope of "default"

docs/commands.rst
daniel-gomes-sociomantic: OK so if I understood correctly, it's fine the way it is right now and then later we can add support for `-d/--date` values like `2016-07-20`, `3 hours ago` etc. That it?

src/Phinx/Console/Command/Rollback.php
daniel-gomes-sociomantic: all good, but, why `CCYY` and not `YYYY`? What does the `C` stand for? Century? Afaik, date formats simply refers to `Y` for years

rquadling: CCYY or YYYY, but never YY or Y.

daniel-gomes-sociomantic:
so, I think currently we only have these open points:

* Should I rename the `version_order` values to `creation` and `execution` or keep them as `creation-time` and `start-time` (respectively)?
* The Config getter/setter where-should-the-guard-go-to discussion.

rquadling:
I prefer the shorter form.

The guard for the current config options is based upon the consumer, not in the config class.

By having it in the config class, potentially we are restricting subclassing. Maybe.


For the guard, the default value is assigned during config and then using a switch, the default case could either be the same as the default value or an exception could be thrown.

Thinking ...


Exception. Basically, if someone has gone to the trouble of editing the config and done it wrong, BOOM!

If no config changes, then everything operates exactly as it is at the moment.

daniel-gomes-sociomantic:
done! :-)

anything else missing??

rquadling:
That's all looking very good!!! I'll be doing a thorough review tomorrow (and testing too. Hopefully unit tests have increased coverage.



One thing I've found in my own code, especially where there are many options all working together,  generating the test data for unit testing can be as troublesome as the code itself.

So (doing this blind without looking at the tests), the following options all play a part in rollback.

1. Version order setting - null/undefined, 'creation', 'execution', invalid.
2. Migrations order - same order, reversed order, mixed.
3. Breakpoint settings - setting is after target, setting is on target, setting is after before target.
4. Rollback target - after all migrations, is an exact match for a migration, between 2 migrations, before all migrations.
5. Number of migrations - 0, 1, 2, 3+

So, just taking into account those options, potentially, 432 combinations. Some won't be valid (0 migrations for example), but the code still needs to respond appropriately.

Having a rollbackProvider that can present all these tests with their expected results or exceptions, and, ideally, a message indicative of the test data (Version order undefined, migrations in order, no breakpoint settings, target after all migrations with 0 migrations failed).

It is the large combinations of options that cover all bases that reveal those broken edge cases.

If need be, write a script (and commit it) that generates the tests. That way, if the fundamentals change, we can change the generator and generate the tests again.

rquadling:

daniel-gomes-sociomantic:
replaced by https://github.com/robmorgan/phinx/pull/926

PR #905, opened 2016-07-15
Constructor declaration removed from ConfigInterface

f3ath:
Having constructor declared in \Phinx\Config\ConfigInterface makes it unreasonably hard to make non file based configurations. I have a Silex project with its own configuration infrastructure and I want to use Phinx as the migration tool. My plan is to inherit from `\Symfony\Component\Console\Application` roughly like this:

```php
class MyPhinxApplication extends \Phinx\Console\PhinxApplication
{
    public function __construct(MySilexApplication $app)
    {
        parent::__construct();
        $config = new MyConfigAdapter($app);
        foreach ($this->all() as $command) {
            if ($command instanceof Phinx\Console\Command\AbstractCommand) {
                $command->setConfig($config);
            }
        }
    }

}

class MyConfigAdapter implements \Phinx\Config\ConfigInterface
{
    public function __construct(MySilexApplication $app)
    {
        // here i can extract whatever i need from my Silex app
    }
    ...
}
```

The constructor declaration effectively prevents me from doing this, instead i have to fake the `$configArray` and `$configFilePath`, and inject my Silex app through a setter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Version History
 
+**0.7.x-dev** (***)
+
+* New features
+  * [#885](https://github.com/robmorgan/phinx/pull/885) Add `collation` and `encoding` options for MySQL text and char columns
+
 **0.6.x-dev** (***)
 
 * Documentation updates
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -711,6 +711,15 @@ Option   Description
 signed   enable or disable the ``unsigned`` option *(only applies to MySQL)*
 ======== ===========
 
+For ``string`` and ``text`` columns:
+
+========= ===========
+Option    Description
+========= ===========
+collation set collation that differs from table defaults *(only applies to MySQL)*
+encoding  set character set that differs from table defaults *(only applies to MySQL)*
+========= ===========
+
 For foreign key definitions:
 
 ====== ===========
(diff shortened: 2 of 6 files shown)


rquadling:
Hmmm. I see.  Loosening the constructor allows for greater flexibility. The constructor doesn't really enforce any behaviour beyond that of the other methods needed by Phinx, which your `MyConfigAdapter` would need to do anyway.

Hmmm.

The `\Phinx\Config\Config::replaceTokens()` method is protected so not governed by the interface contract. The `\Phinx\Config\Config` implementation for the constructor is a simple injector with a guard for the tokens.

Hmmm.

Still no badness in removing the constructor from the interface.

Let me think on if for a bit and lets see if there is any other input.

echo511:
+1 Removing constructor from interface is a must. Interface should define public methods for other classes to use and since calling $class->__construct() is quite a stupidity it should be removed!

Guarding tokens should be solely the Phinx\Config\Config class responsibility. If someone wishes to implement their own Config class then it's theirs responsibility.

As best approach I see extending current config class which is not possible right now. Example below:

```php
class Config extends \Phinx\Config\Config
{
    public function __construct(Configurator $configurator)
    {
        parent::__construct($configurator->getPhinxConfigArray());
    }
}
```

rquadling:
Any arguments for keeping the constructor in the interface?

f3ath:
Should this be re-committed against master?

rquadling:
No. Master gets updated upon release. Until that happens, it will be in the dev branch.

As this is an interface change, SemVer requires this to go into a new tag for the release, so that will need to go into a new 0.7.x-dev branch.

Hence the 0.7.0 Milestone.

rquadling:

PR #908, opened 2016-07-19
When setting options as a non-associative array, Column fails with "Call to undefined method Phinx\Db\Table\Column::set0()"

f3ath:
This check makes sure the developer gets an appropriate message.

### DIFF
--- a/src/Phinx/Db/Table/ForeignKey.php
+++ b/src/Phinx/Db/Table/ForeignKey.php
@@ -216,8 +216,8 @@ class ForeignKey
         // Valid Options
         $validOptions = array('delete', 'update', 'constraint');
         foreach ($options as $option => $value) {
-            if (!in_array($option, $validOptions)) {
-                throw new \RuntimeException('\'' . $option . '\' is not a valid foreign key option.');
+            if (!in_array($option, $validOptions, true)) {
+                throw new \RuntimeException(sprintf('"%s" is not a valid foreign key option.', $option));
             }
 
             // handle $options['delete'] as $options['update']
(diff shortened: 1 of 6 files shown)


rquadling:
Should the `Index` and `ForeignKey` also have the same guarding?

src/Phinx/Db/Table/ForeignKey.php
```diff
@@ -216,8 +216,8 @@ public function setOptions($options)
         // Valid Options
         $validOptions = array('delete', 'update', 'constraint');
         foreach ($options as $option => $value) {
-            if (!in_array($option, $validOptions)) {
-                throw new \RuntimeException('\'' . $option . '\' is not a valid foreign key option.');
+            if (!in_array($option, $validOptions, true)) {
+                throw new \RuntimeException(sprintf('"%s" is not a valid foreign key option.', $option));
```
f3ath: Unified the exception messages

src/Phinx/Db/Table/Index.php
```diff
@@ -165,8 +165,8 @@ public function setOptions($options)
         // Valid Options
         $validOptions = array('type', 'unique', 'name', 'limit');
         foreach ($options as $option => $value) {
-            if (!in_array($option, $validOptions)) {
-                throw new \RuntimeException('\'' . $option . '\' is not a valid index option.');
+            if (!in_array($option, $validOptions, true)) {
+                throw new \RuntimeException(sprintf('"%s" is not a valid index option.', $option));
```
f3ath: Unified the exception messages

f3ath:
@rquadling done, with a cleaner solution

rquadling:
Excellent work. Thank you.

PR #911, opened 2016-07-19
Printing migration and seed paths more nicely

f3ath:


rquadling:
One small issue. If the seeds directory doesn't exist (as it is optional), nothing is printed for the seed directory, even though it may be configured in the config.

For example.

Without realpath I get
```
Phinx by Rob Morgan - https://phinx.org. version 0.6.3

using config file .\phinx.php
using config parser php
using migration path C:\wamp\www\app/migrations
using seed path C:\wamp\www\app/seeds
using environment development
```

which does look ugly with all the slashes not being the right way around for Windows.

With `realpath()`, I get ...

```
Phinx by Rob Morgan - https://phinx.org. version 0.6.3

using config file .\phinx.php
using config parser php
using migration path C:\wamp\www\app\migrations
using seed path
using environment development
```

Also, the location of the config file could probably do with the same love.

f3ath:
@robmorgan I believe the second commit indirectly solves that issue.

lorenzo:
Thanks!

PR #921, opened 2016-07-25
To use phinx for unit testing, sometimes is necessary to run not just…

cauelorenzato:
… the migrations, but the seeds as well. This commit add a new method getSeed that runs the seeds.

One commit only as you asked.

### DIFF
--- a/src/Phinx/Wrapper/TextWrapper.php
+++ b/src/Phinx/Wrapper/TextWrapper.php
@@ -126,9 +126,9 @@ class TextWrapper
 
     /**
      * Returns the output from running the "seed:run" command.
-     * @param  string $env environment name (optional)
-     * @param  string $target target version (optional)
-     * @param  string $seed seed name (optional)
+     * @param  string|null $env environment name
+     * @param  string|null $target target version
+     * @param  string|null $seed seed name
      * @return string
      */
     public function getSeed($env = null, $target = null, $seed = null)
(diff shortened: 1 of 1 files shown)


src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -125,6 +125,29 @@ public function getMigrate($env = null, $target = null)
     }
 
     /**
+     * Returns the output from running the "seed:run" command.
+     * @param  string $env environment name (optional)
+     * @param  string $target target version (optional)
+     * @return string
+     */
+    public function getSeed($env = null, $target = null, $seed = null)
```
rquadling: Please add $seed to docblock

cauelorenzato: Done!

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -125,6 +125,30 @@ public function getMigrate($env = null, $target = null)
     }
 
     /**
+     * Returns the output from running the "seed:run" command.
+     * @param  string $env environment name (optional)
```
rquadling: Can you use `string|null` as the type and drop `(optional)` please.

cauelorenzato: of course, done now.

rquadling:
Thank you.

PR #926, opened 2016-07-28
Rollbacks with Version Order

daniel-gomes-sociomantic:
A new Configuration option named `version_order` was added. Its possible values are:
- `creation` (the default): causes the Rollback and Status commands to order the executed migrations by their Creation Times (aka Migration IDs or Names).
- `execution`: causes the Rollback and Status commands to order the executed migrations by their Execution Times (aka Start Times).

This means that when invoking the Rollback command:
- With an `execution` version order, the last executed migration will be rollbacked.
- With an `execution` version order and a `-d (date)` option, the migration start times will be used to determine the target version to rollback to. In other words, all migrations which were executed after that date will be rollbacked.
- With an `execution` version order and a `-t (target version)` option, all migrations which were executed after the target version was executed will be rollbacked.

In terms of testing:
- More test cases were added.
- General test improvements were done (like adding the expected arguments to the underlying Manager calls in `RollbackTest` and `StatusTest`.

Other notes:
- The old `rollback` and `rollbackToDateTime` methods were merged into a single `rollback` method that handles all cases and version orders.
- This PR replaces https://github.com/robmorgan/phinx/pull/901.


daniel-gomes-sociomantic:
To do: act on https://github.com/robmorgan/phinx/pull/901#issuecomment-234794682

daniel-gomes-sociomantic:
@rquadling regarding your test cases, I think I can simplify the list a bit:

> Version order setting - null/undefined, 'creation', 'execution', invalid.

Invalid version orders are tested in the [PdoAdapterTest](https://github.com/daniel-gomes-sociomantic/phinx/blob/4e045b417bc66ccf4d7e7c7e9ec056d4a0ae0a96/tests/Phinx/Db/Adapter/PdoAdapterTest.php#L95) and the [ManagerTest](https://github.com/daniel-gomes-sociomantic/phinx/blob/4e045b417bc66ccf4d7e7c7e9ec056d4a0ae0a96/tests/Phinx/Migration/ManagerTest.php#L440).

> Migrations order - same order, reversed order, mixed.

Not quite sure what you mean, maybe you meant version order? If so, it is tested in the [PdoAdapterTest](https://github.com/daniel-gomes-sociomantic/phinx/blob/4e045b417bc66ccf4d7e7c7e9ec056d4a0ae0a96/tests/Phinx/Db/Adapter/PdoAdapterTest.php#L37).

> Breakpoint settings - setting is after target, setting is on target, setting is after before target.

Should be included in the PR already.

> Rollback target - after all migrations, is an exact match for a migration, between 2 migrations, before all migrations.

Should also be included in the PR already. I also made different tests for real targets (ie. `-t/--target`) and date targets (ie. `-d/--date`).

> Number of migrations - 0, 1, 2, 3+

Coverage here could probably be improved.

t1gor:
This sounds really promising. @daniel-gomes-sociomantic could you please provide an example with status commands outputs to make it easier? I guess that could also work as documentation in future. Of maybe that exists somewhere already ...

rquadling:
The code is in the `0.7.x-dev` branch. If you want to try it out .. and see if it all works as you expect.

daniel-gomes-sociomantic:
here it is @t1gor :

default version order (creation):

phinx 1

execution version order:

phinx 2

rquadling:
Hello. Sorry for the significant delay in getting back to you.

We have now moved to using master as our code base, so can you please, once again (truly sorry) rebase onto master.

HOPEFULLY everything will be good to go as soon as this is done!!!

daniel-gomes-sociomantic:
replaced with https://github.com/robmorgan/phinx/pull/1045

PR #927, opened 2016-07-28
Ability to specify multiple paths for migrations and seeds (0.7.x-dev branch)

bartosz-maciaszek:
Entering the inception now. Rebased and squashed version of [this pull request](https://github.com/robmorgan/phinx/pull/875).


src/Phinx/Console/Command/Create.php
```diff
@@ -79,6 +83,63 @@ protected function getCreateMigrationDirectoryQuestion()
...
+            throw new \Exception(
+                'You probably used curly braces to define migration path in your Phinx configuration file, ' .
+                'but no directories have been matched using this pattern. ' .
+                'You need to create a migration directory manually.'
+            );
+        }
+
+        // Only one path set, so select that:
+        if (1 === count($paths)) {
+            return array_shift($paths);
```
mathbr: Why not simply return `$paths[0]` instead of an array operation?

rquadling: One reason to not use `$paths[0]` is if the indicies aren't that. Having said that, they should be, but, someone may have done something clever in `Util::globAll()`.

Maybe `return reset($paths);` as that is the first path, regardless of index.

bartosz-maciaszek: Does it really make any difference for an array with just one item?

rquadling: Only if the indicies are not numeric or ordered. I'm happy with what it is at the moment.

mathbr: @rquadling That indeed makes sense, thanks for the hint.

rquadling:
Thank you.

PR #930, opened 2016-08-01
bugfix: schema support for postgresql is broken

tpetry:
bugfix: handles support for creating a phinx migration inside custom postgresql schemas (referencing issue robmorgan/phinx#915)

PostgreSQL supports schemas to group tables which phinx has supported since introduction of the postgresql adapter. But with commit robmorgan/phinx@a5d7e65e475e6579ff64adad7da1db36c116d29e the setting of the schema search path has been moved from connection setup (set schema search path at connection phase for every following query) to only the adapter functions that need working on different schemas than the default `public` schema. But the functions for getting the actual version and the applied versions seems to have been missed.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -1039,7 +1039,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
             $this->createSchema($this->getSchemaName());
         }
 
-        $this->fetchAll(sprintf('SET search_path TO %s', $this->getSchemaName()));
+        $this->execute(sprintf('SET search_path TO %s,"$user",public', $this->getSchemaName()));
 
         return parent::createSchemaTable();
     }
@@ -1179,5 +1179,25 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
     public function castToBool($value)
     {
         return (bool) $value ? 'TRUE' : 'FALSE';
+    }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function getVersions()
+    {
+        $this->fetchAll(sprintf('SET search_path TO %s,"$user",public', $this->getSchemaName()));
+
+        return parent::getVersions();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getVersionLog()
+    {
+        $this->fetchAll(sprintf('SET search_path TO %s,"$user",public', $this->getSchemaName()));
+
+        return parent::getVersionLog();
     }
 }


rquadling:
Wouldn't have `\Phinx\Db\Adapter\PostgresAdapter` override `\Phinx\Db\Adapter\PdoAdapter::getSchemaTableName()` to include the schema be the way to go?

rquadling:
And then, all the `SET SEARCH`s related to the phinxlog table could be removed?

tpetry:
Tried it and it's possible but requires some more modification, which are not as clear this pull request:

For example the `\Phinx\Db\Adapter\PostgresAdapter::createTable(Table $table)` creates a primary constraint based on the table name. If the table name contains the schema (which would be true for everyone operating phinx with schema support) it would leak into the constraint name which seems kind of ugly and fight the idea of schemas to have e.g. _identical_ tables within different schemas. To remove the schema name from the constraint the schema would have been to be connected through all the layers...

I would strictly prefer the actual committed solution while it is clean and simple and does not need any refactoring on the phinx core and really only bugfixes the actual problem. What is your opinion on this @rquadling.

tpetry:
@rquadling Anything to do before merging? This is actually a blocking task for us.

rquadling:
Hi. Can you rebase onto 0.7.x-dev please.

tpetry:
@rquadling Done. Is there an expected release date for 0.7?

rquadling:
Hi. Been away. Back now. I'll be clearing various PRs this week. Looking for a release next week.

tpetry:
updated the pull request because working with schemas did not only break the phinxlog table it even broke working with any of the core postgresql functions because phinx's setting of the search path did not include the `$user` and `public` schema

rolfvreijdenberger:
core functions/extensions are only there if you place them there. administrators often remove the public schema.
it is a convention, not a given

rquadling:
Does the result of `getSchemaName()` need to be escaped/quoted before using it?

rolfvreijdenberger:
@rquadling apologies if I sounded a bit absolute in my comment. It's not a hard requirement, just something I noticed :)

the patch as is would be fine already. just a bit of feedback that might be nice for some if implemented, that's why I said I would be fine with doing some work for it.
schemanames are usually lowercase and limited in accepted characters. 
setting the search path does not fail for nonexistent schemas, so also not for schemas with invalid names I assume.

see https://www.postgresql.org/docs/9.4/static/ddl-schemas.html 5.7.6 usage patterns. for some more info on the previous discussion

rolfvreijdenberger:
on rereading your comment I now understand what you're saying.
the schema name can be input by the user and will be used in a query, so in theory SQL injection would be possible. Syntax for 'set search_path'  is very limited in functionality. But if you can run multiple queries in one go (query1;query2;query3;) then injection is a threat. so yes, escaping would be prudent.

tpetry:
Setting the charset is a configuration option. Would you really target yourself with an sql injection in a configuration option, i mean it's not running user input.

Will maybe add escaping the next week, but has absolutely no priority on my list.

rolfvreijdenberger:
@tpetry no priority for me also and it's a minor thing.

But, through escalation of privileges etc someone might gain access to the configuration settings used to do an attack on the db. 
So theoretically it is possible. My previous comment is more of a theoretical nature than something I expect
The config would  probably contain the database credentials anyway, making an attack possible. but then, maybe you do not have access to the host where the database server is located, but the builds/deployment server does. so attacks via a different channel would be possible, but highly improbable.

rolfvreijdenberger:
otoh: I tested the patch on a production server and it work well. thanks.

rolfvreijdenberger:
Hi @tpetry and @rquadling,

It's been awhile since there was activity here.We really need this functionality before a certain time and if there is no clear target for this PR we need to start running against a fork (which I do not prefer).
is there any progress on merging this PR?
Also is there a release date for the 0.7 branch?

If there is anything I can do to help you please let me know.

tpetry:
Hi @rolfvreijdenberger,
there has not been any code commit to phinx since Oct 27, 2016 - neither for 0.7 nor 0.6. I don't expect my change to be merged anytime soon. You will have to use a fork like i did.

robmorgan:
hi @rolfvreijdenberger sorry about the delay. We'll ship the first of 0.7 releases next month. This change can go. I'm not concerned about the sql injection attack but we should probably change the `fetchAll()` statements to `execute()` as we don't need a return value.

rolfvreijdenberger:
@robmorgan no need for any apologies whatsoever :) I understand these things can take time and we have other options so no worries.
With that out of the way: good news :+1: 
It's not clear to me who will make the change to `execute()`?
Is this something blocking?

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -1039,7 +1039,7 @@ public function createSchemaTable()
             $this->createSchema($this->getSchemaName());
         }
 
-        $this->fetchAll(sprintf('SET search_path TO %s', $this->getSchemaName()));
+        $this->fetchAll(sprintf('SET search_path TO %s,"$user",public', $this->getSchemaName()));
```
robmorgan: change `fetchAll()` to `execute()`

JayPHP:
Failing test needs to be resolved before this is merged

dereuromark:
This requires a new PR now it seems.

dereuromark:
PR https://github.com/cakephp/phinx/pull/1072

PR #936, opened 2016-08-16
Adding new options to phinx. Up and Down will deploy only an specific…

caherrera:
… migration (for dev env propuses only)

### DIFF
--- a/src/Phinx/Console/Command/Down.php
+++ b/src/Phinx/Console/Command/Down.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * Phinx
+ *
+ * (The MIT license)
+ * Copyright (c) 2015 Rob Morgan
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated * documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


caherrera:
This is a simply backdoor to work with your migration only and use multiple devs works in parallel.

jradtilbrook:
This is absolutely a beneficial function for my usage. One comment though, are you planning to update the documentation to reflect the new commands?

caherrera:
Sure. Let me try.

Same pull request right ?



Carlos

jradtilbrook:
If you are asking me specifically I'm afraid I can only assume yes same pull request because I'm as new as you to contributing. I just though I would register my appreciation for such a function :+1:

rquadling:
I'm not sure adding new commands that perform very similar functionality as existing commands is the way to go.

Can this be handled by additional options on the `migrate` and `rollback` commands?

In 0.7.x-dev branch, we have a new method of handling migrations/rollbacks based upon execution datetime, rather than migration creation datetime.

This solves issues with multiple developers creating migrations that get merged into master where they interleave with pre-existing (and potentially already executed) migrations. By organising migrations based upon execution time, when the migration was created becomes irrelevant and the `--target` is of less use. For rollbacks, rolling back to a datetime is now as simple as using the `--date` option.

So, nominating a single migration to execute or rollback, is this really a migration? If it has no impact as to when it can run, would a seed be a better option? Of course, undoing a 'seed' isn't possible, so maybe 

Can you take a look at the new branch before moving forward with any PRs please.

Having said all of that, you can add your own commands very easily by extending `\Phinx\Console\PhinxApplication`, and implement `doRun()` to add your commands before calling the `parent::doRun()`.

This would require zero effort on our part.

If you already have your own tool chain, then you can get the commands from Phinx easily enough.

Assuming you are using Symfony as a base (very easy to use)...

```php
$phinx = new \Phinx\Console\PhinxApplication;
$phinxCommands = [];
foreach ($this->phinxApplication->all() as $commandName => $command) {
  $phinxCommands[] = $command->setName('phinx:' . str_replace(':', '-', $commandName));
}
$this->addCommands($phinxCommands);
```

This will then make your tool contain the phinx commands, so, when you run `tool --help`, you'll see
```
 phinx
  phinx:breakpoint                    Manage breakpoints
  phinx:create                        Create a new migration
  phinx:help                          Displays help for a command
  phinx:init                          Initialize the application for Phinx
  phinx:list                          Lists commands
  phinx:migrate                       Migrate the database
  phinx:rollback                      Rollback the last or to a specific migration
  phinx:seed-create                   Create a new database seeder
  phinx:seed-run                      Run database seeders
  phinx:status                        Show migration status
  phinx:test                          Verify the configuration file
```

JayPHP:
Looking to close this branch due to it's age and no response from author. If you would like this kept open, please let me know and open a new issue.

dereuromark:
Closing as this got stale.

PR #939, opened 2016-08-16
Close #938. Rollback target uses migration name

### LINKED ISSUE #938: Rollback with target id
There is an issue with rolling back using the -t option and not passing a valid parameter. If passing a string or a number with a missing digit it will rollback all migrations. This is because from https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Migration/Manager.php#L390 to line 409 the logic is wrong. As long as any number supplied is less than the first migration, it will rollback all. It should be an explicit check for 0, not a comparison against the first migration id.

The current concept of rolling back to a target "id" doesn't make much sense anyway. Would it not be better to specify the migration name rather than the migration id since the -d date parameter basically does what -t does?

jradtilbrook:
Since the -d and -t are so similar, this changes the functionality of -t to use the migration name as specified when creating a migration, rather than an arbitrary integer. This accounts for a typo entered with the -t option causing all migrations to rollback.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -119,20 +119,21 @@ parameter or ``-t`` for short.
 
 .. code-block:: bash
 
-        $ phinx rollback -e development -t 20120103083322
+        $ phinx rollback -e development -t MyNewMigration
 
-Specifying 0 as the target version will revert all migrations.
+Specifying all as the target version will revert all migrations. This also
+works if specifying 0.
 
 .. code-block:: bash
 
-        $ phinx rollback -e development -t 0
+        $ phinx rollback -e development -t all
 
 If a breakpoint is set, blocking further rollbacks, you can override the
 breakpoint using the ``--force`` parameter or ``-f`` for short.
 
 .. code-block:: bash
 
-        $ phinx rollback -e development -t 0 -f
+        $ phinx rollback -e development -t all -f
 
 The Status Command
 ------------------
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -119,21 +119,21 @@ parameter or ``-t`` for short.
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


robmorgan:
Hmmmm I don't actually mind this, but I need to think about all the use cases. This is obviously a significant change in how Phinx was originally designed to work and what our users expect. Even Rails still works with integers and versions.

jradtilbrook:
Yeah, I understand it's quite a change and might have to wait for at least a minor version release. I truly believe it makes more sense this way. I think it's important to explicitly check for `0` passed to the `-t`, rather than a `<` comparison because basically any typo or a missed number in the id will cause it to revert all migrations.

I'll work on fixing the unit tests and just leave the pull request here for if/when you are ready :smiley:

jradtilbrook:
Okay, so what I've done is add an explicit check for `0` or `all` passed to the `-t` parameter. I think this isn't too much of a change and could be included quite quickly. It will work as the documentation states - use 0 to reset all, but also isn't vulnerable to typos that may cause all migrations to be reverted by mistake.

I cannot seem to get the HHVM tests to pass, I haven't dealt with HHVM before and can't understand whats going wrong - it seems to be an issue with code I haven't touched.

rquadling:
Can you rebase against 0.7.x-dev please. See some of the changes there.

jradtilbrook:
Okay, sorry for the delay. I have now rebased the changes onto 0.7.x-dev. Would you like me to open a new pull request since I can't change the from branch of this PR? I also see that this branch is failing the HHVM tests with the same error as me if that is any help.

rquadling:
Travis's HHVM is not happy. Create a new PR and close this one.

rquadling:
But add a note saying that it is replacing this PR.

jradtilbrook:
Closed in favour of #945

PR #943, opened 2016-08-23
Adding hint on migration name in CamelCase

igorsantos07:
Nothing much to say besides I get annoyed as I never remember if the migrations should be snake_case as the file or CamelCase as the class. A tip about that on the help could be useful, instead of hitting the errorwall :)

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -52,7 +52,7 @@ class Create extends AbstractCommand
 
         $this->setName('create')
             ->setDescription('Create a new migration')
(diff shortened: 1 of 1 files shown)


rquadling:
I wonder if snake_case and UpperCamelCase should both be accepted?

One thing I have locally is an amendment that allows me to apply a mask to the name. We enforce our devs to use DT_nnn_xxxx_xxxx_xxxx as the DTnnn number is our Jira task. That way, each migration is easily identifiable to a single task.

robmorgan:
@igorsantos07 thanks

PR #945, opened 2016-08-24
Rollback --target using migration name

jradtilbrook:
This replaces PR #939.

There is a bug where a any number less than the first migration time stamp (or a typo) supplied to the `--target` option will cause all migrations to be rolled back. Referring to PR #939, I don't think this should be the case and optimistically the `--target` option should actually take the migration name, not the migration id (time stamp) because this functionality is basically incorporated into `--date`. 

What this pull request does is add an explicit check for `all` or `0` passed to the `--target` option to trigger a rollback of all migrations. This doesn't allow typos or small numbers to incorrectly rollback all.

Going forward, I would either prefer the `--target` functionality be changed to take the name, or a new option be added to specify the name. Either way, there may be an issue of non-unique migration names.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -420,10 +420,6 @@ class Manager
         // note that the version log are also indexed by name with the proper ascending order according to the version order
         $executedVersions = $this->getEnvironment($environment)->getVersionLog();
 
-        if ($target === "0") {
-            $target = 0;
-        }
-
         // get a list of migrations sorted in the opposite way of the executed versions
         $sortedMigrations = array();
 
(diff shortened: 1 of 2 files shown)


jradtilbrook:
This will close #938.

jradtilbrook:
Looks like HHVM is working again of Travis. My fork of this is passing all tests, but I can't re-trigger a build for this PR. Do you think you will add this functionality change?

src/Phinx/Migration/Manager.php
```diff
@@ -392,13 +392,16 @@ public function rollback($environment, $version = null, $force = false)
...
 
-            // If the target version is before the first migration, revert all migrations
-            if ($version < $first) {
-                $version = 0;
+            // Try to find a migration id with the supplied version name
+            $found = array_search($version, $migrationNames);
+            if ($found !== false) {
+                $version = $found;
             }
         }
```
lorenzo: You should check whether or not the version was actually found. If it was not found, throw an error. No one wants to rollback to the first migration just because they made a typo in the cli

jradtilbrook: Nice catch. I'll update the tests and add that change in. I'm a little busy for the next few days but I'll see how I go by the weekend

lorenzo: No problem, thanks for your work!

jradtilbrook:
I didn't get much time over the weekend to look at this because I had to set up my laptop again.  
I see that this PR is for an old branch, can you advise what I should change it to? The Travis config is also old and contains HHVM, looks like I'll need to merge in master?

lorenzo:
@jradtilbrook Try to rebase the branch on top of the current master. It should work without any issues

tariquesani:
@jradtilbrook Hey! are you still working on this?

jradtilbrook:
@tariquesani I still plan to but have been quite busy recently. I'm not sure when I'll be able to get some code out for you - if you're wanting something soon maybe its best for someone to take it over instead?

jradtilbrook:
Alright, so I finally got around to working on this. Rebasing turned out to be too hard so I started again from master. I also updated this PR to point to master as per the CONTRIBUTING doc. Hope its all fine, let me know your thoughts/intentions

lorenzo:
@jradtilbrook thanks for doing this!

PR #949, opened 2016-08-29
Fix for Config::getMigrationBaseClassName

thorne51:
Ran into an issue where if the `migration_base_class` option is specified without a namespace, the class name returned empty in the migration generation.


shadowhand:
Thanks!

thorne51:
You're welcome! Thanks for the great library!

PR #955, opened 2016-09-10
Merge with basilfx and adding Postgres Adapter Dump

celorodovalho:


dereuromark:
Too much unclear changes. Please re-do in smaller chunks and with some text if possible.

PR #956, opened 2016-09-11
Fix seed command accepting string

ninjapenguin:
Fixes TextWrapper which was unable to accept seed name correctly due to SeedRun::execute always expecting an array

### DIFF
--- a/src/Phinx/Wrapper/TextWrapper.php
+++ b/src/Phinx/Wrapper/TextWrapper.php
@@ -126,9 +126,9 @@ class TextWrapper
 
     /**
      * Returns the output from running the "seed:run" command.
-     * @param  string|null $env environment name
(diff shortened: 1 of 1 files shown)


src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -143,6 +143,7 @@ public function getSeed($env = null, $target = null, $seed = null)
             $command += array('-t' => $target);
         }
         if ($seed) {
+            $seed = is_array($seed) ? $seed : [$seed];
```
shadowhand: Why not type cast it?

ninjapenguin: Yep - nice call, I've updated the PR. Thanks

PR #968, opened 2016-10-12
Document supported table options

adduc:


docs/migrations.rst
```diff
@@ -464,6 +464,38 @@ To simply change the name of the primary key, we need to override the default ``
...
+========= ===========
+collation set collation of table (defaults to utf8_general_ci)
+comment   set a text comment on table
+engine    set engine (defaults to InnoDB)
+========= ===========
+
+For the PostgreSQL adapter:
+
+======= ===========
+Option    Description
```
rquadling: A couple of spaces.

adduc: Thanks, I've fixed the spacing. Are there any additional changes needed?

docs/migrations.rst
```diff
@@ -464,6 +464,38 @@ To simply change the name of the primary key, we need to override the default ``
...
+id          set name of automatically created identity field
+primary_key set field to use for primary key (defaults to value of id)
+=========== ===========
+
+For the MySQL adapter:
+
+========= ===========
+Option    Description
+========= ===========
+collation set collation of table (defaults to utf8_general_ci)
```
adduc: Yes, charset is also available (I'll add it in a subsequent commit).

By config overrides, are you referring to setting charset/collation in the configuration file?

adduc: Looking into [`Phinx\Db\Adapter\MysqlAdapter::createTable`](https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L243), it doesn't look like charset is supported as a table option (collation is used to determine character set).

dereuromark:
ping @adduc 
Anything missing or can we merge?

adduc:
I think so, although I'm not sure if there's been any changes to supported table options since 0.6.

dereuromark:
I see. But even with it being not fully complete this would already be an improvement here.

dereuromark:
@lorenzo Can we already merge this partial docs upgrade?

PR #976, opened 2016-11-01
dropConstraint method added

chilimatic:
This adds a "dropConstraint" method to the table class:

it will drop the primary keys in MySQL and SQLite and in pgSQL it invokes the drop foreign key with the contraint parameter which is basically the same behaviour. I did not want to automate it like I did with mysql or sqlite since in pSQL everything is a constraint :) so only dropping things based on your naming convention does not cover all cases.

The reason I did not name it "dropPrimaryKey" is because pgSQL does not have primary keys and primary keys are only special cases of constraints.

I ran and added unittests for SQLite / MySQL / psql.

I updated the unittests to use the mockbuilder and not getMock for all libraries 
and DRY the SQLiteadapter.

This was blocking my development so please provide me feedback if you want to merge it or i have to maintain it myself.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -189,7 +189,8 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             "SELECT TABLE_NAME
             FROM INFORMATION_SCHEMA.TABLES
             WHERE TABLE_SCHEMA = '%s' AND TABLE_NAME = '%s'",
-            $options['name'], $tableName
+            (string) isset($options['name']) ? $options['name'] : '',
+            $tableName
         ));
 
         return !empty($exists);
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -331,6 +331,18 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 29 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -502,7 +502,7 @@ public function dropColumn($tableName, $columnName)
      * @param string $tableName Table Name
      * @return array
      */
-    protected function getIndexes($tableName)
+    public function getIndexes($tableName)
```
chilimatic: needed for the unitests -> to get the current indexes and not the ones in the runtime within the table object

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -627,7 +639,6 @@ public function hasForeignKey($tableName, $columns, $constraint = null)
             return false;
         } else {
             foreach ($foreignKeys as $key) {
-                $a = array_diff($columns, $key['columns']);
```
chilimatic: not used hence removed

chilimatic:
... how could those test even work ? ::class was never supported in php5.4 ...

shadowhand:
There's a lot of style changes being made here. I would prefer to see style separate from functionality.

chilimatic:
Hmmm I will, if I got time. The minor part is style changes and most of it is DRY in the SQLiteAdapter and rewriting the unitests to use the mockbuilder instead of getMock.

I ran all the tests several times and debugged them and I already switched to my fork so I have the functionality in my project.

But I can understand that to many changes are confusing.

PR #977, opened 2016-11-02
Unable to create log table in mysql version < 5.6 with NO_ZERO_DATE mode. Fixed.

incrize:
This fix solves the problem with the phinxlog table creation in MySQL versions earlier than 5.6.
To reproduce the problem:
1) Install MySQL 5.5.
2) Configure sql_mode this way: `sql_mode=NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE`
3) Start a migration.

Note than in MySQL versions earlier than 5.6 you shouldn't create 2 columns with the `TIMESTAMP` type and `CURRENT_TIMESTAMP` as the default value.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -508,12 +508,11 @@ abstract class PdoAdapter implements AdapterInterface
 
             $table = new Table($this->getSchemaTableName(), $options, $this);
 
-            if ($this->getConnection()->getAttribute(\PDO::ATTR_DRIVER_NAME) === 'mysql'
-                && version_compare($this->getConnection()->getAttribute(\PDO::ATTR_SERVER_VERSION), '5.6.0', '>=')) {
+            if ($this->getConnection()->getAttribute(\PDO::ATTR_DRIVER_NAME) === 'mysql') {
                 $table->addColumn('version', 'biginteger', array('limit' => 14))
                       ->addColumn('migration_name', 'string', array('limit' => 100, 'default' => null, 'null' => true))
-                      ->addColumn('start_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
-                      ->addColumn('end_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
+                      ->addColumn('start_time', 'timestamp', array('default' => null, 'null' => true))
+                      ->addColumn('end_time', 'timestamp', array('default' => null, 'null' => true))
                       ->addColumn('breakpoint', 'boolean', array('default' => false))
                       ->save();
             } else {


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -508,12 +508,11 @@ public function createSchemaTable()
 
             $table = new Table($this->getSchemaTableName(), $options, $this);
 
-            if ($this->getConnection()->getAttribute(\PDO::ATTR_DRIVER_NAME) === 'mysql'
-                && version_compare($this->getConnection()->getAttribute(\PDO::ATTR_SERVER_VERSION), '5.6.0', '>=')) {
```
shadowhand: Isn't this `version_compare` condition trying to solve the exact issue you are describing? It looks like the baseline would be get worse for everyone with this change, not just users on MySQL < 5.6.

incrize: No it is not.
This code

```php
$table->addColumn('version', 'biginteger')
  ->addColumn('migration_name', 'string', array('limit' => 100, 'default' => null, 'null' => true))
  ->addColumn('start_time', 'timestamp')
  ->addColumn('end_time', 'timestamp')
  ->save();
```
will generate the following SQL query:
```sql
CREATE TABLE `phinxlog` (
    `version` BIGINT(20) NOT NULL, 
    `migration_name` VARCHAR(100) NULL, 
    `start_time` TIMESTAMP NOT NULL, 
    `end_time` TIMESTAMP NOT NULL, 
    PRIMARY KEY (`version`)
) ENGINE = InnoDB CHARACTER SET utf8 COLLATE utf8_general_ci;
```
If MySQL is configured with `sql_mode=STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE`, then the query won't be executed due to the
`Invalid default value for 'end_time'` error, because there is no default value.

It is logical to use this code:
```php
$table->addColumn('version', 'biginteger')
    ->addColumn('migration_name', 'string', array('limit' => 100, 'default' => null, 'null' => true))
    ->addColumn('start_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
    ->addColumn('end_time', 'timestamp', array('default' => 'CURRENT_TIMESTAMP'))
    ->save();
```
As a result, the following SQL query will be formed:
```sql
CREATE TABLE `phinxlog` (
    `version` BIGINT(20) NOT NULL, 
    `migration_name` VARCHAR(100) NULL, 
    `start_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, 
    `end_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, 
    PRIMARY KEY (`version`)
) ENGINE = InnoDB CHARACTER SET utf8 COLLATE utf8_general_ci;
```
But this will only work with MySQL 5.6 or higher.

----
Because phinx sets timestamps automatically when adding entries to the log, I suggest using `NULL` as the default value, because it will work for all MySQL versions.

Matt3o12:
For those who also encounter this issue (and don't want to change the SQL Mode globally), one workaround is to use the PHP config file and change the sql mode manually for the current session:

```php
//file: phinx.php
<?php
$conn = PDO("...");
$conn->query("SET SESSION sql_mode = ''");

return ["environments" => ["db" => ["connection" => $conn]]]
```

[Check the docs](http://docs.phinx.org/en/latest/commands.html#configuration-file-parameter) for more information about how the php configuration works.

ravage84:
Shouldn't this be fixed by #1049?

rquadling:
@ravage84, I think this issue is now dead. As Phinx always sets both values when inserting the row, any defaults are redundant.

rquadling:
I'm closing this request. If anyone feels that is inappropriate, please create a new PR based upon master.

PR #979, opened 2016-11-07
Add gitattributes to exclude tests folder

pavlakis:
Added `.gitattributes` to exclude the tests on export - e.g. when pulled in via composer.


robmorgan:
@pavlakis thanks

Issue #986, opened 2016-11-22
sqlite adapter is suppressing errors


aldas:
MySql/Postgres have `array(\PDO::ATTR_ERRMODE => \PDO::ERRMODE_EXCEPTION)` set for PDO options which enables exceptions when dml/ddl fail but sqlite does not have it.

so if you have for example faulty DDL
```
$table = $this->table('users');
$table->addColumn('username', 'text')
    ->addColumn('username', 'integer')
    ->create();
```

this migration is applied "successfully" but no table is created

but if you would add `$db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);` into `SQLiteAdapter.connect()` it would fail with

> [PDOException]                                                    
  SQLSTATE[HY000]: General error: 1 duplicate column name: username

dereuromark:
What is your suggested fix?

aldas:
me PR to fix this https://github.com/cakephp/phinx/pull/989 was closed as already implemented. this is pretty old issue so I assume it got fixed

PR #989, opened 2016-12-02
SqliteAdapter should set be throw exceptions when queries fail

aldas:
fixes issue #986 and some failing tests under windows


semoriil:
Why this is not applied? AppVeyor build failed - that's network error of service itself...
I had some trouble because of this silent error dropping and reporting failed migrations as successful.

lorenzo:
already done

PR #990, opened 2016-12-02
Sqlite create column with notnull without default

aldas:
Fix for issue #265 - Allow table creating with SqliteAdapter to have NOT NULL columns without DEFAULT keyword.

no need to worry about ALTER COLUMN as
```
alter table person add column age integer not null;
```
will result with anyway

> SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL

but this has one sideeffect - when user tries to `$table->changeColumn(...)`:
* change NULL column to NOT NULL 
* and does not provide DEFAULT options
* and existing table has rows with null value in this column

Phinx would recreate table with NOT NULL column and `insert select` all data from old table to new. During this Sqlite SILENTLY ignores all values. I tried to do `INSERT OR FAIL ...` but this will just ignore problematic rows. so data loss occurs (Tested on Windows PhP 5.6.x).

My PR will check before first DDL if dataloss could occur and if there are rows with null values then Exception will be thrown

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -407,7 +407,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
             $writeName = $selectName;
             if ($selectName === $columnName) {
-                if ($column['notnull'] === '0' && !$newColumn->isNull() && empty ($newColumn->getDefault())) {
+                if ($column['notnull'] === '0' && !$newColumn->isNull() && trim($newColumn->getDefault()) !== false) {
                     // it is necessary to check when changing NULL column to NOT NULL without DEFAULT options that
                     // there should be no existing rows with null values. Sqlite will SILENTLY IGNORE these rows when we would
                     // insert select them to the new table
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -1043,7 +1043,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
         $default = $column->getDefault();
 
-        $def .= ($column->isNull() || is_null($default)) ? ' NULL' : ' NOT NULL';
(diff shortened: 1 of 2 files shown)


dereuromark:
Given the SQL errors otherwise this sounds like a good fix! 👍

jakejohns:
any word here?

gizahNL:
any reason why this was never merged?!?

dereuromark:
Can we resolve the conflicts? Then we can merge.

aldas:
This is not that simple. Change at https://github.com/cakephp/phinx/blob/9c4918e3ff12353217c78a79cacb6e016a7bd8c9/src/Phinx/Db/Adapter/SQLiteAdapter.php#L1461

to
```
$def .= (!$column->isIdentity() && $column->isNull()) ? ' NULL' : ' NOT NULL';
```

is **breaking** change. 

By Phinx defaults all columns are NOT NULL `column->null = false`  and if we would remove  that  `OR default is null` part  SQLiteAdapter would start creating columns  by default as `NOT NULL`  instead of current behaviour of nullable column. 

This means that how columns default to nullable/not nullable is different for Mysql/Postgres and SQLite in Phinx.

so currently 
```
$table->addColumn('email', 'string')->save();
```

would create column as nullable in Sqlite

after change column would be NOT NULL e.g. equal to 
```
$table->addColumn('email', 'string', ['null' => false])->save();
```

and not to get that behaviour users should change their code to
```
$table->addColumn('email', 'string', ['null' => true])->save();
```

charlienux:
@aldas, I don't know if it will be an appropriate solution, but to define by default that columns are created as NULL in case of **sqlite**, the following condition can be included in the **addColumn** method of **phinx/src/Phinx/Db/Table.php**:
```
            if ('sqlite' == $this->getAdapter()->getAdapterType()) {
               array_key_exists('null', $options) ?: $options['null'] = true;
            }
```
Before the sentence:
```
$action = AddColumn::build($this->table, $columnName, $type, $options);
```
Thus the code can be applied without a breaking change.

dereuromark:
Is this resolved with 0.12 now?

othercorey:
Closing due to no activity.

Issue #991, opened 2016-12-06
Custom DB schema for SQL Server


tm8747a:
I tried adding "schema" to the phinx YAML config file and noticed it had no impact with a SQL server adapter. Seems only Postgres has that feature. Would be nice to have this working for SQL Server as well. I'll see if I can work on it and put in a pull request.

rolfvreijdenberger:
referencing #930 fix for postgres schema functionality

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

PR #992, opened 2016-12-06
#991 - Add ability to use different schema with SQL Server adapter

tm8747a:
This allows for specifying a DB schema, similar to what exists for Postgres, so that you can use a schema different than the default "dbo". This assumes that all tables will need to be in that schema, including the phinxlog table. Related to issue #991 .


rolfvreijdenberger:
referencing #930 fix for postgres schema functionality

PR #994, opened 2016-12-06
Add ability to use custom schema with SQL Server adapter

tm8747a:
This is to allow using a default schema other than dbo, similar to what is possible with the Postgres adapter. This fix assumes that all tables, including the phinxlog one, will be stored in the schema specified.


dereuromark:
There was never any team feedback on this one. @ team We should at least comment on why this is not further processed.

PR #1004, opened 2017-01-03
[Migration][Refresh] add command migrate:refresh

VaderSV:
Simple refresh command

### DIFF
--- a/src/Phinx/Console/Command/MigrateRefresh.php
+++ b/src/Phinx/Console/Command/MigrateRefresh.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * Phinx
+ *
+ * (The MIT license)
+ * Copyright (c) 2015 Rob Morgan
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated * documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


VaderSV:
image
How fix this error :(

robmorgan:
@VaderSV probably just needs `master` merged into your branch

robmorgan:
@VaderSV is this essentially just re-running migrations again?

VaderSV:
yes. Like migrate:refresh in laravel [migrations](https://laravel.com/docs/5.4/migrations)

robmorgan:
@VaderSV ok i'll wait to see if theres any further interest from the community before accepting this one.

VaderSV:
I take this task https://github.com/robmorgan/phinx/issues/982 for branch.

rquadling:
A couple of concerns. Not a complete list by any stretch of the imagination.

What identifies a migration as being safe to be run a second time? Or more times?
If a rollback is required, do the subsequent invocations of the migration cause a separate rollbacks? Or is it only the original migration that gets rolled back? What identifies a migration as being safe to rollback more than once?

If migrations are repeatable, they need to be explicitly marked as such. See https://github.com/robmorgan/phinx/issues/1046. In addition, controlling when/if repeatable migrations can/should be run is also important.

Whilst a sole developer may easily manage all the work themselves, in a multi-developer environment, migrations are mainly developed for a particular branch and are expected to be run all forward and backward as often as required for testing and deployment. Few, if any, migrations are developed with the necessary guards to either allow or to protect against a subsequent migration without a rollback in between.

Further, whilst this is an opt-in in terms of using the refresh command, I feel it provides a false sense that migrations will be fine to run twice or more times.

rquadling:
Normally the command name is the same as the class name.

Please update documentation and provide tests.

I would STRONGLY recommend that you disallow the `refresh` command for `version_order` being `creation`. This would break for multi-branch development at different release times. The initial execution order would/could be on a per branch order.

Branch A released, branch A's migrations executed.
Branch B released, branch B's migrations executed.

But if A and B were developed in parallel, the reverse creation order (the default value for Phinx prior to 0.8.0 and unless changed by the local dev team, the current default for 0.8.0), the order of rollback and subsequent re-migration may not be the same.

The potential of "Phinx refresh broke my database" is high. But, simply disallowing the option that could break is far saner.

I will move the default version_order to be execution, not creation in a subsequent release as we needed to introduce this feature and get people used to the benefit first. The next release will have a warning saying that the order will be changed in the next version.

dereuromark:
How should we proceed with this enhancement?

rquadling:
From my own perspective, this functionality is simple as 1 alias of `pxr && pxm`.

I'm not sure of the value of it for other developers. I can see even fewer usages of this feature on a production database. By the time you have a CI/CD pipeline, rollbacks are basically never going to happen unless you are in a dire emergency.

I think dropping the ability to rollback in reverse creation order would be the first step. Rolling back in a different order to execution is just going to mess with people's data in very very unknown ways.

But other than that, the code needs tests. Take into account various options of targets/dates make sure that the rollbacks are in reverse execution order.

dereuromark:
> yes. Like migrate:refresh in laravel migrations

I think this is now called "migrate:fresh" there

Refs https://github.com/cakephp/migrations/issues/246

othercorey:
Closing due to no activity.

PR #1005, opened 2017-01-04
Using `--verbose` to display raw queries.

dinamic:
This was requested in https://github.com/robmorgan/phinx/issues/93 and seems like a necessity when something goes wrong with the migrations. I am rather surprised it has not been implemented yet.

This functionality gives you the ability to troubleshoot custom queries using either `execute()` or `query()`. The currently implementation works only for PdoAdapter, but seems like all the other adapters are based on top of it. Should the solution get approved, I will revisit that statement and verify I am not missing any.

I am using the solution suggested by @tomzx, props to him for suggesting it!


dereuromark:
Thank you for you valuable input and improvement here.

If you resolve the conflict this looks it can be merged right away.
Not sure why this was stalled for so many months - but we should fix it now.

lorenzo:
Can we have this as an adaptor wrapper instead? Like the timer wrapper

ravage84:
How about adding an additional parameter ``--queries`` for that specific need?
When we add too much stuff to the verbosity flag, we might need a ``-vv`` (verbose verbosity) flag one day 😼

abekroenem:
This is a great improvement to Phinx.

I do agree with @ravage84 , this modification may need an specific argument to be called, like `--queries` or more explicit `--show-sql`, `--show-queries`.

lorenzo:
This was implemented and already in the master branch

PR #1006, opened 2017-01-04
Fix update column with , in definition (SQLite)

fquffio:
When using SQLite adapter, if a column is created and contains one or more `,` in its definition (e.g. in its default value, or comment), that column cannot be updated anymore, as the regex that changes the corresponding SQL fragment fails to properly detect the end.

This PR _should_ solve the problem, along with adding a regression test.


fquffio:
Hey @robmorgan, sorry to bother.

Have you had the time to take a look at this? Is there any additional information that I should provide in order for this bugfix to be examined?
I hate to stress out people, but this bug is something that's actually breaking things up for us, and despite using branch aliases in Composer does indeed work, we'd like to see this just as a temporary solution. 🙃 

Thanks aforehand for your attention!

robmorgan:
@fquffio thanks for your contribution, sorry about the delay.

PR #1029, opened 2017-01-30
Added vendor to phinx path to be aligned with other docs sections

derpoho:


robmorgan:
thanks

shadowhand:
This is not in agreement with [`bin-dir` configuration](https://github.com/robmorgan/phinx/blob/2ad5c22c9c41134b1ad12b8462542269a515e104/composer.json#L42).

rquadling:
The content of vendor/bin are shell scripts.

I think the confusion is coming from developing phinx - in which case the bin directory contains the PHP script - and the composer bin/shell/wrapper - in which case this is only shell scripts to the scripts in the projects bin directory.

So, the documentation should be just `vendor/bin/phinx` and not `php vendor/bin/phinx`.

The bin-dir configuration is about what files in the Phinx project need shell scripts generated by composer and thereby accessible via `vendor/bin`.

If someone is using Phinx *outside* of composer, then `bin/phinx` is what the documentation should say. Otherwise `vendor/bin/phinx`.

Of course, those who are running on Windows will be using `\`, rather than `/`.

In both cases, there is no need to put the `php` interpreter in as the file is shell executable.

On Unix, via composer, `vendor/bin/phinx` is an executable shell script.
On Unix, not via composer `bin/phinx` is an executable shell script that uses the PHP interpreter to run - assuming it is installed appropriately.
On Windows, via composer, `vendor\bin\phinx.bat` is a batch file.
On Windows, not via composer `bin\phinx.bat` is a batch file.

derpoho:
This commit was only synchronising the states, it was referenced in multiple different paths. That's solved now. I didn't really think about the correct way to reference it on windows or any other variants.

rquadling:
`bin/phinx` or `vendor/bin/phinx` (`bin\phinx` or `vendor\bin\phinx` for Windows users).

But never `php bin/phinx` or `php vendor/bin/phinx`. The `php` is wrong.

PR #1030, opened 2017-02-02
index type and partial index

mvrhov:
* added support for partial indexes via WHERE keyword, postgresql, Sqlserver and sqlite
* added support for index type (btree, hash, gin,..) via USING keyword, postgresql, mysql feature

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -1150,6 +1150,46 @@ ensure the table uses the ``MyISAM`` engine.
             }
         }
 
+The MySQL and Postgresql adapters also supports ``using`` keyword when creating indexes.
+This defines the method used to create index . The method is database specific, please consult the manual.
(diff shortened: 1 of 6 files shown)


mvrhov:
Added docs and partial's support for all DBs that support them

mvrhov:
What's missing to get this merged?

rquadling:
Principle looks good. It would be nice to have some tests.
Escaping/guarding of invalid values may be a good idea.

lorenzo:
Having tests would be nice!

mvrhov:
* Written my own migration library in this time [migratum](github.com/mvrhov/migratum)
* moving to golang

Somebody else should take over. Or this should be closed.

lorenzo:
Closing until this can come up again in a future version

PR #1035, opened 2017-02-09
Adding backwards compatibility to table insert

3amprogrammer:


robmorgan:
@3amprogrammer thanks, but I don't think many people are still using the older definition?

3amprogrammer:
Well, probably they don't but the existing migrations are not passing if you update to the latest version, at least it has just bite me. I have added tests also. Sorry for the commits spread, but that is my first contribution ever :)

robmorgan:
@3amprogrammer are not able to change the code in your project?

3amprogrammer:
@robmorgan I am, but I thought that would be helpful and save some time people like me.

robmorgan:
@3amprogrammer how many migrations are we talking about here?

3amprogrammer:
In my case it was around 30

rquadling:
I think a helper function to allow older usages to be coerced to the "correct" format would be a better option.  The change was 15 months ago. That's a long time (at least in software time). A one off conversion of calling the helper to wrap the older parameters should be relatively simple to achieve.

lorenzo:
Closing as phinx is still not marked as stable, and this is quite old already. Thanks for your contribution anyways!

PR #1042, opened 2017-02-23
change glob to adapter the .phar files

jhonata-menezes:
When generating .phar file the glob function did not find the migrations, changing to DirectiryIterator it is possible to find the migrations.

### DIFF
--- a/src/Phinx/Util/Util.php
+++ b/src/Phinx/Util/Util.php
@@ -213,6 +213,13 @@ class Util
      */
     public static function glob($path)
     {
-        return glob($path, defined('GLOB_BRACE') ? GLOB_BRACE : 0);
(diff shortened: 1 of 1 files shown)


rquadling:
As you can see, this change breaks our unit tests. You will need to fix those. I am unsure why you would embed migrations and seeds into the phar file as that would mean you couldn't add more migrations. If you can explain better what you are attempting to achieve by packaging your migrations, then maybe I can get a better understanding and see what is needed.

dereuromark:
@jhonata-menezes Any update here regarding the tests?
Then we could merge.

jhonata-menezes:
I'll do the tests this week.

lorenzo:
Closing as this seems abandoned.

PR #1050, opened 2017-03-01
Update install.rst

murilozilli:
In  the default ``phinx.yml`` it sets up the migration folder like this:
migrations: %%PHINX_CONFIG_DIR%%/db/migrations

Having the documentation stating that the user should only create one folder called ``migrations`` on project directory might mislead the user


robmorgan:
@murilozilli thanks

PR #1053, opened 2017-03-03
Docker Development Kickstart

Matt3o12:
When I was trying to contribute to phinx, I found it rather difficult to setup the testing environment for running the unit tests.

This is why I have decided to create [docker](https://www.docker.com) and [docker-compose](https://docs.docker.com/compose/) files to help you kickstart a development environment. 

The advantage of using docker for developing is that running `docker-compose run phinx` will bring up a freshly created MySQL 5.5 and Postgres 9.2 database (the versions can be easily adjusted in `docker-compose.yml`), and a PHP 5.4 container running the unit tests. The PHP 5.4 container uses CentOS 7. I chose CentOS 7 since it still supports (and maintains PHP 5.4, unlike the official docker image) and I chose PHP 5.4 because it is the oldest PHP version supported by phinx.

So, what this does is: it will download & install all depencies (mysql, postgres, sqlite-lib, php, git and a few others), download all php composer packages (using --prefer-stable --prefer-source --prefer-lowest), and then run the tests, just with one command.

Besides being able to easily test it against every database, the database containers are volatile which means once they are removed (docker-compose down), they loose all data (this is helpful if you manage to corrupt the database because recreating it is really easy). They also isolate your other database (security warning: docker does not provide secure isolation to the host os, which means that there might be zero-days, so do not run any untrusted code inside docker. It will, however, save you from an accidental  `DROP DATABASE myOtherDatabase`). 

## Things to do (can be done in another merge): ##

- [ ] **Create an image for the Dockerfile** (this will decrease the start time for the first start dramatically since there is no "build" step -- this should only be done by the maintainers but creating the image can be fully automated & I am happy to assist)
- [ ] **Add support for Microsoft SQL** (since Microsoft added Linux support, it should be possible to run MSSQL tests on Mac, Linux, and Windows from Docker. Furthermore the linux build should run identically to the Windows build -- if it does not, it is considered a bug by Microsoft).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,18 @@
+FROM centos:7
+
+RUN yum install -y php php-cli php-pdo php-mbstring php-intl php-pdo \
+	php-pgsql php-mysqlnd php-xml git && \
+	curl -sS https://getcomposer.org/installer | php && \
+	mv composer.phar /usr/local/bin/composer
+
+COPY docker/entrypoint.bash /usr/local/bin/entrypoint.bash
+COPY composer.json /vendor_preinstalled/
+
+RUN cd /vendor_preinstalled/ && \
+	composer update --no-interaction --prefer-source --prefer-stable \
+		--prefer-lowest
+
+WORKDIR /app/
+ENTRYPOINT ["/bin/bash", "/usr/local/bin/entrypoint.bash"];
+CMD ["phpunit"]
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,20 @@ Great, so you want to contribute. Let's get started:
 
 1. Open a pull request on the Github project page. Ensure the code is being merged into `master`.
 
+## Getting Started with docker
+
+If you do not wish to install MySQL and Postgres locally to run unit tests
+can use [docker-compose](https://docs.docker.com/compose/) which will start 
+both database, install all development dependencies and run all unit tests.
+
+To get started, just run `docker-compose run phinx`. It will download all 
+images, install all dependencies, hook them up and run all unit tests
+(note: SQLServer tests are not supported, yet).
+
+You can also use `docker-compose run phinx install` to update/reinstall
+php composer packages manually. If you wish to drop into a shell, just run
+`docker-compose run phinx bash`, which will drop you right into a shell.
+
 ## Documentation
 
 The Phinx documentation is stored in the **docs** directory using the [RestructedText](http://docutils.sourceforge.net/rst.html)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "2"
+services:
+    phinx:
+        build:
+            dockerfile: docker/Dockerfile
+            context: .
+        volumes:
+            - .:/app
+        links:
+            - mysql:mysql
+            - postgres:postgres
+    mysql:
+        image: mysql:5.5
+        environment:
+            - MYSQL_DATABASE=phinx_testing
+            - MYSQL_ALLOW_EMPTY_PASSWORD=yes
+    postgres:
+        image: postgres:9.2
+        environment:
+            - POSTGRES_DB=phinx_testing
+
--- a/docker/entrypoint.bash
+++ b/docker/entrypoint.bash
@@ -1,0 +1,76 @@
+#!/bin/bash
+set -e
+
+VENDOR_DIR=/app/vendor/
+
+function prepareConfig {
+	cat phpunit.xml.dist |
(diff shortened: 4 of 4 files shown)


westonplatter:
@Matt3o12 I'd like to help out with this issue. How can I help?

Matt3o12:
@westonplatter Well, first of all, you could tell me whether your not, you are interested in this pull request. I found this docker file incredible helpful for setting up a test environment for the development of this project (because I did not have to install any dependency on my computer). I also believe other might find this helpful because settings up the test environment is incredible easy using docker if you are familiar with it.

Secondly, you can create a repository "robmorgan/phinx" on docker hub to host the image. This would speed up the first time the tests are run under docker. I can assist you with creating an automated build process so that the image is always up-to-date and you there is no maintenance. Please only do this if you are at least a maintainer of this project so this image can be official and trusted. 

Lastly, if this is merged I could look into ways to also set up the docker-developemtn environment (this is this pull requests) for Microsoft SQL server tests (this would work on Linux, macOS, and Windows without having to install windows). This is thanks to the now official Linux support for Microsoft SQL Server. But I cannot guarantee if I am even able to do that because I am not familiar with MSSQL (though, I will try my best if there is interest and I have some spare time).

Please keep in mind that this will not replace the automated Travis tests and they will always be more accurate because the project would be tested against all Php versions. Using my method, tests are only run for PHP 5.4 (because it is the oldest -- read my official pull requests for more information). This is rather a more convenient method for getting started with contributing to phinx.

docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM centos:7
```
newPOPE: Why you did not use official PHP image?

docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+COPY docker/entrypoint.bash /usr/local/bin/entrypoint.bash
+COPY composer.json /vendor_preinstalled/
+
+RUN cd /vendor_preinstalled/ && \
+	composer update --no-interaction --prefer-source --prefer-stable \
+		--prefer-lowest
+
+WORKDIR /app/
+ENTRYPOINT ["/bin/bash", "/usr/local/bin/entrypoint.bash"];
```
newPOPE: Is really necessary use that complex entrypoint? From my point of view (after different project migrated to Docker development) is better switch devs directly to the command line when common commands are present.

Something like `docker-compose run phinx bash` and there is prepared tools like `composer` and others.
On the other side you can use [composer scripts](https://getcomposer.org/doc/articles/scripts.md) for that commands.

docker/Dockerfile
lorenzo: @newPOPE If I merge this pull request would you volunteer to send a new PR changing it to use the official image?

newPOPE: With a pleasure

lorenzo: @newPOPE great, feel free to improve anything after this is merged. I also agree the entrypoint is too complex

lorenzo:
Thanks a lot @Matt3o12

lorenzo:
@newPOPE Now it's your turn :)

Matt3o12:
@newPOPE @lorenzo 

Sorry for taking so long to answer. Here are my reasons why I used centos instead of the official PHP image: the official image no longer maintains a PHP 5.4. In fact, PHP 5.6 is their oldest maintained image  means that the image for the old systems are insecure and might put developers machines at harm (though it would take quite a lot of work to exploit that). Furthermore they might delete the tag later which might break the dockerfile or a security update changes PHP's API and breaks phinx (unlikely but could happen). Since I would expect contributes to also work for the oldest supported PHP version, i chose centos since they are the only "modern" distro which still supports such an ancient PHP version (and they also support their distro for a very long time, so things should not break for a while). 

Secondly, I chose such a complicated entry point because of composer. When having a slow internet connection, composer takes forever to build the vendor dir. Ideally, we would publish the docker image, which already contains the vendor dir. Pulling a docker image with the vendor dir prepacked is so much quicker then using composer install (at least for me). We can setup docker hub to automatically build a new image with the updated dependencies for push to the main branch.  My entry point just copies vendor dir if it doesn’t exist already. Furthermore, a different configuration file is automatically used, which is configured for Postgres, MySQL and SQLite. This is necessary because the database runs in another container.

newPOPE:
@Matt3o12 thank you for clarifying. You can see PR #1139 for simplifying Docker dev environment. 

1. your selection of centos sounds reasonable. My advice for this situation is switch docker image to 7.*. Yes I know that phinx supports 5.4+ but that version and later versions of 5.* branch is unsupported.

2. yes, composer's cacche is a problem but solution is map cache folder to host via docker-compose file.

We can move discussion about Docker to that issue.

Issue #960, opened 2016-09-21
Text Wrapper crashes when calling rollback twice


moritz-h:
When i try to call getRollback twice on a text wrapper i get a `Symfony\Component\Console\Exception\RuntimeException: Unable to write output.` Exception.

Example:

``` php
$app = new \Phinx\Console\PhinxApplication();
$wrap = new \Phinx\Wrapper\TextWrapper($app);
$wrap->setOption('environment', 'db');
$wrap->setOption('configuration', 'phinx.php');
$wrap->setOption('parser', 'php');

echo $wrap->getRollback(null, 0);
echo $wrap->getMigrate();
echo $wrap->getRollback(null, 0); // <- exception here
```

When i setup a new wrapper in between it works:

``` php
$app = new \Phinx\Console\PhinxApplication();
$wrap = new \Phinx\Wrapper\TextWrapper($app);
$wrap->setOption('environment', 'db');
$wrap->setOption('configuration', 'phinx.php');
$wrap->setOption('parser', 'php');

echo $wrap->getRollback(null, 0);
echo $wrap->getMigrate();

$app = new \Phinx\Console\PhinxApplication();
$wrap = new \Phinx\Wrapper\TextWrapper($app);
$wrap->setOption('environment', 'db');
$wrap->setOption('configuration', 'phinx.php');
$wrap->setOption('parser', 'php');

echo $wrap->getRollback(null, 0); // <- no exception here
```

Edit: Version is 0.6.4

robmorgan:
hi @moritz-h, what are you putting in your `phinx.php` file?

moritz-h:
@robmorgan I created a minimal example: https://gist.github.com/moritz-h/579f43c8af02f3644115d91185b73f6e
(The `20160922000000_init_db.php ` file should be in a `migrations` subdirectory, could not name it with path in the gist.)

I run `composer install` and start it with `php -S localhost:8000 index.php`.

The database settings in the phinx.php match to an existing mysql database and my php version is `7.0.8` (default Ubuntu 16.04).

moritz-h:
@robmorgan I found out where this error comes from:

The TextWrapper creates a new Output Instance for each call:
https://github.com/robmorgan/phinx/blob/master/src/Phinx/Wrapper/TextWrapper.php#L215

This OutputInterface Object is recieved by execute() of a command (i.e. Status) and then passed to bootstrap()
[...]

PR #1056, opened 2017-03-14
Update Input- and OutputInterface on cached Manager

moritz-h:
Fix: #960 
The commands get a fresh InputInterface and OutputInterface Instance for each time, the command runs. But the commands cache the Manager instance, which gets a reference to the Input/OutputInterface on construction. So when the command runs a second time, they get fresh Input/OutputInterfaces but uses the cached Manager Instance with reference to old Input/OutputInterface Instances.
This PR updates the Input and OutputInterface of the cached Manager to the Input/OutputInterface of the current command run. (See #960 comment for more details)

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -285,6 +285,10 @@ abstract class AbstractCommand extends Command
         if (null === $this->getManager()) {
             $manager = new Manager($this->getConfig(), $input, $output);
             $this->setManager($manager);
+        } else {
+            $manager = $this->getManager();
+            $manager->setInput($input);
+            $manager->setOutput($output);
         }
     }
 


rquadling:
That makes sense to me. Can you write a test that proves getting the manager gets the injected Input/Output.

moritz-h:
I thougth of writing a test for this and read througth your tests. I wanted to start here:
https://github.com/robmorgan/phinx/blob/master/tests/Phinx/Console/Command/StatusTest.php

I saw you are using a manager stub in the tests and my idea was to test somehow if the stub gets the new input/output Interfaces on command execution. But i don't find any code where you are testing the bootstrap part of a command. So i had no startpoint where i could add these tests.
If you can give me a hint where i could add these tests, i will try to write some.

piotr-cz:
Thanks, this fixed the issue

PR #1060, opened 2017-03-18
Fix rollback on only one migration

### LINKED ISSUE #1059: Rollback one migration fails
Do `phinx rollback` with only one migration fails: `No migrations to rollback`

`phinx rollback -t 0`still works correct for one migration. Also error is the same if I start a new project with only one migration, and try to roll this back or if I have more migration and rolling them back one by one.
Error is in v0.8.0, is working with v0.7.2.

moritz-h:
Fix #1059

I think error comes with this commit: https://github.com/robmorgan/phinx/commit/bba44b381fef9756c1d2f5a6c65ddc6757e7f6bb
If only only migration is left the first migration is selected as target instead of 0. So nothing is done.

I found a test case for this, but don't know why the test does not work:
https://github.com/robmorgan/phinx/blob/master/tests/Phinx/Migration/ManagerTest.php#L712

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -455,7 +455,7 @@ class Manager
         if (null === $target) {
             // Get the migration before the last run migration
             $prev = count($executedVersionCreationTimes) - 2;
-            $target = $prev >= 0 ? $executedVersionCreationTimes[$prev] : $executedVersionCreationTimes[0];
+            $target = $prev >= 0 ? $executedVersionCreationTimes[$prev] : 0;
         }
 
         // If the target must match a version, check the target version exists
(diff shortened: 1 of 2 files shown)


rquadling:
A negative assertion, only tests what shouldn't happen, not what does happen. Hmmm.

moritz-h:
I played with the assertions. So ether add to assert not contains `No migrations to rollback`
or to contains some of this:
```
== 20120111235330 TestMigration: reverting
 == 20120111235330 TestMigration: reverted 0.0000s
```

rquadling:
Hi.

If you can make the test assert positively (i.e. that when there is 1 or 0 migrations, the result is appropriate, then I can merge this.

moritz-h:
I added both now. Positive assertion for the "reverting" and "reverted" status output.
Also the negative assertion to not contain "no migrations", so for what ever could go wrong, the test could fail on this.

rquadling:
Thank you.

glensc:
@moritz-h (since you are frequent contributor): do not link to branches, use git tag (or commit hash) to link to sourcecode, because the line (or even file) referenced gets out of sync, so the links in issue make no sense in the future.

moritz-h:
@glensc ok, I try to do this in future.

Issue #338, opened 2014-10-28
Multiple write databases


BardiaAfshin:
In a multiple read write database environment, If you have to write databases that the migrations need to get applied to, it'd be nice to have a nested phinx yaml configuration that apply the migrations to more the one database when you kick off the migration. That is, take in the credentials for all write databases and connects and runs the valid migrations one by one.

Upon error, I'd see the migration doing a down and recovering with the proper algorithm that ensures application integrity and uptime. This will require further development as the community provides input on their requirements.

keithy:
The library http://github.com/keithy/primo-pdo-php has this basic tools for this feature.

In this example, the migration is applied to all environments. (choices)
It is applied to a configuration which is derived from the base environment, but adds a prefix to the database name.
```
$reader = new Primo\Phinx\ConfigReader( __DIR__ . "/_fixtures/phinx.php");

$clean = true; 
foreach( $clients as $client ) 
{
  foreach( $reader->choices() as $choice )
  {
    $reader->choose($choice)->which('per-client')->atPut('%%CLIENT%%',  "$choice_$client")->create(true);
  }
}
```

```
'sqlite' => [
    'which' => [
        'per-client' => [ 
          'name' => '%%CLIENT%%'
     ]
]
```

Issue #180, opened 2014-02-02
Multiple Databases


wpillar:
It would be cool if Phinx could handle migrations for different databases all from the same instance. In our company setup we have one application which uses n databases, with different migrations for each that need to be handled.

What are you thoughts on handling multiple databases? I might fork and see if I can get it working but I want to know if you'd ever merge something like that?

jamespsterling:
+1

BardiaAfshin:
The configuration you are proposing doesn't fit being able to drive multiple databases with different revisions. Since sharing the same PhinxLog, you'd be dependent on collapsing the sequencing migrations after another instead hence removing a major flexibility. Unless, that's not a requirement. Then, why not.

laurin1:
I know this is kind of a hijack, and apologies....

https://github.com/robmorgan/phinx/issues/515

Actually, having the phinxlog table as part of the same database is a problem for us as well and if we moved instead had a structure like this, maybe it would that part of this issue as well:

* schema: db1
* schema: db2
* schema: phinxlog
** table: db1 (phinxlog table for db1)
** table: db2 (phinxlog table for db2)

temperedvision:
Phinx already has the ability to run migrations across different databases without any kind of hack (although the process to set up and use is a little manual/ugly). The solution is simply to use a different config file for each database, and set the migration path to a different sub folder for each DB. Each DB will then have its own migration path and phinxlog table for migration history.

To reduce the amount of code duplication for the configs, I switched my example to use PHP config files. My example also uses $_SERVER vars, as the config in my examples comes from a capistrano integration, but you should get the gist.

file structure:
```
migrations/
| - db1/
| - db2/
| - (etc)
```
db1.php:
```php
$_SERVER['PHINX_DBNAME'] = 'castle';
// set other params here

return include 'phinx.php';
```
phinx.php
```php
return [
	'paths' => [
		'migrations' => 'migrations/' . (isset($_SERVER['PHINX_DBNAME']) ? $_SERVER['PHINX_DBNAME'] : ''),
	],
	'environments' => [
		'default_migration_table' => 'phinxlog',
		'default_database' => 'default',
		'default' => [
			'adapter' => 'mysql',
			'host' => isset($_SERVER['PHINX_DBHOST']) ? $_SERVER['PHINX_DBHOST'] : '',
			'name' => isset($_SERVER['PHINX_DBNAME']) ? $_SERVER['PHINX_DBNAME'] : '',
[...]

PR #1064, opened 2017-03-21
Change adapter in migration

t-moennich:
This PR implements a new `setAdapterByAlias($alias)` method in the `AbstractMigration` class. This allows to change the used adapter during the migration process.

To keep the adapter selection handy, I've also created a configuration extensions to provide a list of databases which may be choosen by an alias.

Specific test for the new configuration section and `setAdapterByAlias` method will be pushed asap.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -337,13 +337,13 @@ class Config implements ConfigInterface
     /**
      * {@inheritdoc}
      */
-    public function getDatabases()
+    public function getAvailableAdapters()
     {
         if (isset($this->values, $this->values['databases'])) {
             $adapter_factory = AdapterFactory::instance();
             $default_env = $this->getDefaultEnvironment();
             $default_database = $this->getEnvironment($default_env);
-            $databases = [];
+            $adapter_list = [];
             foreach ($this->values['databases'] as $key => $options) {
                 $options = array_merge($default_database, $options);
                 $adapter = $adapter_factory->getAdapter($options['adapter'], $options);
@@ -357,10 +357,10 @@ class Config implements ConfigInterface
                     $adapter = $adapter_factory->getWrapper('prefix', $adapter);
                 }
 
-                $databases[$key] = $adapter;
+                $adapter_list[$key] = $adapter;
             }
 
-            return $databases;
+            return $adapter_list;
         }
 
         return null;
@@ -369,12 +369,12 @@ class Config implements ConfigInterface
     /**
      * {@inheritdoc}
      */
-    public function getDatabase($name)
+    public function getAvailableAdapter($name)
     {
-        $databases = $this->getDatabases();
+        $adapters = $this->getAvailableAdapters();
 
-        if (isset($databases[$name])) {
-            return $databases[$name];
+        if (isset($adapters[$name])) {
+            return $adapters[$name];
         }
 
         return null;
@@ -383,9 +383,9 @@ class Config implements ConfigInterface
     /**
      * {@inheritdoc}
      */
-    public function hasDatabase($name)
+    public function hasAdapter($name)
     {
-        return (null !== $this->getDatabase($name));
+        return (null !== $this->getAvailableAdapter($name));
 
     }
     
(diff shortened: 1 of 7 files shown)


t-moennich:
Would be great to know if this PR has a future (as we need this in our company).
If so, I'll push detailed unit tests of course.

src/Phinx/Config/Config.php
```diff
@@ -333,6 +334,60 @@ public function isVersionOrderCreationTime()
...
+
+        return null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAvailableAdapter($name)
+    {
+        $adapters = $this->getAvailableAdapters();
```
rquadling: Would the list of adapters change during the lifetime of running phinx? If not, should $adapters be cached?

src/Phinx/Migration/Manager.php
```diff
@@ -675,7 +675,7 @@ public function getMigrations()
                     }
 
                     // instantiate it
-                    $migration = new $class($version, $this->getInput(), $this->getOutput());
+                    $migration = new $class($this, $version, $this->getInput(), $this->getOutput());
```
rquadling: OOI, as `$this` is being supplied, why bother supplying the `input` and `output`?

src/Phinx/Config/Config.php
t-moennich: In theory no, as they are persistent per configuration. So yes, caching seems to be an good idea!

rquadling:
What is the use case you have for changing the adapter from the configuration? (Purely for interest).

src/Phinx/Migration/Manager.php
t-moennich: Good point, will be changed

t-moennich:
@rquadling some of our migrations in a single project have to access multiple databases.
Therefore we have to change the adapter.

We tried also to change the `name` option in the `PdoAdapter` options but this would into uncontrollable migrations which could write into everything.
So we decided to provide a list of database connections and allow to refer on them by the alias.

rquadling:
Some thoughts.


Is this really choosing an environment (from Phinx's terminology)?

'Adapter' is the term related to the database driver layer, not a specific connection or credential set.

Whilst commonly, people have 'development', 'staging'/'test', 'production', each of those 'environments' can have a different adapter and credential set.

For you, I guess you have something like 'prod_1', 'prod_2', 'prod_3' and you are wanting to run the migrations against these environments (either all against all, or some against some).

Where does the phinxlog table exist? I would again guess that you have a central server (one you consider primary in some regard) and that's the default environment?

What I'm concerned with is calling this 'adapter' when that has no direct bearing on credentials or specific server. Just which driver to use.

Given that this looks to be an environment switch, you'd have to guarantee that the environment is switched back to the default one in case someone doesn't set it in the migration. So cloning the manager for a migration rather than passing it as a reference. This would mean each migration gets a clean copy of the manager to do what it wants with, but it isn't persisted to the next migration.

t-moennich:
> Is this really choosing an environment (from Phinx's terminology)?

Nope it isn't. It's realy the change of the database connection/the adapter.
We have some old projects there tables are logically split among different databases.
Therefore we had to switch them either in a second migration or also in the same migration multiple times.

> Where does the phinxlog table exist? I would again guess that you have a central server (one you consider primary in some regard) and that's the default environment?

The `phinxlog` is stored at the environment 'adapter' connection, which is given in the environment configuration itself.
But this is - our advantage - stored in `Manager\Environment` and dosen't change while changing the migration 'adapter'. :)

> 'Adapter' is the term related to the database driver layer, not a specific connection or credential set.
> What I'm concerned with is calling this 'adapter' when that has no direct bearing on credentials or specific server. Just which driver to use.

I'm with you! I've also thought about naming the required method `setConnectionByDatabaseAlias` rather than ~~setAdapterByAlias~~ but as the `AbstractMigration` already uses the `setAdapter` term I just adopted that.

But in the end WE just switch the database name. A full switch of the adapter brings more features but on the other side, another adapter for the same environment shouldn't be necessary at all..

dereuromark:
This should be re-done on current master - otherwise closed as outdated.

dereuromark:
Requires a new PR

refs 
- https://github.com/cakephp/phinx/pull/1241
- https://github.com/cakephp/phinx/issues/180
- https://github.com/cakephp/phinx/issues/338

Issue #539, opened 2015-04-23
Default value for UUID in Postgresql


srigi:
I want to be able to do this:

``` php
$this->table('users')
    ->addColumn('uuid', 'uuid', ['default' => 'uuid_generate_v4()'])
```

However column option gets quoted so it leads to error

```
[PDOException]
  SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for uuid: "uuid_generate_v4()"
```

Are you open to this request?

RQuadling:
http://stackoverflow.com/questions/22446478/extension-exists-but-uuid-generate-v4-fails maybe?

srigi:
@RQuadling that's not the issue. Phinx is quoting that column option so it literally creates table like

```sql
CREATE TABLE user (
  uuid uuid DEFAULT 'uuid_generate_v4()'
);
```

that function call is wraped in quotes which is wrong. I got extension enabled.

steffkes:
looks like i've overlooked this one while opening #625 - underlying problem is the very same, though this one is missing a solution or at last proposal for one. @srigi is this still a relevant request for you?

srigi:
SQL Literal is a way to go for me. If that gets implemented, this would become irrelevant.

karllhughes:
I know this is old, but what's the workaround? Can I write raw sql to make the id an auto-generated uuid?

daneren2005:
I updated this thinking it would have been fixed in the past two years, but it's still broken.  Is this not going to be fixed?

rquadling:
If someone can produce a PR with unit tests, then we will certainly consider it. I don't have any experience with Postgres to determine what would be a suitable option.

I would guess a simple string test to see if the supplied default is a predefined constant, then treat the default as is without quoting.

karllhughes:
@rquadling I made a PR for this. Let me know if there's anything else I can do to improve it. https://github.com/robmorgan/phinx/pull/1065

dereuromark:
PR open

tertek:
For anyone still coming here. It has been resolved by using Literals:

https://book.cakephp.org/phinx/0/en/migrations.html#custom-column-types-default-values

Issue #625, opened 2015-08-23
Introducing literals?


steffkes:
While using phinx more and more, i've had a problem using a function as `default` since it gets quoted all the time no matter what.

```
/**
 * Get the defintion for a `DEFAULT` statement.
 *
 * @param  mixed $default
 * @return string
 */
protected function getDefaultValueDefinition($default)
{
    if (is_string($default) && 'CURRENT_TIMESTAMP' !== $default) {
        $default = $this->getConnection()->quote($default);
    } elseif (is_bool($default)) {
        $default = $default ? 'TRUE' : 'FALSE';
    }
    return isset($default) ? 'DEFAULT ' . $default : '';
}
```

this snippet coming from `PostgresAdapter`, we do have a special check for `CURRENT_TIMESTAMP` which probably is an indicator that we already should have something like that in place :?

my idea would be something like:

```
$table->addColumn(
  'id',
  'uuid',
  [
    'default' => new WhateverTheNamespaceWillBe\Literal('gen_random_uuid()')
  ]
);
```

and extending `getDefaultValueDefinition` for a check on `Literal`, which - if true - would call `Literal#__toString` and use that w/o quoting. 

having used Doctrine ORM before, i'm pretty sure there are more places that could use Literals :)

all thoughts welcome! while waiting on #559 i've implemented that as additional check in my custom adapter, but would like to contribute that back.

mvrhov:
Why don't you provide a PR with the Literal implementation. I'm sure it will get merged.

steffkes:
@mvrhov i'm not using phinx anymore, at least not for the time being. i didn't come up with a PR straight away since i was not entirely sure that it would be something that would be helpful for others as well.

dereuromark:
@mvrhov Are you able to provide a patch with your suggested changes as PR?

mvrhov:
@dereuromark: Unfortunately no. I've switched away from phinx as I found it to limited.
I've decided to write my own migration tool which utilizes plain SQL with some templating sprinkled on top of everything to allow some things to change on runtime. 
With that I can define tables stored procedures etc with the DSL that database provides.

PR #1065, opened 2017-03-22
Allowing postgres function in default column values

karllhughes:
Addresses but #539.

I added a method to check whether the default column value is a postgres function (either `CURRENT_TIMESTAMP` or the regex that tests for function name.

I'll admit that my method of determining whether or not the Postgres function is valid is simple, but it works and doesn't break anything.

I have not used Phinx with MySQL or other db's, so I'm not sure if this problem extends to those.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -918,7 +918,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
      */
     protected function getDefaultValueDefinition($default)
     {
-        if (is_string($default) && 'CURRENT_TIMESTAMP' !== $default) {
+        if (is_string($default) && !$this->isDbFunctionName($default)) {
             $default = $this->getConnection()->quote($default);
         } elseif (is_bool($default)) {
             $default = $this->castToBool($default);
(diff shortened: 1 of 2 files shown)


mvrhov:
The recommendation in that post was to use literal e.g. The example is here #625

karllhughes:
@mvrhov that recommendation was a couple years old and nobody was working on implementing it. While I think literals is a fine solution, I wanted to make my PR a simple bug fix rather than a new feature.

With this PR, you can use any Postgres function (including `now()`, so I think it accomplishes the goal. Correct me if I missed something though. I'm not super familiar with this library and it's my first PR to Phinx. Thanks!

mvrhov:
Yeah, but having more and more of such features makes it harder to upgrade to literal and only literal.

karllhughes:
Fair enough. My take is more pragmatic: if there's a bug I'd like to fix it today rather than wait on the feature that is not even in development yet.

If Literals is close to completion or someone's working on it and just needs some help, let me know and I'll join in that conversation. I think it's a great idea!

mvrhov:
It's not. But it is as simple as your solution.

```php
class Literal
private $literal;
pufblic functin __contricpotr($literal){
{$this->literal=$literal}
}

public function getLiteral()
return $this->literal;
}
```
And then in the metod you just do if is_object && instanceof Literal { ....

karllhughes:
Awesome. @mvrhov if you will submit a PR for literals, I'll gladly merge it into this PR and make my updates work with that strategy. Sounds like you have a good feel for what we need.

Otherwise, I'll just leave this here for consideration as it still addresses the open issue. If the maintainers decide they don't want to introduce the literals feature, they can choose to merge this in to fix the bug, or if they go the literals route, they can close this.

Thanks for the feedback!

mvrhov:
With Literals your PR is obsolete.

dereuromark:
Any change of the alternative to get drawn up? Then we can close this one. The regex seems to have a potential of causing other side-effects.

PR #1067, opened 2017-03-23
Dependent seeders

antonienko:
**Update**
I have rebased the code against master, as asked in the [previous pull request](https://github.com/robmorgan/phinx/pull/790). Didn't have the time to check everything is working again as initially, but there are tests in place from the original pull request.

----

Created a way to execute Seeders in order of Dependency.

Now the SeedInterface provides a method getParentSeed. Seeder instances should set their parent property to the name of the Seeder they depend on.

The new object SeedFetcher will take care of returning the seeders in the proper order of execution.

Parent seeders can be dependent on other seeders too. Old seeders should still work without modification.

Proposal to solve [issue 737](https://github.com/robmorgan/phinx/issues/737)

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -28,11 +28,11 @@
  */
 namespace Phinx\Migration;
 
+use Phinx\Seed\SeedFetcher;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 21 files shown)


antonienko:
I've tried to solve the problems, but I run out of time, so anybody that wants to take over is welcome

lorenzo:
I'm closing this as there was no more interest in it. I like the idea, so if anyone has time to bring back this to life, it would be appreciated

Issue #1054, opened 2017-03-06
BUG: The same migration name can be created in multiple directories.


rquadling:
If using multiple migration paths, only the chosen path at migration creation time is verified for the class name being unique, not all the other migration paths.

As such, the duplication is only realised when all the migrations are interacted with, say via a `phinx status` call.

Possibly/probably also true for using multiple seed paths.

With a `phinx.yml` of 

```yml
paths:
    migrations:
      - %%PHINX_CONFIG_DIR%%/db/migrations/1-Tables
      - %%PHINX_CONFIG_DIR%%/db/migrations/2-Views
    seeds: %%PHINX_CONFIG_DIR%%/db/seeds
...
```

And the following steps.
```
$ bin/phinx create Number1
Phinx by Rob Morgan - https://phinx.org. 0.8.0

using config file ./phinx.yml
using config parser yaml
using migration paths 
 - /Users/richardquadling/dev/php/robmorgan/db/migrations/1-Tables
 - /Users/richardquadling/dev/php/robmorgan/db/migrations/2-Views
using seed paths 
Which migrations path would you like to use?
  [0] /Users/richardquadling/dev/php/robmorgan/db/migrations/1-Tables
  [1] /Users/richardquadling/dev/php/robmorgan/db/migrations/2-Views
 > 0
using migration base class Phinx\Migration\AbstractMigration
using default template
created /db/migrations/1-Tables/20170306121851_number1.php

$ bin/phinx create Number1
Phinx by Rob Morgan - https://phinx.org. 0.8.0

using config file ./phinx.yml
using config parser yaml
using migration paths 
 - /Users/richardquadling/dev/php/robmorgan/db/migrations/1-Tables
 - /Users/richardquadling/dev/php/robmorgan/db/migrations/2-Views
using seed paths 
Which migrations path would you like to use?
  [0] /Users/richardquadling/dev/php/robmorgan/db/migrations/1-Tables
  [1] /Users/richardquadling/dev/php/robmorgan/db/migrations/2-Views
 > 1
using migration base class Phinx\Migration\AbstractMigration
using default template
created /db/migrations/2-Views/20170306121857_number1.php

$ bin/phinx status
Phinx by Rob Morgan - https://phinx.org. 0.8.0

using config file ./phinx.yml
using config parser yaml
using migration paths 
 - /Users/richardquadling/dev/php/robmorgan/db/migrations/1-Tables
[...]

Issue #756, opened 2016-01-19
Namespace loading failed


valetzyzy:
When set namespace in migration file PHINX can't load file

``` php
<?php

namespace app\db\migrations;

use Phinx\Migration\AbstractMigration;

class CreateUserSettings extends AbstractMigration {}
```

Some Exception

``` php
  [InvalidArgumentException]                                                                                                              
  Could not find class "CreateUserSettings" in file "~/app/db/migrations/20160118105051_create_user_settings.php"

```

rquadling:
Migrations, at the moment, cannot be namespaced.

shadowhand:
See #168

tariquesani:
#1070 Pull request solves this issue, is mergeable and passes all tests
Duplicate of #168 ?

lorenzo:
closing as duplicate.

Issue #168, opened 2013-12-26
Move migrations under the Phinx\Migration namespace


robmorgan:
In order to stop polluting the global namespace we will move migration classes under the `Phinx\Migration` namespace. This also conforms to PSR-2.

evought:
If you go as @cyrusboadway  suggests and allow a namespace to be specified in the config, then that will allow each project to put the migrations in their own vendor namespace and still preserve existing projects using the global namespace. New projects could default to Phinx\Migration . It would then become a backwards-compatible change.

Basically:

* If the configuration value is defined, put/find migrations in the specified namespace.

* The the configuration value does not exist, use the global namespace (current, backwards-compatible behavior).

* When a new project is created (or transitions to phinx), have "phinx init" set the configuration value to Phinx\Migration in the generated phinx.yml. The user will have the opportunity to edit the value if desired.

t1gor:
That's exactly what I'm looking for. Any update on the issue?

arogachev:
I switched to Phinx from native framework migrations (I'm using Yii2) just to give it a try and see it in action. But this is for now is one of the biggest disadvantages for me. Any update on this?

rquadling:
Some random thoughts.

The namespace will have to be added as a PSR-4 surely?

``'Phinx\\Migration' => '%%PHINX_CONFIG_DIR%%/migrations'``

The namespace will also need to be injected into the template so that custom templates and custom template generators will have the option of using the default namespace or not.

wandersonwhcr:
@rquadling +1

josegonzalez:
Would be great to do the same for Seeds :)

garex:
Also namespaces should be different per path.

Currently we have single path and it's possible to use single namespace per this path in future.

But if we will use glob in path then we force to use same namespace everywhere.

May be implement class loading in another way? Let it parse migration and use class name from parsed file. I think https://github.com/nikic/PHP-Parser will do the job easily. And it will not be backward breaker.

glensc:
https://github.com/robmorgan/phinx/issues/168#issuecomment-69807486 makes most sense,
https://github.com/robmorgan/phinx/issues/168#issuecomment-242325250 the parser solution is way overkill, it's like using tank to kill a fly.

dereuromark:
Is this ticket still valid?
What is the plan of attack?

dereuromark:
Looks like this has been done.

glensc:
[...]

PR #1070, opened 2017-04-03
Supporting namespaces in configuration, Migrations and Seeds.

andrey-mokhov:
Please consider my pull request, in which I made some enhancement.
The pull request fixes following issues:

- #168 Move migrations under the Phinx\Migration namespace
- #756 Namespace loading failed
- #1054 BUG: The same migration name can be created in multiple directories. (if add namespace in configuration)

I implemented an idea of @rquadling ([view comment](https://github.com/robmorgan/phinx/issues/168#issuecomment-160646325)).
So, support of namespaces looks like this:
php like:
```php
return [
    'paths' => [
        'migrations' => [
            '/path/to/migration/without/namespace',
            'Foo' => '/path/to/migration/Foo',
        ],
        'seeds' => [
            '/path/to/seeds/without/namespace',
            'Baz' => '/path/to/seeds/Baz',
        ]
    ],
];
```
yaml like:
```yaml
paths:
    migrations:
      0: ./db/migrations
      Foo\Bar: ./db/FooBar
    seeds:
      0: ./db/seeds
      Foo\Bar: ./db/seeds/Foo/Bar
```

json like:
```json
{
    "paths": {
        "migrations": {
            "0": "./db/migrations",
            "Foo\\Bar": "./db/FooBar"
        }
    }
}

```
Migrations/Seeds contain namespace just at the beginning of the file.

Migrations/Seeds files without namespace are still working for backward compatibility.

For using namespaces in migrations/seeds make new folders and change configuration.

The solution doesn't fully compatible with PSR-4 (there are versions in file names).

I tryed to covered the code with the unit's tests as full as possible.


mvrhov:
I don't agree with namespaced migrations as this prevents me to have them in different modules.

andrey-mokhov:
@mvrhov, current solution doesn't prevent you use migrations in different modules. Example:
```php
return [
    'paths' => [
        'migrations' => [
            './module_first/migrations',
            './module_second/migrations',
            // etc
            './module_other/migrations',
        ],
        'seeds' => [
            // ...
        ]
    ],
];
```
If you need Namespace you can add an item with string key to configuration. Example:
```php
return [
    'paths' => [
        'migrations' => [
            './module_first/migrations',
            './module_second/migrations',
            // etc
            './module_other/migrations',
            'Foo\Bar' => './module_new/migrations',
        ],
        'seeds' => [
            // ...
        ]
    ],
];
```
Will work both new and old (without namespace) migrations.

mvrhov:
Oh. Sorry. I've not reviewed the code. I've just read the PR description. If the old way works then I have no objections.

src/Phinx/Console/Command/Create.php
```diff
@@ -274,10 +279,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         // inject the class names appropriate to this migration
         $classes = array(
-            '$useClassName'  => $this->getConfig()->getMigrationBaseClassName(false),
-            '$className'     => $className,
-            '$version'       => Util::getVersionFromFileName($fileName),
-            '$baseClassName' => $this->getConfig()->getMigrationBaseClassName(true),
+            '$defineNamespace' => null !== $namespace ? ('namespace ' . $namespace . ';') : '',
```
rquadling: Supply `$namespace` separate to `$defineNamespace` please.

src/Phinx/Console/Command/SeedCreate.php
```diff
@@ -173,10 +174,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         // inject the class names appropriate to this seeder
         $contents = file_get_contents($this->getSeedTemplateFilename());
+
+        $config = $this->getConfig();
+        $namespace = $config instanceof NamespaceAwareInterface ? $config->getSeedNamespaceByPath($path) : null;
         $classes = array(
-            '$useClassName'  => 'Phinx\Seed\AbstractSeed',
-            '$className'     => $className,
-            '$baseClassName' => 'AbstractSeed',
+            '$defineNamespace' => null !== $namespace ? ('namespace ' . $namespace . ';') : '',
```
rquadling: Supply `$namespace` separate to `$defineNamespace` please.

src/Phinx/Console/Command/Create.php
andrey-mokhov: I'm not sure I understand your proposals.

$namespace variable can contain string like `Foo\Bar`

$defineNamespace is a result of concatenation: `namespace Foo\Bar;`

When namespace for path is defined in configuration, then result string injected in template (see ./src/Phinx/Migration/Migration.template.php.dist, ./src/Phinx/Seed/Seed.template.php.dist).

If namespace for path not found, then namespace definition in template is absent ($defineNamespace var is empty).

Please clarify your sentence.

Do you mean to create separate template for migrations with namespace? I suppose it would be more convenient to leave a single template.

Or maybe you suggest rename $defineNamespace var to $namespaceDefinition?

src/Phinx/Console/Command/SeedCreate.php
rquadling: Simply add ` '$namespace' => $namespace,`

Basically, I would need to see JUST the namespace in my template. This is independent to the `$defineNamespace` value.

rquadling: And `$namespaceDefinition` is a nicer name, so `$namespace` and `$namespaceDefinition`. I'd have use for both of these in my templates.

andrey-mokhov:
Renamed `$defineNamespace` to `$namespaceDefinition`, add `$namespace` to replace part.

andrey-mokhov:
@rquadling Do you think, i can add $namespaceBackslash?

rquadling:
> @rquadling Do you think, i can add $namespaceBackslash?

What use case?

andrey-mokhov:
hm, i refuce the idea (namespaceBackslash)

andrey-mokhov:
Hello! Could you please give a cue about this PR. In case of any doubts, I will be happy to offer any assistance.

rquadling:
Just need to review it and play with it a bit.

JapSeyz:
@rquadling Any news on whether or not this is a suitable implementation? It looks solid and good to me.

lorenzo:
@JapSeyz it only needs docs. Would you like to contribute them?

JapSeyz:
@lorenzo Yeah sure, I'll take a look at it next weekend when I have some time. Where should I put the PR?

lorenzo:
This looks like it requires configuration in the yaml file. That should be explained, with an example of a namespaced migration file.

You can open a pull request to the docs folder in this repository.

thanks for the help!

rquadling:
It has to work with existing migrations that won't be namespaced. If you want to start using namespaces, what's the consequences for existing migrations? Having some documentation for migrating the migrations (that's a bit self referential!) would be useful. That could be in a suitable UPGRADE file. This is assuming that namespaced and non-namespaced migrations cannot be used simultaneously.

JapSeyz:
@rquadling Alright, I'll check it out. I'll do a bit of testing and write down my findings a long with documentation for how to use the namespaced migrations.

JapSeyz:
@rquadling Is it possible to add onto other peoples PRs? I'm not too experienced in Git, but while writing the documentation I found a few things that could benefit the implementation.

It doesn't seem as if it is possible to have the same classname in different namespaces because `UTIL::isUniqueMigrationClassName()` only checks for the classname without paying attention to whether or not there's a namespace.

The ability to write the namespace in the Phinx console command, so it wouldn't first be placed in the global folder, and then the user would manually have to move and namespace it on his/her own.

JapSeyz:
@lorenzo I have created a draft here: https://github.com/JapSeyz/phinx/blob/master/docs/namespaces.rst
Is that explanative enough, or does it need something more?

In truth, it's a very easy thing to switch to namespaces. The above comment does however highlight a potential trap for users who want's to use namespaced and non-namespaced migrations/seeders in the same project. This has been written into the documentation though.

I have not yet tested this, but I presume you shouldn't escape the backslash in the console command?
`$ phinx create Foo\MyNewMigration`

andrey-mokhov:
Creating migration with command `$ phinx create Foo\MyNewMigration` must return an error. `create` command accepts argument `--path` for defining valid namespace for migration as path.

In case if configuration file defines several paths to directories where migrations stored, e.g.:
```yaml
paths:
    migrations:
      0: ./db/migrations
      Foo\Bar: ./db/FooBar
```
command `$ phinx create MyNewMigration` cause script to ask user for valid path for migration to store. If user chose first option (no namespaces required) - command will create migration file in default location, otherwise migration file will be created with namespace in specified location.

Calling `$ phinx create MyNewMigration --path ./db/FooBar` would lead to creation of migration, without additional prompt, with namespace.

lorenzo:
@JapSeyz I think the document is good, but misses some examples. When you define the paths for migrations, can you show an example of a file with its path that would be resolved by each of the configuration entries?

JapSeyz:
@andrey-mokhov Perfect. The namespace thing did bother me. I'll reflect this in the documentation.

@lorenzo I'll add some more examples.

JapSeyz:
I have added more examples and reflected the --path flag.

Let me know if it's acceptable, as I've found it really difficult to give examples of the path resolving as it's basically just saying what the different parts of the string mean.

lorenzo:
@JapSeyz It looks great! I'll be merging this soon

JapSeyz:
@lorenzo Sounds great! I've created a PR for the documentation so it's simply to merge it. Thanks for the quick responses

lorenzo:
Thanks!

johnrobertporter:
Is this going any further seen as the docs were merged?

I really need this as I have a lot of Seeds for different testing scenarios, and namespaces would organise them somewhat.

PR #1071, opened 2017-04-04
Postgresql can't have limit on bigint, same as integer and smallint

atmacola:
Have I implemented my feature for as many database adapters as possible? No, it is to correct a bug on postgresql
    Does my new feature improve Phinx's performance or keep it consistent? No change
    Does my feature fit within the database migration space? ???
    Is the code entirely my own and free from any commercial licensing? Yes
    Am I happy to release my code under the MIT license? Yes
    Is my code formatted using the PSR-2 coding standard? Yes


lorenzo:
Thanks!

PR #1085, opened 2017-05-02
allow symfony dotenv external parameters style in config

iampersistent:
http://symfony.com/doc/current/configuration/external_parameters.html

This PR leaves the current `%%PROOPH_VARIABLE%%` syntax in place but also allows for `%env(NEW_VARIABLE)%`

### DIFF
--- a/tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
+++ b/tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
@@ -24,9 +24,9 @@ class ConfigDefaultEnvironmentTest extends AbstractConfigTest
     {
         $_SERVER['PHINX_DBHOST'] = 'localhost';
         $_SERVER['PHINX_DBNAME'] = 'productionapp';
-        $_SERVER['PHINX_DBUSER'] = 'root';
-        $_SERVER['PHINX_DBPASS'] = 'ds6xhj1';
         $_SERVER['PHINX_DBPORT'] = '1234';
+        $_SERVER['DB_PASS'] = 'ds6xhj1';
+        $_SERVER['DB_USER'] = 'root';
         $path = __DIR__ . '/_files';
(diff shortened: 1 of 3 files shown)


kaoz70:
This would be great if it was merged!

tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
```diff
@@ -24,9 +24,9 @@ public function testConfigReplacesTokensWithEnvVariables()
     {
         $_SERVER['PHINX_DBHOST'] = 'localhost';
         $_SERVER['PHINX_DBNAME'] = 'productionapp';
-        $_SERVER['PHINX_DBUSER'] = 'root';
-        $_SERVER['PHINX_DBPASS'] = 'ds6xhj1';
         $_SERVER['PHINX_DBPORT'] = '1234';
+        $_SERVER['DB_PASS'] = 'ds6xhj1';
+        $_SERVER['DB_USER'] = 'root';
```
lorenzo: Why change this?

iampersistent: just to have a clear distinction between the existing way of handling the environment variables and using `env()`

lorenzo: I don't get it. So it wouldn't be possible to configure the port with env vars?

lorenzo: I think a cleared test would be to duplicate this test and make all parameters use env()

dereuromark:
So this is not needed anymore with https://github.com/cakephp/phinx/pull/1236 ?

saeideng:
>So this is not needed anymore with #1236 ?

@dereuromark  I think so 
you can close it IMO

Issue #566, opened 2015-06-10
Possibility of UnNamed Migrations?


csnuknet:
Given create currently requires a camel case string, would it be possible if no string is provided to create a migration class such that a defined string (config) or fixed sting is used? 

The result would being:
 <datetime>_Version with the class name being Version<DateTime> 

Alternatively possibly a config option to enable the additon  he date time to the end of any string provided for the class name. 

The idea being to avoid classes with the same names if the same string where provided by differing devs.

dereuromark:
In general, not having to use creative names can be useful for some. Letting them get auto-generated for example could work.
Are you able to provide a pull request with your suggested changes?

othercorey:
Closing due to lack of activity or out of scope. If there is new interest, please open a PR.

PR #1087, opened 2017-05-08
Remove name requirement and create classes same as filename

cleentfaar:
see https://github.com/robmorgan/phinx/issues/566

Proposed changes in this PR:
- (new) configurable option `migration_locator_class` which is set to `Phinx\Migration\Locator\LegacyLocator` by default (to support BC), and can be set to `Phinx\Migration\Locator\PsrLocator` to support the following:
  - filename is the same as the classname (using prefix `v` to prevent class-name starting with number)
  - support namespaces
  - make the `name` argument optional instead of required (`phinx create` command);
    - `PsrLocator`: if provided will suffix the file (and class!) with that string, e.g.: 
      - If name is provided (`foo`) phinx makes `v123_foo.php` and `v123_foo` class.
      - If no name is provided it will simple create the file and class without a suffix: `v123.php` and `v123` (class). 
      - The reason for this is that imo only versioned-classnames (`class v123 extends ...`) can be trusted and phinx should therefore never force me to use a non-unique name (`class myAmbigiousName extends ...`). Also see issue #566.
   - last but not least, by making the class configurable everyone can use their own logic for determining file paths and class names.


cleentfaar:
For people reviewing this, the following issues are still open:
- approval of this concept/change
- write tests (waiting for approval first)
- naming of `locator`, `LegacyLocator`, `PsrLocator`, `MigrationDefinition`; is this really clear to everyone? Suggestions are very welcome!

rquadling:
I like the idea a lot. One area that is a bit of a pain for my team is when we work in multiple branches, it would be very useful to have the branch id in the migration name. Our branches are DT-nnnn-blah-blah-blah. Having a mechanism that we can extend to extract the current branch and incorporate that into the final name would be very useful for us.

Also, by having the same mechanism to NOT allow us to create a migration in the master branch would be useful, so throwing an exception forces the devs to obey the rules - migrations must be in a branch.

cleentfaar:
@rquadling Thanks! Glad you can see a scenario for it. I've updated the description to reflect the flexibility you get from creating your own `Locator` class like you mentioned, not just using the `PsrLocator`.

dereuromark:
I also like it.
Please rebase so we can continue here.

filips123:
@cleentfaar @dereuromark Is there any update? I want to use something like this. See my issue #1459.

cleentfaar:
@dereuromark unfortunately I can not afford much spare time to work on this. Besides that I have since moved to a different solution (doctrine/migrations) which already supports some of the things I tried implementing in phinx. Feel free to finish it yourself, I will keep my fork available for a little longer so you can copy it if needed.

filips123:
Can someone finish this PR?
@cleentfaar What needs to be finished so it can be used?

dereuromark:
Looks like it just needs some CS fixes as outlined in the PR above.

cleentfaar:
Todo's are in my 1st comment: https://github.com/cakephp/phinx/pull/1087#issuecomment-299832086,  but I'd recommend starting from scratch if you want a proper solution (mine is mostly workarounds, short-term), perhaps use the same ideas but better to rewrite it imo.

PR #1089, opened 2017-05-12
[RuntimeException] Adapter "dblib" has not been registered

MubasharIqbalBajwa:
[RuntimeException] Adapter "dblib" has not been registered. This run time error occur when user command  phinx operation via linux


lorenzo:
We don't have any dublin adapter. You can use `sqlserver` adapter, though

PR #1090, opened 2017-05-12
doc: fix $this->query return type

glensc:
`$this->query` does not return array. altho it returns PDOStatement which is Iterable like array.

for example `print_r($rows)` will just print you:
```
PDOStatement Object
(
    [queryString] => SELECT * FROM users
)
```

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -180,10 +180,12 @@ Queries can be executed with the ``execute()`` and ``query()`` methods. The
             public function up()
             {
                 // execute()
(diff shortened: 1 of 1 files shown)


AubreySLavigne:
To be fair, the documentation immediately above the example correctly identifies the return object as a PDOStatement.

I've created a similar PR with https://github.com/cakephp/phinx/pull/1120 for a slightly different approach to the documentation issue.  I'm new to open source development, so feel free to yell at me, if I'm suggesting an amendment incorrectly this way.

PR #1091, opened 2017-05-14
Fix pgsql quoting in comment clause

ctizen:
Bug: pgsql adapter doesn't allow to add comments to columns if table or column has name matching one of reserved keywords (such as `user`).

Proposed fix: add quoting for table name and column name.


lorenzo:
Thanks!

Issue #1086, opened 2017-05-05
dropColumn fails on SQLite databases if the type is enum


betterphp:
Removing a column for SQLite works by creating a new table without the requested column then dropping it and renaming the new table to replace it.

If the column is an enum with more than one value option then this fails as the regex to remove the entry from the create table statement does not handle commas correctly.
~~~
[InvalidArgumentException]
The specified column doesn't exist: type

Exception trace:
() at ./vendor/robmorgan/phinx/src/Phinx/Db/Adapter/SQLiteAdapter.php:489
Phinx\Db\Adapter\SQLiteAdapter->dropColumn() at ./db/migrations/20170505100124_split_affects_fields.php:22
SplitAffectsFields->up() at ./vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php:125
Phinx\Migration\Manager\Environment->executeMigration() at ./vendor/robmorgan/phinx/src/Phinx/Migration/Manager.php:366
Phinx\Migration\Manager->executeMigration() at ./vendor/robmorgan/phinx/src/Phinx/Migration/Manager.php:342
Phinx\Migration\Manager->migrate() at ./vendor/robmorgan/phinx/src/Phinx/Console/Command/Migrate.php:113
Phinx\Console\Command\Migrate->execute() at ./vendor/symfony/console/Command/Command.php:264
Symfony\Component\Console\Command\Command->run() at ./vendor/symfony/console/Application.php:835
Symfony\Component\Console\Application->doRunCommand() at ./vendor/symfony/console/Application.php:200
Symfony\Component\Console\Application->doRun() at ./vendor/robmorgan/phinx/src/Phinx/Console/PhinxApplication.php:83
Phinx\Console\PhinxApplication->doRun() at ./vendor/symfony/console/Application.php:124
Symfony\Component\Console\Application->run() at ./vendor/robmorgan/phinx/bin/phinx:28
~~~
The message is confusing but it comes from the table essentially not existing due to failed queries.

I've attached ([phinx-sqlite-bug-demo.zip](https://github.com/robmorgan/phinx/files/979114/phinx-sqlite-bug-demo.zip)) migrations that cause this error, to reproduce it extract the zip then do
~~~bash
composer install
./vendor/bin/phinx migrate
~~~
from the folder.

A workaround I'm using for now is to change the column to a TEXT field before removing it so that the regex doesn't create invalid SQL.

pr0ggy:
Taking a look at this now, working on a fix

[...]

PR #1093, opened 2017-05-15
Fix issue with dropColumn command in SQLite (#1086)

pr0ggy:
Fix for the regex issue found (#1086) in `Phinx\Db\Adapter\SQLiteAdapter::dropColumn()` which resulted in malformed SQL being executed.  Note that included in this fix is a call to set the PDO error mode to throw exceptions in the event of error (malformed SQL being executed, etc) as I believe this will make such errors much more visible in the future (there were some failing unit tests that were covered up and marked as passing because of the exact same problem...malformed SQL being executed but no exceptions being thrown and no error codes checked for).  This was done in the `Phinx\Db\Adapter\PdoAdapter::setConnection()` method, so it will apply to all adapter classes derived from it.  I wanted to call this to attention as I'm not 100% certain this omission wasn't by design.  Let me know if any issues are found, will get them fixed ASAP.  Appreciate the time/effort put into Phinx thus far, I've found it very useful!

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -205,6 +205,9 @@ abstract class PdoAdapter implements AdapterInterface
     {
         $this->connection = $connection;
 
+        // ensure connection throws exceptions on errors (SQL syntax errors, etc)
+        $this->connection->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+
         // Create the schema table if it doesn't already exist
         if (!$this->hasSchemaTable()) {
             $this->createSchemaTable();
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -205,9 +205,6 @@ abstract class PdoAdapter implements AdapterInterface
     {
         $this->connection = $connection;
 
-        // ensure connection throws exceptions on errors (SQL syntax errors, etc)
-        $this->connection->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
-
         // Create the schema table if it doesn't already exist
         if (!$this->hasSchemaTable()) {
             $this->createSchemaTable();
--- a/tests/Phinx/Migration/Manager/EnvironmentTest.php
+++ b/tests/Phinx/Migration/Manager/EnvironmentTest.php
@@ -11,6 +11,7 @@ class PDOMock extends \PDO
 {
     public function __construct()
     {
+        parent::__construct('sqlite::memory:');
     }
 
     public function getAttribute($attribute)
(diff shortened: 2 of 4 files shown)


nebez:
I've been hunting down a Phinx bug for the last 5 hours and this PR pointed me in the right direction.
My issue is with `changeColumn()` while using the SQLite adapter. [This regex](https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/SQLiteAdapter.php?utf8=%E2%9C%93#L433-L438) is wrong and, when removing the default value of `''` from a string-like column, it almost entirely nukes the table and deletes a bunch of columns.
I haven't fixed it but at least now I know where the issue is coming from. I'll re-visit the issue when I have more time, for now I've just disabled migrations on my sqlite database.

nebez:
I fixed it for myself in this commit: https://github.com/lxrco/phinx/commit/cd20df1d503de98431334a70824c2f1d05b74854
If you want to use it, you can go ahead and use the branch. I can't promise I'm going to keep it up to date with upstream changes, but this fixes a huge problem for me (and, I imagine, for others as well).

lorenzo:
@nebez how about just sending a pull request with your fix?

nebez:
@lorenzo - the change has no test coverage, but I will send a PR now regardless.

pr0ggy:
@lorenzo any thoughts on my PR here?  I see that the appveyor build failed but everything passes locally for me...judging from AppVeyor output, some needed PDO driver isn't being loaded, but I'm not sure why this would be the case.  I'm assuming it has something to do with the change I made in [EnvironmentTest.php](https://github.com/cakephp/phinx/pull/1093/files#diff-b2ce894cb8d20bcaf61fd69f7bd3fe9e) to create an in-memory DB but I'm not entirely sure why that would be throwing a missing driver exception...nothing I can find online indicates any required voodoo to enable the memory driver.  Appreciate any thoughts you have, thanks!

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -205,6 +205,9 @@ public function setConnection(\PDO $connection)
     {
         $this->connection = $connection;
 
+        // ensure connection throws exceptions on errors (SQL syntax errors, etc)
+        $this->connection->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
```
lorenzo: Please remove this line, this is done here already https://github.com/cakephp/phinx/pull/1122/files

tests/Phinx/Migration/Manager/EnvironmentTest.php
```diff
@@ -11,6 +11,7 @@ class PDOMock extends \PDO
 {
     public function __construct()
     {
+        parent::__construct('sqlite::memory:');
```
lorenzo: I think the problem is here. You are forcing all tests to use `sqlite::memory`, regardless of what the tests is setup to do.

pr0ggy: Appreciate it Lorenzo, will take a look at those 2 items ASAP

pr0ggy:
@lorenzo I apologize for the delay, finally had some time to look into this again.  Requested changes have been made.

lorenzo:
@pr0ggy Thanks!

PR #1111, opened 2017-06-11
fixed miss method and return type error

vtumi:
use the table_prefix option, i found an error.

$table = $this->table('user');
$table->truncate();

PHP Fatal error: Uncaught Error: Cannot call abstract method Phinx\Db\Adapter\AdapterInterface::truncateTable()


yashtodi94:
Please do provide a description for the pull request.

robmorgan:
thanks!

PR #1113, opened 2017-06-13
Quote columns in postgres index definitions

peets:


lorenzo:
Seems like this was already done

PR #1120, opened 2017-06-22
Correct `Migration->query($sql)` Example

AubreySLavigne:
The Description in the preceding documentation says `query()` will return a PDO Statement, but the sample code and comment implies it will return an array instead.


AubreySLavigne:
Alternate PR https://github.com/cakephp/phinx/pull/1090 by glenrc

lorenzo:
Thanks!

PR #1122, opened 2017-06-26
Force exception onto PDO

Grafikart:
## Problem

If someone uses an unconfigured PDO instance as a connection migration won't stop if an error is encountered (by default pdo is set to PDO::ERRMODE_SILENT)

## Solution

Add a PDO attribute to throw an exception on failed queries.


lorenzo:
This make a lot of sense to me.
Do you think you can also fix the test this change broke?

Grafikart:
@lorenzo I fixed failing tests and added an extra test to ensure this parameter is set correctly. (I can rebase the history if needed or you can squash the PR)

lorenzo:
Thanks @Grafikart

PR #1126, opened 2017-07-01
[SQLite] Fix invalid changeColumn regex causing deletions

nebez:
This fixes an issue in the SQLiteAdapter that causes columns to be destroyed when removing a default value of `''` from a string column.
See comment here: https://github.com/cakephp/phinx/pull/1093#issuecomment-312373192

Unfortunately, there are no existing tests for this specific functionality that prove what I've changed hasn't broken anything.


lorenzo:
Thanks, I can actually see how the previous regex was wrong and this one is safer.

Issue #5, opened 2011-12-27
Phinx Export Feature


robmorgan:
It would be nice to run `phinx --export` to export a json/xml/SQL representation of your database migrations. What purpose would this serve though?

weemen:
+1 It would help me to get a nice solution for this problem: http://stackoverflow.com/questions/28792430/workflow-with-database-create-alter-and-drop-commands-and-web-applications-in-pr

RQuadling:
After a few months, we clear out our migrations. After a year and 5 developers, we had well over 200 migrations. And that was from a DB that was NOT phinx controlled. When we deploy a feature branch, we clear out the migrations and rebuild our base install.sql. We also have a preset set of data which is retrieved from the live DB so that a new developer gets a decent clean set of data and a completed DB schema.

Having a standardised approach (one which we can expand upon to include our specifics) would be a great feature.

nickdietel:
+1 would be helpful for someone who needs to review/debug migrations. Also it would support executing queries on another machine without phinx.

zondor:
+1 for export feature to compact big list of migrations in long project.
+1 running automation tools became faster, running 1 sql file rather then all migrations

As extra will be nice to have first migration generation from existing database. As phinx is really useful in custom projects, others frameworks already have database migrations and seeding. But custom projects usualy already have database and its regular monky job converting sql to phinx

wweaver:
What about a way to reconcile the files down to minimal creates and updates. Meaning if a table is created in one migration, then updated in another, to be able to reconcile this down to one file that has a create? I can see this not being quite so simple, especially if there are multiple things happening in one migration, but it still would be a nice to have.

Another plus might be to have the files generated into subdirectories by year/month so that there aren't so many files in one directory. Right now my company is up to 834 files in almost 2 years.

celorodovalho:
:+1:

washingtonphp:
+1 on that

foalford:
+1 In fact missing of this feature makes me hesitate of adding phinx into my project.

1. Plain SQL is easy to review and be inspected on in troubleshooting. 
2. Less dependency, i.e. phinx, less moving parts in automated deployment.

MightyMimi:
+1

masom:
@foalford Phinx is amazing, don't let that missing feature stop you from using it.

foalford:
[...]

PR #1127, opened 2017-07-04
Add dump option

kerneldemon:
Hey,

I'd like a way to dump plain sql using the `phinx migrate` and `phinx rollback` commands, so I added a `--dump` option.

This is useful for those who would like to read raw queries before running them in production, there's a discussion about this [here](https://github.com/cakephp/phinx/issues/5).


src/Phinx/Console/Command/Migrate.php
```diff
@@ -47,6 +47,7 @@ protected function configure()
              ->setDescription('Migrate the database')
              ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
              ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')
+             ->addOption('--dump', '-x', InputOption::VALUE_NONE, 'Dump query to standard output instead of executing it')
```
lorenzo: Do you mind renaming this to `dry-run` or `attempt`?

lorenzo:
I'm not sure this solution is actually going to work for all cases, but it is a good start.

kerneldemon:
@lorenzo the name `dry-run` sounds about right

lorenzo:
Thanks @kerneldemon! Would you mind adding a brief explanation of this new flag to the docs?

kerneldemon:
@lorenzo I contemplated adding a separate section that would describe the option, but since the option can only be used by two commands, I thought it would be better to just have the same text for migrate and rollback command section.

lorenzo:
@kerneldemon duplicating the test sounds good

kerneldemon:
@lorenzo Moved the text slightly. Is anything else missing?

lorenzo:
Thanks for the contribution!

PR #1132, opened 2017-07-13
Update yml file to surround parameters with quotes

eddturtle:
Hi, getting a php non fatal error being shown in error_get_last() due to a symfony deprecation.

    Not quoting the scalar "%%PHINX_CONFIG_DIR%%/database/migrations" starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0.

Info about this error: http://symfony.com/blog/new-in-symfony-3-1-yaml-deprecations#deprecated-starting-scalars-with-characters


rquadling:
Thank you.

PR #1133, opened 2017-07-13
SQLite - Revert back to RegEx for dropping enum columns (#1086)

nebez:
Related to https://github.com/cakephp/phinx/pull/1093 - this PR modifies the original RegEx that was used for dropping columns.
The issue was that ENUMs weren't properly caught by the RegEx and thus produced invalid SQL when re-creating the table. This regex should fix that.


lorenzo:
Thanks @nebez

PR #1144, opened 2017-08-07
Add support for PostgreSQL citext column type

SeinopSys:
This pull request is meant to add support for the [`citext`](https://www.postgresql.org/docs/9.2/static/citext.html) data type (generally available under `postgresql-contrib-*` packages).

I'm migrating my own project to Phinx (eventually CakePHP) and the lack of `citext` support caused me to hit a roadblock in the process. I believe my changes follow the guidelines found in the repository, but if not please do let me know.

I verified that all PostgresAdapter tests passed on a local PostgreSQL 9.6 server and I attempted the `docker-compose` route as well, but I got the following:
```
$ docker-compose run phinx
Starting phinx_postgres_1 ... 
Starting phinx_mysql_1 ... done
This version of PHPUnit is supported on PHP 5.6, PHP 7.0, and PHP 7.1.
You are using PHP 5.4.16 (/usr/bin/php).
```

In essence the only difference in the testing process after this change is having to run `CREATE EXTENSION citext` after the database creation and before the testing of tables starts, for which I added commands into both the `PostgresAdapterTest` class as well as the `.travis.yml` file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_script:
     - travis_retry composer update --no-interaction --prefer-source --prefer-stable ${COMPOSER_FLAGS}
     - mysql -e 'create database phinx_testing;'
     - psql -c 'create database phinx_testing;' -U postgres
+    - psql -c 'create extension if not exists citext;' -U postgres
 
 script:
     - vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -788,6 +788,14 @@ collation set collation that differs from table defaults *(only applies to MySQL
 encoding  set character set that differs from table defaults *(only applies to MySQL)*
 ========= ===========
 
+For the ``text`` column only:
+
+============== ===========
+Option         Description
+============== ===========
+case_sensitive set case-sensitivity via the ``citext`` extension *(only applies to PostgreSQL)*
+============== ===========
+
 For foreign key definitions:
 
 ====== ===========
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -143,7 +143,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
      */
     public function quoteColumnName($columnName)
     {
-        return '"'. $columnName . '"';
+        return '"' . $columnName . '"';
     }
 
     /**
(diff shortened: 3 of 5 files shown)


lorenzo:
I'll review this over the weekend.

After a quick glance I'm not convinced in the way you approached doing this. Although I like the general idea of supporting this type of column

lorenzo:
In general it looks good and I like the idea, but I don't feel comfortable adding the `caseInsensitive()` function to the column, without having code supporting the feature in other drivers.

SeinopSys:
As far as I could tell, the other DBMS-specific features (e.g. `collation` for MySQL/SQLite) don't have respective methods in other drivers.

I haven't used other DBMS systems, but in MySQL case-sensitivity is handled by the collation (`*_ci`) and not the column type, which would make it difficult to change with a simple boolean.

I personally think it's fine to have such an option that only applies to one DBMS, but it's ultimately up to you to decide.

lorenzo:
@SeinopSys Yeah, it is not bad having it for only one driver, but the other drivers should at least emit a warning / log line explaining that the feature is not going to be applied since it is not supported

lorenzo:
What I mean is: The feature as it is currently implemented is quite generic, but only generalises one specific case.

Maybe we can add a different column type instead of a flag to the column?

SeinopSys:
Adding a different column type is certainly possible. I originally wanted to add a separate type, but seeing another Postgres-specific feature (`PostgresAdapter::INT_SMALL`) was handled I decided to make it a flag instead.

I also wanted to ask if other [Postgres-only](https://www.postgresql.org/docs/9.2/static/datatype.html#DATATYPE-TABLE) types (e.g. `xml`, `serial` - shortcut for making a column auto-increment, useful when adding a sequential primary key to a table that didn't have one as part of a migration) types are expected to be supported somehow, e.g. through a "raw" method (passing a special class instance as the name) or a drop-in replacement to `addColumn` like `addCustomColumn($name, $type, $options)` that'd support universal options like `null` that's often available for all columns or making it a string that just gets appended to the column definition.

If there's any initiative or discussion on this I'd love to contribute as that could be a better and more universal solution than adding in specific types like this one by one. If that's not or the radar then let me know and I'll just go ahead and make this a different column type for now.

lorenzo:
@SeinopSys I like the idea of the custom column method

SeinopSys:
I'm going to close this PR for now then and hopefully work out a solution like that and submit a new PR once that's ready.

PR #1147, opened 2017-08-14
Added documentation about PHPUnit integration

Grafikart:
English is not my main language but tell me if something is wrong and i'll correct it.


lorenzo:
Thanks!

PR #1148, opened 2017-08-15
When execute command `seed:run`, insert all data to a table in a bulk.

kenzo0107:
When execute command `seed:run`, insert all data to a table in a bulk.
It take much time to insert data to a table one by one until now.

An example is shown below:
```
$table->bulk()
      ->insert($data)
      ->save();
```

It became more than triple as fast by Bulk Insert when inserting 1,000 records.
It is very effective when many data samples are required.

However, it is assumed that the keys of data to be set are unified. 
The method `bulk()` are not available in the following example:

```
$data = array(
    array(
        'column1' => 'value1',
        'column2' => 1,
    ),
    array(
        'column1' => 'value2',
        'column3' => 'foo',
    ),
    array(
        'column1' => 'value3',
        'column2' => 3,
        'column3' => 'foo',
    )
);

$table->bulk()
  ->insert($data)
  ->save();
```

The function of dynamically checking the key of the set data is not added,  
because it will cause performance degradation.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -182,6 +182,7 @@ abstract class PdoAdapter extends AbstractAdapter
         $keys = array_keys($current);
         $sql .= "(". implode(', ', array_map(array($this, 'quoteColumnName'), $keys)) . ") VALUES";
 
+        $vals = array();
         foreach ($rows as $row) {
             foreach($row as $v) {
                 $vals[] = $v;
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -182,9 +182,11 @@ abstract class PdoAdapter extends AbstractAdapter
         $keys = array_keys($current);
         $sql .= "(". implode(', ', array_map(array($this, 'quoteColumnName'), $keys)) . ") VALUES";
 
-        $objTmp = (object) array('aFlat' => array());
-        array_walk_recursive($rows, create_function('&$v, $k, &$t', '$t->flat[] = $v;'), $objTmp);
-        $vals = $objTmp->flat;
+        foreach ($rows as $row) {
+            foreach($row as $v) {
+                $vals[] = $v;
+            }
+        }
 
         $count_keys = count($keys);
         $query = "(" . implode(', ', array_fill(0, $count_keys, '?')) . ")";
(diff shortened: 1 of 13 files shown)


src/Phinx/Db/Table.php
```diff
@@ -75,6 +75,11 @@ class Table
     protected $data = array();
 
     /**
+     * @var boolean
+     */
+    protected $bulk = false;
```
lorenzo: I don't understand what this flag is for

kenzo0107: This flag is prepared to judge whether to use the bulk insert function.

lorenzo: Why not always use it?

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -167,6 +167,36 @@ public function insert(Table $table, $row)
...
+        $sql = sprintf(
+            "INSERT INTO %s ",
+            $this->quoteTableName($table->getName())
+        );
+
+        $current = current($rows);
+        $keys = array_keys($current);
+        $sql .= "(". implode(', ', array_map(array($this, 'quoteColumnName'), $keys)) . ") VALUES";
+
+        $objTmp = (object) array('aFlat' => array());
```
kenzo0107: Oh, typo !
fix it !

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -167,6 +167,36 @@ public function insert(Table $table, $row)
...
+        array_walk_recursive($rows, create_function('&$v, $k, &$t', '$t->flat[] = $v;'), $objTmp);
```
kenzo0107: Thank you for your review !

I tried to simply coded in one line.
But there was not enough consideration to Phinx.

Fix it !

Thx.

src/Phinx/Db/Table.php
kenzo0107: I see.
There is a way I can implement this without adding the bulk flag.
Thx.

kenzo0107:
@lorenzo 

I could implement Bulk Insert without adding the bulk flag.

lorenzo:
I think the only thing this is missing is tests showing that the method actually works

kenzo0107:
Thank you for review !

I added to test the new method with the following method.

- tests/Phinx/Db/Adapter/TablePrefixAdapterTest::testInsertData
- tests/Phinx/Db/TableTest::testInsertSaveData
- tests/Phinx/Migration/AbstractMigrationTest::testInsert

Is it necessary elsewhere ?

lorenzo:
@kenzo0107 I didn't see any tests where the method was actually used. Only mocked tests were used, so the actual implementation is not tested

kenzo0107:
@lorenzo Thank you for your review. I added some tests where the method is actually used. 
Please review again.

lorenzo:
Thanks @kenzo0107 !

lorenzo:
@kenzo0107 Would you mind sending another pull request with a short example for the docs?

kenzo0107:
@lorenzo Thank you 😄 I will send another pull request soon.

PR #1151, opened 2017-08-21
Add direction parameter to migration change method

Roy-Orbison:
Enables otherwise irreversible migrations to become reversible. e.g. inside a `change()`:

```php
$this->table('foo', ['id' => false])
    # more columns
    ->create();
if ($direction === MigrationInterface::UP) {
    # some change to new table/data that requires no down() counterpart, e.g. custom compound primary key.
}
```

### DIFF
--- a/src/Phinx/Migration/Manager/Environment.php
+++ b/src/Phinx/Migration/Manager/Environment.php
@@ -114,7 +114,7 @@ class Environment
                     ->getWrapper('proxy', $this->getAdapter());
                 $migration->setAdapter($proxyAdapter);
                 /** @noinspection PhpUndefinedMethodInspection */
-                $migration->change();
+                $migration->change($direction);
                 $proxyAdapter->executeInvertedCommands();
                 $migration->setAdapter($this->getAdapter());
             } else {


lorenzo:
Would you mind adding a couple tests for this new feature?

lorenzo:
The base migrations class would also need the method signature changed

garex:
Can this be implemented without touching `change` signature? Somehow by protected method like `isUp(): bool` inside migration?

Roy-Orbison:
I don't see any binding signature for the method, as it is optional.
Therefore, adding a parameter where none existed should be a non-breaking
change.

Had a deeper look at this the other night but it keeled over, after table
creation, upon `insert()`ing data, without any error to the console.

On 26 Aug 2017 02:25, "Alexander Ustimenko" <notifications@github.com>
wrote:

Can this be implemented without touching change signature? Somehow by
protected method like isUp(): bool inside migration?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/cakephp/phinx/pull/1151#issuecomment-324977460>, or mute
the thread

Roy-Orbison:
I don't see any binding signature for the method, as it is optional.
Therefore, adding a parameter where none existed should be a non-breaking
change.

Had a deeper look at this the other night but it keeled over, after table
creation, upon `insert()`ing data, without any error to the console.

garex:
I just see another issue here: to know if we up/down we should import some `MigrationInterface::UP`.

From developer usage simple `isUp()` or `youNameIt()` flag method is better, that scalar variable, that you need to know from which interface to import to use it.

Current suggestion is not usable and could be potential error-source.

Roy-Orbison:
A method would be fine. How about "goingUp()"?

garex:
Anything you wish. Main idea: I'm lazy and just want to write inside `$this->go` + ctrl + space and my IDEtor will just do all the things for me.

PR #1153, opened 2017-08-25
Fixed documentation example of usage phpunit

XAKEPEHOK:
Code block was loosed because no empty line after .. code-block:: php


XAKEPEHOK:
What about fixing it on site?
http://docs.phinx.org/en/latest/commands.html#using-phinx-with-phpunit

lorenzo:
The documentation will soon be moved to another URL and updated there.

PR #1155, opened 2017-08-25
Fill up column values from schema

garex:
When we create enum's in DB we should be able to read their's values back in migrations too.


lorenzo:
Tests are failing

garex:
rebased on latest master, fixed and forced back

lorenzo:
Thanks!

PR #1156, opened 2017-08-26
Fix testing under docker 2

garex:
Current dockerized testing experience is this:

We should bump dockerized version to 7.0 as it's a minimal possible version from dev-dependencies.

PHP 5.4 and above will be caught by `travis`.

````
root@74dd8a8e4d55:/src# composer install                
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for symfony/config v3.3.6 -> satisfiable by symfony/config[v3.3.6].
    - symfony/config v3.3.6 requires php >=5.5.9 -> your PHP version (5.4.45) does not satisfy that requirement.
...
  Problem 18
    - phpdocumentor/reflection-docblock 3.2.2 requires php >=5.5 -> your PHP version (5.4.45) does not satisfy that requirement.
    - phpspec/prophecy v1.7.0 requires phpdocumentor/reflection-docblock ^2.0|^3.0.2 -> satisfiable by phpdocumentor/reflection-docblock[3.2.2].
    - Installation request for phpspec/prophecy v1.7.0 -> satisfiable by phpspec/prophecy[v1.7.0].
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,10 +63,10 @@ Great, so you want to contribute. Let's get started:
 ## Getting Started with docker
 
 If you do not wish to install MySQL and Postgres locally to run unit tests
-can use [docker-compose](https://docs.docker.com/compose/) which will start 
+can use [docker-compose](https://docs.docker.com/compose/) which will start
 both database, install all development dependencies and run all unit tests.
 
-To get started, just run `dco run --rm phinx`. It will download all 
+To get started, just run `docker-compose run --rm phinx`. It will download all
 images, build&start development container create `phpunit.xml` configuration and switch you to that container. So just install dependencies with `composer install` and run unittests with `vendor/bin/phpunit`.
 
 ## Documentation
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:5.4
+FROM php:7
 
 # system dependecies
 RUN apt-get update && apt-get install -y \


lorenzo:
What's the purpose from your perspective for having the docker image?

If we cannot test on the minimum supported version, we should either fix the testing scripts or raise the minimum supported PHP version

garex:
@lorenzo docker image is for developer and for quick local testing. So it's for speed-up develoment.

Travis is for checking, that system works at minimal supported PHP version.

> we should either fix the testing scripts

We can't as may dev dependencies are above 5.4

>  or raise the minimum supported PHP version

Also radical solution which will not be good for those, who stuck at php 5.4

garex:
Currently docker image is not works at all. And I'm not sure -- it worked before.

So another solution is just to remove docker image and docker testing from project. But it's bad option.

garex:
1uurui

May be I missed `--prefer-lowest` flag?

garex:
In this case we should just update docs here: https://github.com/cakephp/phinx/blob/master/CONTRIBUTING.md#getting-started-with-docker

To run not install, but `composer update --prefer-lowest` before running tests.

lorenzo:
Can we at least make it work with php 5.6? I think that is a good compromise

garex:
No. Let's stick on 5.4, but update docs about `--prefer-lowest`. I will prepare PR 3 regarding this ))

PR #1157, opened 2017-08-28
Migrate/Rollback by specifying migration class names.

kristijorgji:
Allows to migrate or rollback specifc migration(s) by specifying the class names.
Phinx TextWrapper was also updated to accept the new option migrations which can contain the string in the following format "className1 className2..."

Example usage:

phinx migrate -m CreateUsersTable CreateSomethingElse ...
phinx rollback -m CreateUsersTable...


lorenzo:
will this run only the specified migration? Or the list of migrations leading to that one?

This requires tests before it can be merged.

kristijorgji:
@lorenzo It allows more then one to be executed. I use this for a complex system and integration with phpunit via phinx textwrapper. I will write the unit tests and make another pull request.

So you can execute one or more migration/rollback by specifying the class name as in my examples phinx migrate -m CreateUsersTable CreateSomethingElse ...
phinx rollback -m CreateUsersTable...

lorenzo:
@kristijorgji My question is more if the migrations sitting in the middle of the current state and the one you are pointing at are ignored entirely.

Fore example, Imagine you have

* Migration 1
* Migration 2 <- currently here
* Migration 3 <- pending
* Migration 4 <- pending

If I do:

```
phinx migrate -m Migration4
```

Will it ignore `Migration 3` ?

kristijorgji:
@lorenzo yes, it will only execute the specified migrations. This is the intended behavior.

And can be used with phpunit also. Example one test method may want to migrate only 2 tables and roll back those again.

You guys have other methods for migration to version/date, the purpose of my changes is to allow to migrate only the specified classes.

Another use case may be for example a migration named UpdateFrResetEmail, and I want to run only these and not all migrations before this one.

kristijorgji:
It is so funny that after running my tests 3 other tests fail only for  your build, not on my local machine...
I am checking again to figure out what is happening because from the commits make no sense.

lorenzo:
I don't think the intended behaviour is a good one though. Migrations are by definition dependent on past state: You cannot change a column if it is not present, you cannot add a index to a column that is not there.

In my opinion, migrating to a specific class, should also execute all the steps in between.

kristijorgji:
@lorenzo That functionality already exist. Other migration frameworks allow specifying only the classes that you want, why should Phinx not offer that too?

After all is up to the end users to use whatever commands that they want. I gave you 2 useful examples.
The main goal is for unit testing though. In Laravel it was so easy to just define what you want to run/rollback for each test method and could cut down db testing by at least 5 or more times.

lorenzo:
What other migrations frameworks, apart from Laravel's? I'm not saying this is not useful, just that it needs minor adjustments: run the migrations from current state util specified class.

Unit testing can be solved more elegantly than running migrations, in my opinion. A different take is what CakePHP does: you define fixture files that represent the state of the database you want to have for your test, and these migrations-like files are executed before each of your tests.

If you are interested in the alternative solution, I think I could package it as a separate library and publish it in packagist.

kristijorgji:
@lorenzo We are also using fixtures. We use migrations to create/drop certain tables and fixtures to populate the test data. Is a very efficient and fast combination.

So basically there exist an empty database test. Before/after each method the specified migrations are run/rolled back then fixtures populate the test data that you want. We are having great results with this approach at least 10 times faster our unit tests. And now will also use in memory engine

I found out by the way what was causing 3 tests fail (I added 2 more dummy migrations that were read by other tests). Will fix and push.

lorenzo:
@kristijorgji CakePHP fixtures are different to what you mean. They are literally what you described: you specify in your test load these 3 tables and wit will create and optionally drop them after the test.

I know how efficient that workflow is, because that's the workflow I use. What I mean is, I don't think it belongs to Phinx. At least it does not belong to the CLI.

If this were to be merged, it should only be possible using the Phinx PHP api, and not possible from the command line, as it opens the door for many future problems.

kristijorgji:
@lorenzo I see, ok I was referring to Phpunit fixtures as I never used CakePhp so far.  By the way I am curious what problems may this cause if used by the command line?

This will be a separate command with -m parameter as in the examples and the users that will use this will have their own reasons to do so.

lorenzo:
The problem is leaving the database in an inconsistent state because previous migrations were not run (either up or down). I really want to prevent those cases as breaking other peoples database is a very serious thing.

We can of course blame others for not reading the docs or using the tool incorrectly, but it would we our responsibility for actually allowing database to be inconsistent. I'd rather not do that.

kristijorgji:
@lorenzo Hm... ok I understand your concern.

Well what about this suggestion than:
Aside from the description in the CLI of the command and in the manual I also wait for confirmation prompt? 

And also honestly I don't think anyone would complain if they use some tool without reading what it does though, especially devs.

lorenzo:
@kristijorgji what about just allowing it when calling it form PHP and not expose it from the command line?

> And also honestly I don't think anyone would complain if they use some tool and cause damage without reading what it does though, especially devs. (at least I have not seen anyone of my devs xd so far do this with other migration frameworks)

This actually has happened in the past, both to me and others. Developers are not great at reading docs, we make a lot of assumptions.

kristijorgji:
@lorenzo ok then I can remove any hint about this functionality from the CLI, but I cannot find some way to add this functionality without having to enable it also in the CLI.

lorenzo:
@kristijorgji let me investigate other options and I'll come back to you

dereuromark:
@lorenzo ping

jmorganmartin:
@lorenzo I wish you would reconsider this. The justification that people might use it without reading the manual and wreck their DBs is pretty weak, IMO. People can just as easily wreck their DB by using Phinx without reading the manual _without_ the `-m` option (e.g., they could write a migration with an incorrect/missing `down()` method, etc).  Adding the `-m` option would not significantly increase the risk of people wrecking their DBs, but it would increase the flexibility and usefulness of Phinx.

lorenzo:
@jmorganmartin I would accept something like this if a system of "dependencies" is implemented for Phinx.

A list of dependencies for a migration file is just a list of class names pointing to other migrations. If there are no dependencies, then we can assume it is safe to rollback in any order. If there are dependencies, then does dependencies need to be rollback first.

@jmorganmartin Would you like to implement this feature? I can help you and guide you through the process.

dereuromark:
ping @jmorganmartin

othercorey:
Closing due to no activity.

PR #1158, opened 2017-08-28
Fix testing under docker 3

garex:
Ok, now I just simplified docker-compose a little and actualize howto about running composer inside docker image.

So it's now same as in travis: `--prefer-lowest`


lorenzo:
Thanks!

PR #1164, opened 2017-08-31
Better reversible migrations

Roy-Orbison:
Enables otherwise irreversible migrations to become reversible, and makes
it possible to use a single migration where previously two were required.

Pursuant to https://github.com/cakephp/phinx/pull/1151

### DIFF
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -66,6 +66,11 @@ abstract class AbstractMigration implements MigrationInterface
     protected $input;
 
     /**
+     * @var bool
+     */
+    protected $goingUp;
+
+    /**
      * Class Constructor.
      *
      * @param int $version Migration Version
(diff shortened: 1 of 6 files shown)


src/Phinx/Migration/AbstractMigration.php
```diff
@@ -66,6 +66,11 @@
     protected $input;
 
     /**
+     * @var bool
+     */
+    protected $goingUp;
```
lorenzo: This would not be a boolean or be accurate when using the up and down methods, would it?

Roy-Orbison:
No, but you move the setter call up so it would. It seemed superfluous.

lorenzo:
It does seem superfluous, but it is not. Having the state of an object consistent is important because you don't know the different permutations a particular object can be used

src/Phinx/Migration/AbstractMigration.php
Roy-Orbison: Fixed.

mvrhov:
So should I close the #788 ?

lorenzo:
@mvrhov yes

lorenzo:
Thanks!

mvrhov:
@lorenzo I know you already merged but? Wouldn't it be better if `getGoingUp` would be named `isMigratingUp` and the set method `setMigratingUp`

lorenzo:
@mvrhov Sure, it like that.. care to open a new PR with the change?

mvrhov:
Sorry I don't have time :/ Maybe @Roy-Orbison can find some.

PR #1165, opened 2017-08-31
Apply default collation on tables as per documentation

nebez:
As per the documentation on [environment configuration](http://docs.phinx.org/en/latest/configuration.html#environments), there should be an option named `collation`. This configuration option doesn't do anything when you set it.

This change will set the default collation of newly created tables in this order:

1. The `collation` option provided in the array of options
2. The `collation` option provided in the environment configuration
3. `utf8_general_ci`


lorenzo:
Thanks!

nebez:
@lorenzo - done. I moved the PR to #1169 instead.

Issue #1003, opened 2016-12-29
Microseconds support for MySQL >= 5.6.4


lorenzomar:
Starting from version 5.6.4 MySQL added fractional seconds support for `DATE`, `DATETIME` and `TIMESTAMP` [doc](https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html). The standard practice is to use date fields without fractional precision but i suppose it is due to absence of the feature. 
What about to add the `limit` option to date fields too?

dereuromark:
Are you able to provide a pull request with your suggested changes?

Issue #904, opened 2016-07-14
Support milliseconds for timestamp columns on adapters that support it.


rquadling:
In mysql, `CURRENT_TIMESTAMP` can also be `CURRENT_TIMESTAMP(x)` where `x` is from 0 to 6.

In addition to that, `ON UPDATE CURRENT_TIMESTAMP(x)` can be used for the `updated_at` column, thereby providing an automatic record to a high resolution of when the row was updated.

For example, in MySQL

``` sql
  `createdAt` TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
  `updatedAt` TIMESTAMP(6) NULL DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP(6),
```

Research what options exist for Postgres, Microsoft SQL Server and SQLite.

dereuromark:
Is this still open?

othercorey:
I believe support was added for both mysql and postgres.

PR #1039, opened 2017-02-15
Precision for PostgreSQL time types

msbit:
This allows specifying a precision value for `time` and `timestamp` in PostgreSQL, which gets mapped to the `datetime_precision` attribute in the `information_schema`:

```
$this->table('table1')->addColumn('timestamp1', 'timestamp', array(
    'precision' => 0
))
```

PR #1166, opened 2017-09-04
add optional limit parameter for mysql datetime, time and timestamp types

gnoddep:
From Mysql 5.7 and MariaDB 5.3 the DATETIME, TIME and TIMESTAMP types support fractional seconds with a precision from 0 to 6. This PR makes it possible to create tables with those columns.

As mentioned in #904 and #1003


adduc:
There's a similar pull request in #1039 that adds precision for PostgreSQL time fields.

Should limit be renamed to precision to match?

gnoddep:
I think `precision` is the better term for this feature, I'll look into it when I'm home

PR #1167, opened 2017-09-05
fix : use the default_migration_table yml directive if present

khayyam90:
When phinx.yml defines a custom default_migration_table value, this value is not used. "phinxlog" is still used.

 .... setSchemaTableName() is never used in the whole project ....

The problem was already present in <0.9 versions but not for PDOAdapter, like if this default_migration_table directive was only available for PDOAdapters. I am using PDOAdapter subclasses so I didn't notice it before. Since 0.9.0 version it is blocking even when using pdo adapters.


JayPHP:
Thanks!

PR #1169, opened 2017-09-05
Apply default collation on tables as per documentation (new #1165)

nebez:
_Closed and re-opened from #1165 as I put new commits into my own master branch - use this one instead please_

As per the documentation on [environment configuration](http://docs.phinx.org/en/latest/configuration.html#environments), there should be an option named `collation`. This configuration option doesn't do anything when you set it.

This change will set the default collation of newly created tables in this order:

1. The `collation` option provided in the array of options
2. The `collation` option provided in the environment configuration
3. `utf8_general_ci`

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -206,7 +206,12 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             'engine' => 'InnoDB',
             'collation' => 'utf8_general_ci'
         ];
-        $options = array_merge($defaultOptions, $table->getOptions());
+
+        $options = array_merge(
+            $defaultOptions,
+            array_intersect_key($this->getOptions(), $defaultOptions),
+            $table->getOptions()
+        );
 
         // Add the default primary key
         $columns = $table->getPendingColumns();
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -300,6 +300,32 @@ class MysqlAdapterTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('MyISAM', $row['Engine']);
     }
 
+    public function testCreateTableAndInheritDefaultCollation()
+    {
+        $options = [
+            'host' => TESTS_PHINX_DB_ADAPTER_MYSQL_HOST,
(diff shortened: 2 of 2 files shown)


nebez:
Any chance this gets merged?

nebez:
Any news on getting this merged?

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -206,7 +206,12 @@ public function createTable(Table $table)
             'engine' => 'InnoDB',
             'collation' => 'utf8_general_ci'
         ];
-        $options = array_merge($defaultOptions, $table->getOptions());
+
+        $options = array_merge(
+            $defaultOptions,
+            array_intersect_key($this->getOptions(), $defaultOptions),
+            $table->getOptions()
+        );
```
dereuromark: are those assoc arrays? then we should merge with + Operator IMO

chinpei215:
What if we remove the hardcoded default settings instead? I think we should not overwrite default settings on the database.

nebez:
@chinpei215 - I totally agree that Phinx setting its own default to override the DB's default is a bit odd. But would that be considered backwards incompatible? Phinx has always been doing it, I think if we changed it now that would break a lot of setups if people relied on that default collation.

chinpei215:
@nebez Fair points. OK, I will confirm your patch on my local  environment, as I have a little concern about charset/collation mismatch between database settings and configurations (I may be wrong though). And if there are no problems, I will also approve your pull request.

chinpei215:
@nebez Personally, I would like to remove the default entirely. I think it would confuse users. And if someone wants some default collation and/or engine, I think they should change `DEFAULT CHARACTER SET` on the database, and/or the `default_storage_engine` variable. But yes, probably we would need to communicate with others.

chinpei215:
OK. First I thought we should merge not only `collation`, but also `charset`. But now I know the option is not used when creating tables. `CHARACTER SET` seems to be determined by collation:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/MysqlAdapter.php#L242-L246

So there is no need to merge the option at this point. If no one argues, I will merge this PR.

nebez:
Sounds good @chinpei215 - thanks for taking this on!

chinpei215:
Merged. Thank you for your contribution.

Issue #1170, opened 2017-09-05
Undefined index when using 'connection' option


thewunder:
This is a regression after updating to 0.9.0, looks like the connection option (the pdo instance) is not getting set on the adapter.

```
Error: Undefined index: host on line 89 of /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php in /src/Lifestyle/ErrorHandler.php:53
Stack trace:#0 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php(89): Lifestyle\\ErrorHandler->errorHandler(8, 'Undefined index...', '/var/www/dev_20...', 89, Array)
#1 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(88): Phinx\\Db\\Adapter\\MysqlAdapter->connect()
#2 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(128): Phinx\\Db\\Adapter\\PdoAdapter->getConnection()
#3 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(146): Phinx\\Db\\Adapter\\PdoAdapter->query('SELECT * FROM p...')
#4 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(231): Phinx\\Db\\Adapter\\PdoAdapter->fetchAll('SELECT * FROM p...')
#5 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(208): Phinx\\Db\\Adapter\\PdoAdapter->getVersionLog()
#6 /vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterWrapper.php(215): Phinx\\Db\\Adapter\\PdoAdapter->getVersions()
#7 /vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php(258): Phinx\\Db\\Adapter\\AdapterWrapper->getVersions()
#8 /vendor/robmorgan/phinx/src/Phinx/Migration/Manager.php(300): Phinx\\Migration\\Manager\\Environment->getVersions()
#9 /vendor/robmorgan/phinx/src/Phinx/Console/Command/Migrate.php(114): Phinx\\Migration\\Manager->migrate('dev', NULL)
[...]

PR #1171, opened 2017-09-07
attempt to fix issue #1170

swiffer:
for me this works fine, seems something got not copied when migrated from PDOAdapter to AbstractAdapter!

Perhaps additional checks are needed but my use cases fork fine again!


lorenzo:
Thanks!

PR #1175, opened 2017-09-08
Add custom column support

SeinopSys:
This is a second attempt at achieving what #1144 couldn't.

This PR simply adds an `addCustomColumn` method to `Table` that skips the column type validation via the use of a `CustomColumn` class that's just a copy of the regular `Column` class for distinction.

I only have PostgreSQL installed so this might need testing in MySQL & others, but I can confirm it's working in Postgres and theoretically it should work everywhere else too.


lorenzo:
This will take me a couple days to review and test locally, but it is looking good!

SeinopSys:
One potential issue is that because of the parent-child relationship `addColumn` will "accept" a `CustomColor` instance as the first parameter, should I add in a check to prevent that?

SeinopSys:
@lorenzo Still plan on reviewing & testing this?

lorenzo:
Yes, sorry for the delay. I took a long holiday

chinpei215:
I think this is a really great enhancement. I was thinking of adding the same feature. If we have this feature, people would no longer need to send pull requests, such as #1204,  #1248, #1251, and #1254, unless they want abstraction.

By the way, what about we name the method `addNativeColumn()` instead `addCustomColumn()`? For example, Doctrine has `createNativeQuery()` method.

SeinopSys:
@lorenzo @chinpei215 Please see my updated approach. I've removed `addCustomColumn` and made a `Literal` class to use with the original `addColumn` method. I think this is a much cleaner approach overall that's easy to understand and work with.

chinpei215:
@SeinopSys Thank you. Sounds like a good approach.

lorenzo:
I'm very sorry for how long this took to merge, Life happened

lorenzo:
Thanks for your contribution!

SeinopSys:
No problem.

corpsee:
@SeinopSys @lorenzo This request breaks tests. See: https://travis-ci.org/cakephp/phinx/jobs/372687856.

SeinopSys:
@corpsee I find that odd considering the [last build of this PR](https://travis-ci.org/cakephp/phinx/jobs/334149431) passed before. I've no clue why it broke just now.

SeinopSys:
Found a fix, but I cannot push here anymore.

```
diff --git a/src/Phinx/Db/Adapter/PostgresAdapter.php b/src/Phinx/Db/Adapter/PostgresAdapter.php
index 60329d9..54ffa16 100644
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -362,8 +362,7 @@ public function getColumns($tableName)
                 $column->setLimit($columnInfo['character_maximum_length']);
             }
 
-            $phinxType = $this->getPhinxType($columnInfo['data_type']);
-            if (in_array($phinxType, [static::PHINX_TYPE_TIME, static::PHINX_TYPE_DATETIME])) {
+            if (in_array($columnType, [static::PHINX_TYPE_TIME, static::PHINX_TYPE_DATETIME])) {
             } else {
~
~
```

corpsee:
@SeinopSys thank you for the details.

lorenzo:
Can you open a new PR?

SeinopSys:
@lorenzo #1340

PR #1173, opened 2017-09-08
fixing configuration item default_migration_table

miranovy:
Fixing configuration item default_migration_table

PR #1176, opened 2017-09-08
Add tests for connection and table name configuration.

ndm2:
Adds tests for #1173, #1171, #1167.

Should this maybe be moved to, or have additional integration tests on `EnvironmentTest` class level? `EnvironmentTest::testGetAdapterWithExistingPdoInstance()` seems to have a hint of testing that could have covered #1170, but actually doesn't touch the underlying adapter to check whether it received the expected connection.

Edit: Didn't noticed the whitespace removal... I blame JetBrains.


lorenzo:
I think this is a good stopgap for now. Having more integration tests for `EnvironmentTest` would be ideal.

Thanks a lot!

PR #1190, opened 2017-09-19
Missing function in AdapterWrapper

khayyam90:
Due to errors encountered while using getConnection() from migration, I added the missing function in the AdapterWrapper


lorenzo:
Thanks!

Issue #1098, opened 2017-05-23
SQL Server adapter ignores foreign key 'constraint' key for naming


NightWulfe:
Per #92, you can add the 'constraint' array key to give a name to a foreign key constraint starting with Phinx 0.6.5. This key is ignored by the SQL Server Adapter.  Even if the the 'constraint' array key is set, the generated constraint name is the default name generated by Phinx.

Issue #92, opened 2013-05-28
Support for named foreign key constraint in MySQL adapter


tomzx:
Hi,

Based on the little experience I have with Phinx, it seems impossible to specify the name of the foreign key I want when creating a new foreign key. Based on the code in robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php, either you have a foreign key with a constraint name OR a foreign key with an automatically generated name.

See https://gist.github.com/tomzx/5663921/revisions for a "potential" quick fix.

wrah:
Hi, Is this issue still up for grabs ? I would love to do the necessary to see it merged.

ehowardtillit:
Hi, Is this issue still up for grabs ? I would love to do the necessary to see it merged.

blaind:
+1

nkoo:
+1

Yess11:
+1

Yess11:
Hello, what's the plan with this feature?

jdenoc:
It looks like you can set a foreign key name using the `constraint` option.

```php
$table = $this->table('your_table');
$table->addForeignKey('foreign_id', 'reference_table', array('id'), array('constraint'=>'your_foreign_key_name'));
$table->save();
```


This is valid as of
```
$ vendor/bin/phinx --version
Phinx by Rob Morgan - https://phinx.org. version 0.6.5
```

othercorey:
Constraint names are supported through the `constraint` option.

PR #1192, opened 2017-09-21
Support Foreign Key constraint names in SQL Server and PostgreSQL

NightWulfe:
Per #92, the 'constraint' option was added for naming foreign key constraints. This ability was not implemented in SQL Server or PostgreSQL (but was in SQLite). I've added support and updated the tests accordingly.

This is related to my issue report #1098.


lorenzo:
Thanks!

PR #1198, opened 2017-10-04
Fix chainability for the setColumns call

Johnnyshaun:
Table->setColumns() was not returning $this despite the documentation and thus was not chainable.

It's fixed :-)


dereuromark:
Thank you!

PR #1215, opened 2017-10-16
Simplify getSqlType types with default name

ntzm:
Also moved grouped cases to the top of the switch statement


dereuromark:
Let's hope this doesnt cause any regressions due to typos.

PR #1216, opened 2017-10-16
Add more files to gitattributes export-ignore

ntzm:
This might be a bit too strict for you, but it's what seems to be non-essential for package installation.

(Sorry for spamming PRs, let me know if you want me to stop 😛)


dereuromark:
The Upgrade files at least should probably stay.

ntzm:
Wasn't sure about those files, thanks for clarifying :) Updated

PR #1219, opened 2017-10-19
Revert back to original documentation

JayPHP:
Revert back to Phinx documentation until the cake documentation is sorted out?

IMO, the Phinx documentation should be standalone and not be included alongside all of the CakePHP documentation. The Cake documentation also doesn't have a phinx-specific search, which makes it much more difficult to find what you're looking for. 

Is there a specific reason why we can't stick to the original Phinx site anyway?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 Phinx makes it ridiculously easy to manage the database migrations for your PHP app. In less than 5 minutes, you can install Phinx and create your first database migration. Phinx is just about migrations without all the bloat of a database ORM system or framework.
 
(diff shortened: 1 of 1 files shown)


lorenzo:
I think we can just merge the whole sphinx documentation into a single page, so you can search the page using your browser search function.

One of the reason to have it in the same place is (apart that it is easier to deploy) that in the newer future we will be using the cake database layer. So looking for its documentation will make sense there as well.

Do you think having the whole docs in a single page would help with the search?

JayPHP:
@lorenzo Yes, single page could work as a solution - if you're happy with that? Another issue for me is the left nav - https://book.cakephp.org/3.0/en/phinx.html - all related to Cake framework. Isn't helpful and is cluttered for users who do not use Cake

lorenzo:
@JayPHP we are considering having a different book installation for each "plugin" in cake. So I guess sphinx will be covered by that soon. For the time being a single page doc is the way to go imo. Would you like to do that?

JayPHP:
no worries

Issue #649, opened 2015-09-26
Adding a boolean column with "limit" causes an SQL error in Postgresql


steinkel:
Hi,

```
            ->addColumn('is_superuser', 'boolean', [
                'default' => 0,
                'limit' => 1,
                'null' => false,
            ])
```

Is causing an error in Postgresql 9.3

SYNTAX error near "(" in BOOLEAN (1)

Would make sense to detect boolean type and ignore the limit option in this case?

Thank you!

dereuromark:
That makes sense IMO
Are you able to provide a pull request with your suggested changes?

othercorey:
Added in https://github.com/cakephp/phinx/pull/1222

PR #1222, opened 2017-10-22
refs #649 ignore limit in bool columns and cast to bool defaults for postgres

steinkel:
Make column definition less error prone when switching from mysql or baked code.

* Ignore limit in boolean columns
* Cast to boolean value the defaults if column type is boolean


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -875,12 +875,14 @@ public function dropDatabase($name)
      * @param  mixed $default
      * @return string
```
ntzm: Needs adding to the docblock

dereuromark: For return statements? I thought only params do.

ntzm: A new param was added

steinkel: fixed docblock

lorenzo:
Thanks!

PR #1224, opened 2017-10-24
Added support for symfony 4 components.

### LINKED ISSUE #1223: Compatibility with symfony/console 4.0
Can we add support for symfony/console 4.0? It should be the same as 3.4 but with deprecations removed. I'm not sure if there are any such compatibility issues, but I doubt many if any.

Currently phinx cannot be used in Symfony 4 due to this version constraint.

nesl247:
This required adding quotes around `%` in the yaml files. See http://symfony.com/blog/new-in-symfony-3-1-yaml-deprecations for more info on that.

In order to support v4, I tried to find all the places that had `%` unquoted and added them. I'm not sure if there should be another PR to have handled that or not, as I think users should be warned.

Fixes #1223

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,9 +25,9 @@
     }],
     "require": {
         "php": ">=5.4",
-        "symfony/console": "~2.8|~3.0",
-        "symfony/config": "~2.8|~3.0",
-        "symfony/yaml": "~2.8|~3.0"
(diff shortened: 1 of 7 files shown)


lorenzo:
Thanks a lot!

kornrunner:
Any chance of releasing this as a patch? Thanks!

PR #1225, opened 2017-10-24
Fix CS for test

Iandenh:
It looks like that the test just works after removing the return.

The Appveyor job is also green: https://ci.appveyor.com/project/Iandenh/phinx/build/1.0.7


ravage84:
Thanks

PR #1181, opened 2017-09-12
Fatal: hasOption() on null

ehimen:
Hi. Not sure if I'm missing something, but we've encountered an issue when instantiating a migration environment.

Sample code:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$manager = new \Phinx\Migration\Manager(
    \Phinx\Config\Config::fromPhp(__DIR__ . '/test-config.php'),
    new \Symfony\Component\Console\Input\ArrayInput([]),
    new \Symfony\Component\Console\Output\NullOutput()
);

class TestSeed extends \Phinx\Seed\AbstractSeed { }

$manager->executeSeed('development', new TestSeed());
```

Where our test configuration is:
```php
<?php

return [
    'environments' => [
        'development' => [
            'name' => 'devdb',
            'connection' => new PDO('sqlite::memory:'),
        ]
    ]
];
```

We encounter this error:
```
Fatal error: Uncaught Error: Call to a member function hasOption() on null in phinx/src/Phinx/Db/Adapter/AbstractAdapter.php on line 248

Error: Call to a member function hasOption() on null in phinx/src/Phinx/Db/Adapter/AbstractAdapter.php on line 248

Call Stack:
    0.0002     371912   1. {main}() phinx/pop.php:0
    0.0030     996640   2. Phinx\Migration\Manager->executeSeed() phinx/pop.php:13
    0.0032    1035552   3. Phinx\Migration\Manager\Environment->executeSeed() phinx/src/Phinx/Migration/Manager.php:396
    0.0032    1035552   4. Phinx\Migration\Manager\Environment->getAdapter() phinx/src/Phinx/Migration/Manager/Environment.php:148
    0.0033    1049152   5. Phinx\Db\Adapter\AdapterFactory->getAdapter() phinx/src/Phinx/Migration/Manager/Environment.php:344
    0.0044    1349128   6. Phinx\Db\Adapter\AbstractAdapter->__construct() phinx/src/Phinx/Db/Adapter/AdapterFactory.php:131
    0.0044    1349128   7. Phinx\Db\Adapter\PdoAdapter->setOptions() phinx/src/Phinx/Db/Adapter/AbstractAdapter.php:71
    0.0044    1349128   8. Phinx\Db\Adapter\PdoAdapter->setConnection() phinx/src/Phinx/Db/Adapter/PdoAdapter.php:56
[...]

PR #1231, opened 2017-11-01
Prevent PDO executing from constructor without input

ElisDN:
In addition to #1181.

By default `AdapterFactory` pass our `$options` to adapter's constructor:

```php
abstract class AbstractAdapter implements AdapterInterface
{
    public function __construct(array $options, InputInterface $input = null, OutputInterface $output = null)
    {
        $this->setOptions($options);
        ...
    }
}
```

but without `$input` and `$output`:

```php
class AdapterFactory
{
    public function getAdapter($name, array $options)
    {
        $class = $this->getClass($name);
        return new $class($options);
    }
}
```

And when we manually pass `connection` from configuration, the `PdoAdapter` immediately calls `setConnection` with creating of schema table:

```php
abstract class PdoAdapter extends AbstractAdapter
{
    public function setOptions(array $options)
    {
        parent::setOptions($options);
        if (isset($options['connection'])) {
            $this->setConnection($options['connection']);
        }
        return $this;
    }

    public function setConnection(\PDO $connection)
    {
        $this->connection = $connection;
        if (!$this->hasSchemaTable()) {
            $this->createSchemaTable();
        } ...
        return $this;
    }
}
```

with calling of `$this->execute($sql)`, but `$this->getInput()` is still empty (it was a reason of #1181 error).

As a solution I propose to move `$this->setConnection($options['connection'])` call to `getConnection()` for lazy loading (as in this PR) or make `$input` and `$output` as required arguments without setters and pass them via factory:

```php
class AdapterFactory
{
    public function getAdapter($name, array $options, InputInterface $input, OutputInterface $output)
    {
        $class = $this->getClass($name);
        return new $class($options, $input, $output);
    }
}
```

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -45,20 +45,6 @@ abstract class PdoAdapter extends AbstractAdapter
     protected $connection;
 
     /**
-     * {@inheritdoc}
-     */
-    public function setOptions(array $options)
-    {
-        parent::setOptions($options);
-
-        if (isset($options['connection'])) {
-            $this->setConnection($options['connection']);
-        }
-
-        return $this;
-    }
-
-    /**
      * Sets the database connection.
      *
      * @param \PDO $connection Connection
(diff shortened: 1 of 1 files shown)


chinpei215:
We are sorry for the late reply. Could you please provide some test cases?

dereuromark:
ping

PR #1234, opened 2017-11-03
Revert renaming the PDO adapter test file.

ndm2:
Unit test files must end with the `Test` suffix, otherwise they're not
being run.

refs #1181 

This was also the reason for the #1231 testrun to not fail.


lorenzo:
Thanks

PR #1236, opened 2017-11-05
Adding support for dotenv using the -c config file option

phillipsharring:
#### Idea

1. josegonzalez/dotenv is required from composer.json. The 2.* version was preferred to have consistency with the cakephp/app composer requirements.
2. And .env.example file is included with example config variables. This would be copied to .env and edited by the developer
3. A phinx.config.php.dist file is included with an example config array. This would be copied to phinx.config.php and edited by the developer to include config arrays for various environments, along with the default database config. This file references the getenv() function, so the values are read from .env. The array keys in this config file are the same as in `phinx.yml`.
4. .gitignore has been updated to include .env and *.config.php files. The name .env is a standard name for the dotenv package. The phinx.config.php file could be named and placed anywhere. The .config.php extension is just a suggestion to make ignoring the config files easier. This file could be phinx.config.php in the root, or `configs/phinx.config.php` or anywhere, really.

#### Usage

When running phinx at the command line, the -c or --configuration flag is used and the value is pointed at the phinx.config.php file, like so:

```bash
$ ./bin/phinx migrate  -e development -c phinx.config.php
```

The `josegonzalez\Dotenv\Loader` class is only loaded in the configuration file, so if the configuration file is not used, that class won't be loaded.

Again, the naming of the config file is merely a suggestion.

**Note:**  Correct me if I'm wrong, but I don't think this needs any tests, since loading a config file is already tested. If it needs tests, I'll add them.


composer.json
```diff
@@ -27,7 +27,8 @@
         "php": ">=5.4",
         "symfony/console": "~2.8|~3.0|~4.0",
         "symfony/config": "~2.8|~3.0|~4.0",
-        "symfony/yaml": "~2.8|~3.0|~4.0"
+        "symfony/yaml": "~2.8|~3.0|~4.0",
+        "vlucas/phpdotenv": "^2.4"
```
ADmad: The cakephp skeleton [app](https://github.com/cakephp/app/blob/master/composer.json#L12) uses `josegonzalez/dotenv` for providing dotenv support. While phinx is standalone lib it would be better if projects under the cakephp family used same libs when possible.

philsown: Good idea. I didn't know that. I'll change this to use the `josegonzalez/dotenv` library soon today.

composer.json
```diff
@@ -28,7 +28,7 @@
         "symfony/console": "~2.8|~3.0|~4.0",
         "symfony/config": "~2.8|~3.0|~4.0",
         "symfony/yaml": "~2.8|~3.0|~4.0",
-        "vlucas/phpdotenv": "^2.4"
+        "josegonzalez/dotenv": "2.*"
```
dereuromark: Why not semver?

composer.json
philsown: I made this change. Went with "2.*" to be consistent with cakephp.

composer.json
philsown: I picked this version to be consistent with the version used in cakephp/app.

chinpei215:
Relates to #1085

dereuromark:
Looks like we can move forward once the conflict is resolved :+1:

.gitignore
```diff
@@ -36,3 +36,9 @@ phpunit.xml
 
 # sqlite test database
 phinx_testing.sqlite3
+
+# ignore dotenv .env file
+.env
```
dereuromark: why is this needed? shouldnt this only be relevant for project's config files, not for inside vendor?

composer.json
```diff
@@ -28,7 +28,8 @@
         "php": ">=5.4",
         "symfony/console": "^2.8|^3.0|^4.0",
         "symfony/config": "^2.8|^3.0|^4.0",
-        "symfony/yaml": "^2.8|^3.0|^4.0"
+        "symfony/yaml": "^2.8|^3.0|^4.0",
+        "josegonzalez/dotenv": "2.*"
```
dereuromark: Does anyone think this should be require-dev as optional one for those that do not use/want it?

.gitignore
philsown: The dotenv file (which is a config file) is typically placed in the root, and is ignored, because it will most likely be different in every environment an application is deployed to.

The .env.example is committed to give people a starting point. The naming of this file is also fairly standard across various project.

I'm not sure what you're asking about the vendor directory.

composer.json
philsown: The migrations will probably be run in every environment, not just while developing, right? In which case, I think this should be a requirement, not a dev requirement. Just my 2 cents but I maybe wrong in my understanding.

dereuromark: A dev requirement (along with suggest text) means that the functionality is there but dormant.
Once you require it in your root level (project application), it then becomes a real functionality.
So no, this should be up to the dev IMO. Along with a documentation on how to enable this of course.

philsown: OK :)

.gitignore
dereuromark: vendor/cakephp/phinx/.gitignore => this file
What does that relate to .env files that are in root dir?

phinx.config.php.dist
```diff
@@ -0,0 +1,38 @@
+<?php
+
+// copy this file to phinx.config.php, or your preferred name/location
```
dereuromark: Should we add the note here? `composer require ... ?`

composer.json
philsown: Changed.

.gitignore
dereuromark: Sounds good.

phinx.config.php.dist
```diff
@@ -0,0 +1,38 @@
...
+
+// return the configuration
+return [
+    'paths' => [
+        'migrations' => __DIR__ . '/migrations',
+    ],
+    'environments' => [
+        'default_migration_table' => 'phinxlog',
+        'default_database' => getenv('APP_ENV'),
+        'develop' => [
```
josephrace: I think this should match the `APP_ENV` value in .env.example - "development"

philsown: Good catch. Thanks!

othercorey:
Closing due to no activity.

PR #1238, opened 2017-11-11
Support PHPUnit 6

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06) and [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/yaml": "~2.8|~3.0|~4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.26|^5.0",
+        "phpunit/phpunit": "^4.8.35|^5.7|^6.5",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
     "autoload": {
--- a/tests/Phinx/Config/AbstractConfigTest.php
+++ b/tests/Phinx/Config/AbstractConfigTest.php
@@ -2,13 +2,15 @@
 
 namespace Test\Phinx\Config;
 
+use PHPUnit\Framework\TestCase;
+
 /**
  * Class AbstractConfigTest
  * @package Test\Phinx\Config
  * @group config
  * @coversNothing
  */
-abstract class AbstractConfigTest extends \PHPUnit_Framework_TestCase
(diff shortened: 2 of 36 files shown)


dereuromark:
The leading \ for the other use statements is wrong and should be removed.

carusogabriel:
I've added PHPUnit 6 support. This will help us test against PHP 7.2, that the `each` method was deprecated.

I added a `setExpectedException` method layer, because [it was removed in PHPUnit 6](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#removed).

composer.json
```diff
@@ -30,7 +30,7 @@
         "symfony/yaml": "~2.8|~3.0|~4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.26|^5.0",
+        "phpunit/phpunit": "^4.8.35|^5.7|^6.5",
```
dereuromark: Is it easier at some point if we would stop supporting completely outdated php version 5.4 and e.g. use 5.6+ instead?
Then we could drop 4.8 here probably at least.

carusogabriel: It's up to you. I've made everything to support all `PHP` and `PHPUnit` versions. But, I can open other PR dropping PHP `5.4` and `5.5` support :man_technologist:

dereuromark: For now its cool then :)

carusogabriel: Gonna work on stop supporting for PHP `5.4` and `5.5` versions soo.

dereuromark: I meant you dont have to. If all works fine :)

carusogabriel: Ahh, okay, sorry :sweat_smile:

PR #1241, opened 2017-11-24
Multiple databases

sincilite:
In response to https://github.com/cakephp/phinx/issues/180 and my own requirement to manage multiple databases for a single application I've implemented the ability to specify multiple database connections within the Phinx config and within many environments and have each track it's own migrations.

The existing functionality remains and it is still possible to use the original configuration schema of environments and single database connections as child properties.  To ensure backwards compatibility the root paths options must remain in config even when using multiple databases with their own migration and seed paths specified.

```
development:
        db_ref_1:
            paths:
                migrations: '%%PHINX_CONFIG_DIR%%/db/migrations/mysql'
                seeds: '%%PHINX_CONFIG_DIR%%/db/seeds/mysql' 
            adapter: mysql
            host: 127.0.0.1
            name: development_db
            user: root
            pass: ''
            port: 3306
            charset: utf8

        db_ref_2:
            paths:
                migrations: '%%PHINX_CONFIG_DIR%%/db/migrations/mysql-2'
                seeds: '%%PHINX_CONFIG_DIR%%/db/seeds/mysql-2' 
            adapter: mysql
            host: 127.0.0.1
            name: development_db
            user: root
            pass: ''
            port: 3307
            charset: utf8
```
Migrations and seed files are separated in different directories and can still utilise namespacing if required.

The command seeds:run has been updated to allow an optional -d or --database option with the expected value being the database reference specified in the config (e.g. db_ref_1)


dereuromark:
Conflicts
This will add quite the complexity to the system - but if done in a good way I think it would be acceptable.

PR #1243, opened 2017-11-25
Added support for "signed" option when using custom name for "id" column

rickhub:
When creating custom id columns (e.g. `["id" => "custom_id"]`) the `signed` option got totally ignored. Example:

    // Note: it's important to set a custom name for the "id" column to reproduce this
    $this->table('foo', [ 'id' => 'custom_id', 'signed' => false ])->save();

This statement created a table called "foo" with an `id` column of type `int(11)` (signed!), which is obviously wrong. This commit fixes this behavior so the created column will be of type `int(11) unsigned`. Additionally I simplified the code a bit.

---

The method `MysqlAdapter::getPhinxType()` totally ignored the `signed` attribute, too. While writing unit tests for the fix mentioned above, I found out that it is not possible to check whetever a column's type is *really* signed or not, when reading the columns through `MysqlAdapter::getColumns()`. This is because

  a) `MysqlAdapter::getColumns()` doesn't set the signed attribute at all (e.g. with `setSigned()`)
  b) `MysqlAdapter::getPhinxType()` doesn't return the information whetever the column's type is signed or not.

That's why this call:

    $this->adapter->getColumns('ntable')[0];

returned:

    class Phinx\Db\Table\Column{
        // ..
        protected $signed => bool(true) // (!)
        // ..
    }

Even if the column actually was an `int(11) unsigned` in the database.

To fix this, this commit added the `signed` property to the return value of `getPhinxType()`, so `getColumns()` can now make use of it to provide accurate informations to the caller.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -224,6 +224,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $column = new Column();
             $column->setName($options['id'])
                    ->setType('integer')
+                   ->setSigned(isset($options['signed']) ? $options['signed'] : true)
                    ->setIdentity(true);
 
             array_unshift($columns, $column);
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -211,7 +211,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
         // Add the default primary key
         $columns = $table->getPendingColumns();
 
-        if(!isset($options['id']) || $options['id'] !== false){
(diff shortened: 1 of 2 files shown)


rickhub:
There's some work required to make all the unit tests pass where `getColumns()` etc. is involved. Also `getPhinxType()` could need some optimizations (e.g. `signed` should be `null` or totally omitted for string types etc.) Maybe I'll find some time during the next week to make this happen.

Any feedback on this?

chinpei215:
Regarding failure 1, it needs to be fixed. Regarding failure 2 ~ 4, I think you can change `$expected`.

Deadpikle:
@rickhub Are you going to finish this PR? If not, @chinpei215 / others, can I submit a new PR with the bugfix?

Thanks. :)

dereuromark:
Any update on this?

Deadpikle:
Shouldn't this PR be closed based on #1294 being merged?

PR #1244, opened 2017-11-26
Add ordered seeds feature

emptyhand:
This feature is for ordering seeds execution by seeds dependencies. You can define other seeds which are dependencies for your seed and change seeds execution order automatically.

This feature is useful when you extend phinx manager (for example in unit tests to use in memory database) and do not have possibility execute seeds command from terminal.

Also if you have a lot of seeds it is not convenient to order seeds writing one by one in terminal to run ordered seeds.

Just run `phinx seeds:run` and your seeds will be executed in ordered way.

Use this function to define dependencies:
```
public function getDependencies()
{
    return [
        'UserSeeder',
        'GSeeder',
    ];
}
```

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -769,6 +769,54 @@ class Manager
     }
 
     /**
+     * Get seed dependencies instances from seed dependency array
+     *
+     * @param AbstractSeed $seed Seed
+     *
+     * @return AbstractSeed[]
+     */
+    private function getSeedDependenciesInstances(AbstractSeed $seed)
+    {
+        $dependenciesInstances = [];
+        $dependencies = $seed->getDependencies();
+        if (!empty($dependencies)) {
+            foreach ($dependencies as $dependency) {
+                foreach ($this->seeds as $seed) {
+                    if (get_class($seed) === $dependency) {
+                        $dependenciesInstances[get_class($seed)] = $seed;
+                    }
+                }
+            }
+        }
+
+        return $dependenciesInstances;
+    }
+
+    /**
+     * Order seeds by dependencies
+     *
+     * @param AbstractSeed[] $seeds Seeds
+     *
+     * @return AbstractSeed[]
+     */
+    private function orderSeedsByDependencies(array $seeds)
+    {
+        $orderedSeeds = [];
+        foreach ($seeds as $seed) {
+            $key = get_class($seed);
+            $dependencies = $this->getSeedDependenciesInstances($seed);
+            if (!empty($dependencies)) {
+                $orderedSeeds[$key] = $seed;
+                $orderedSeeds = array_merge($this->orderSeedsByDependencies($dependencies), $orderedSeeds);
+            } else {
+                $orderedSeeds[$key] = $seed;
+            }
+        }
+
+        return $orderedSeeds;
+    }
+
+    /**
      * Gets an array of database seeders.
      *
      * @throws \InvalidArgumentException
@@ -823,6 +871,7 @@ class Manager
             $this->setSeeds($seeds);
         }
 
+        $this->seeds = $this->orderSeedsByDependencies($this->seeds);
         return $this->seeds;
     }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
 script:
     - if [[ $DEFAULT = 1 ]]; then vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover; fi
     - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.8 && vendor/bin/phpstan analyse -c phpstan.neon -l 4 src; fi
-    - if [[ $PHPCS = 1 ]]; then composer require --dev squizlabs/php_codesniffer:^3.0 && vendor/bin/phpcs -p -s src; fi
+    - if [[ $PHPCS = 1 ]]; then composer require --dev squizlabs/php_codesniffer:~3.1.1 && vendor/bin/phpcs -p -s src; fi
 
 after_success:
   - if [[ $DEFAULT = 1 ]]; then bash <(curl -s https://codecov.io/bash); fi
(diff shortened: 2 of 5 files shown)


src/Phinx/Migration/Manager.php
```diff
@@ -768,6 +769,56 @@ public function setSeeds(array $seeds)
...
+                }
+            }
+        }
+        return $dependenciesInstances;
+    }
+
+    /**
+     * Order seeds by dependencies
+     *
+     * @param AbstractSeed[] $seeds
```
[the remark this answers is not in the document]
dereuromark: Also FQCN please inside comments.

src/Phinx/Migration/Manager.php
```diff
@@ -768,6 +769,58 @@ public function setSeeds(array $seeds)
...
+    {
+        if (empty($this->seeds)) {
+            return [];
+        }
+
+        $dependenciesInstances = [];
+        $dependencies = $seed->getDependencies();
+        foreach ($dependencies as $dependency) {
+            foreach ($this->seeds as $seed) {
+                if (get_class($seed) === $dependency) {
```
chinpei215: Sorry if I am wrong, but if people make a typo in the etDependencies() method, it will be discarded silently. Couldn't it be implemented by using `usort()` and the callback method? I think you want to sort seeds by a custom logic. In such case, I would use `usort()`.

src/Phinx/Seed/OrderedSeedInterface.php
```diff
@@ -0,0 +1,42 @@
...
+ * Ordered seed interface
+ * Use this class if you want to define your seeds dependencies
+ * and run seeds in ordered way without using command line options
+ */
+interface OrderedSeedInterface
+{
+    /**
+     * @return array
+     */
+    public function getDependencies();
```
chinpei215: Why do we need an additional interface? I think we can add some additional method to the `AbstractSeed` class with a default implementation.

src/Phinx/Migration/Manager.php
emptyhand: I can add exception if there is no dependency written in the function. About sorting seeds in custom logic, I can not imagine a case where it would be used. Why you want to order seeds in custom logic? And why write additional code if there is simple and suitable solution with dependencies...

src/Phinx/Seed/OrderedSeedInterface.php
emptyhand: Of course we can add method to AbstractSeed. I did interface if some people do not want to order they seeds and then they keep them without this interface and without getDependencies() function.
If there is a need to move to AbstractSeed class I can do it.

src/Phinx/Migration/Manager.php
chinpei215: Sorry for my English. I didn't say I wanted to order seeds in custom logic. I meant that your implementation looks like that it sorts seeds by using two methods - `orderSeedsByDependencies()` and `getSeedDependenciesInstances()`. But if you have a method to compare seeds, you can do `usort($seeds, [$this, 'compareSeeds'])`.

Also, people cannot use `::class` syntax to avoid typos. Try `NonExisitentClass::class`. It would also returns `NonExisitentClass`.

emptyhand: It could be implemented in many ways :). Please write me explicitly what should I fix, to merge this feature (I need it in production urgently...).

chinpei215: OK. Then please don't add additional interface. Other points are minor things. It could be changed later if someone interests.

emptyhand:
@chinpei215

src/Phinx/Migration/Manager.php
```diff
@@ -768,6 +768,58 @@ public function setSeeds(array $seeds)
...
     }
 
+    /**
+     * Get seed dependencies instances from seed dependency array
+     *
+     * @param AbstractSeed $seed Seed
+     *
+     * @return AbstractSeed[]
+     */
+    protected function getSeedDependenciesInstances(AbstractSeed $seed)
```
chinpei215: Could you please change `protected` to `private`.

src/Phinx/Migration/Manager.php
```diff
@@ -768,6 +768,58 @@ public function setSeeds(array $seeds)
...
+    {
+        if (empty($this->seeds)) {
+            return [];
+        }
```
chinpei215: This check seems unnecessary. If `$this->seeds` is an empty array, I think this method is not called.

src/Phinx/Migration/Manager.php
```diff
@@ -768,6 +768,58 @@ public function setSeeds(array $seeds)
...
+    }
+
+    /**
+     * Order seeds by dependencies
+     *
+     * @param AbstractSeed[] $seeds Seeds
+     *
+     * @return AbstractSeed[]
+     */
+    protected function orderSeedsByDependencies(array $seeds)
```
chinpei215: Could you please change protected to private.

emptyhand:
@chinpei215 if everything is ok, could you merge now those changes?

chinpei215:
@emptyhand Merged. Thank you for your contribution.

Shardj:
Hey does this actually work? When I run my seeder with dependencies it immediately goes into run() and doesn't run the dependency seeders first

Shardj:
I'm going to xdebug through on monday and see if I can find out why it isn't running these. Manager getSeedDependenciesInstances seems to correctly find them and return them. It does run twice though for some reason, first instance nothing is passed in, presumably it's using a non-extended instance of AbstractSeed then.

emptyhand:
Yes it works for me well. You have to run just console command ` php vendor/bin/phinx seed:run`
without adding any seeds, then dependencies will be executed automatically

Shardj:
Definitely not my experience so far, running the seeder with the dependencies specifically jumps straight to run, and running all seeders just executes them in alphanumeric order and when it reaches the file in question it ignores the dependencies. I'll try to find out why when I'm back at work on Monday, I'll make a pull request if it's a flaw and not something idiotic I did (I suspect namespaces and bad backslash encoding).

Shardj:
No idea why but when I came in today it was working fine, I probably did something stupid, or it's standard PHP wizardry.

PR #1246, opened 2017-11-28
Added saving column's comment for PostgresAdapter when you add column

Rymmugygr:


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -357,12 +357,16 @@ public function hasColumn($tableName, $columnName)
...
         $sql = sprintf(
-            'ALTER TABLE %s ADD %s %s',
+            'ALTER TABLE %s ADD %s %s;',
             $this->quoteTableName($table->getName()),
             $this->quoteColumnName($column->getName()),
             $this->getColumnSqlDefinition($column)
         );
 
+        if ($column->getComment()) {
+            $sql .= $this->getColumnCommentSqlDefinition($column, $table->getName());
```
chinpei215: I think this should be done in the `getColumnSqlDefinition()` method, like `MysqlAdapter`:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/MysqlAdapter.php#L1005-L1007
and `SQLiteAdapter`:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/SQLiteAdapter.php#L995

Rymmugygr: The problem is that PostgreSQL requires additional (separate) request for adding a comment for a column.

For example:
- MySQL: "ALTER TABLE books ADD COLUMN author int(10) unsigned NOT NULL COMMENT 'a new comment';"
- PostgreSQL: "ALTER TABLE books ADD COLUMN author int NOT NULL;" + "COMMENT ON COLUMN books.author IS 'a new comment';"

If we add this logic into getColumnSqlDefinition its behavior will be much different from the same methods in other adapters.

Comments treated this way in other methods, for example, createTable adds comments only after column definitions:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/PostgresAdapter.php#L201-L208

chinpei215: Thank you for the explanation. You are right.

PR #1247, opened 2017-11-30
Fix overriding mysql default options by environment options

khayyam90:
I encountered a bug when creating a table with table()->create() : the charset used in the CREATE TABLE sentence was the hardcoded value "utf8_general_ci" even if I defined a charset in my environment file.

In this PR, the default hardcoded options are overridden by the environment options and the environemnt options are override by the table options is specified in the migration (= one more level).

I have fixed it in the MysqlAdapter. I checked in the other adapters if the charset was used in the table creation and it is not use anywhere else.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -710,7 +710,7 @@ limit   set maximum length for strings, also hints column types in adapters (see
 length  alias for ``limit``
 default set default value or action
 null    allow ``NULL`` values (should not be used with primary keys!)
-after   specify the column that a new column should be placed after
+after   specify the column that a new column should be placed after *(only applies to MySQL)*
 comment set a text comment on the column
 ======= ===========
 
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -206,6 +206,10 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             'engine' => 'InnoDB',
             'collation' => 'utf8_general_ci'
         ];
+
+        // use the default options, overriden by the environment options, overriden by the table options (especialy for charset & collation)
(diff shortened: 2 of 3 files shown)


chinpei215:
Could you please provide test cases?

khayyam90:
I added a unit test for this feature

chinpei215:
Thank you for adding tests. I think the root cause of the issue is the following code:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/MysqlAdapter.php#L205-L209
This is incorrect to me, as MySQL supports DEFAULT CHARACTER SET and COLLATE in the CREATE DATABASE statement,and Phinx seems to use the statement if you set the `collation` option:
https://github.com/cakephp/phinx/blob/d84a6c7f4e6919614023b8d17a01992a125a3aae/src/Phinx/Db/Adapter/MysqlAdapter.php#L942-L944
Also, you could configure the `default_storage_engine` variable on the database server if you want to use other storage engines by default.

So what if we remove the `$defaultOptions` and `array_merge()` instead? I would like to hear what others say.

khayyam90:
I also think the hardcoded value "utf8_general_ci" is something wrong. 
If collation is not provided, it should use the default_storage_engine from the server, right. But if something is provided, why couldn't it be used as table's collation ?

chinpei215:
I don't say it couldn't be used. But I am not sure how it is useful. I thought it would work as you expected if there are no hardcoded default settings and if your database collation is set properly. So I want to hear what other people say. If they say it is useful, then I think we should consider adding the same feature to other adapters too. (I don't say you should do it in this PR though.)

chinpei215:
Closing, as this PR seems to duplicate to #1169. I didn't notice the duplicated PR has been already sent. Sorry.

Issue #1106, opened 2017-06-02
Feature Request: MySQL BIT type


shbmira:
BIT Datatype Doc: https://dev.mysql.com/doc/refman/5.7/en/bit-type.html

When I got the error message `"bit" is not supported`, I added it myself. So it goes:

```diff
diff --git a/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterInterface.php b/app/vendor/robmorgan/phinx/src/Phinindex 1201e0b..4fd85a9 100644
--- a/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -46,6 +46,7 @@ interface AdapterInterface
     const PHINX_TYPE_STRING         = 'string';
     const PHINX_TYPE_CHAR           = 'char';
     const PHINX_TYPE_TEXT           = 'text';
+    const PHINX_TYPE_BIT            = 'bit';
     const PHINX_TYPE_INTEGER        = 'integer';
     const PHINX_TYPE_BIG_INTEGER    = 'biginteger';
     const PHINX_TYPE_FLOAT          = 'float';
```

```diff
diff --git a/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php b/app/vendor/robmorgan/phinx/src/Phinx/Dbindex 9c3b334..b97054d 100644
--- a/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/app/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -61,6 +61,8 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
     const INT_MEDIUM  = 16777215;
     const INT_REGULAR = 4294967295;
     const INT_BIG     = 18446744073709551615;
+
+    const BIT = 64;

     const TYPE_YEAR   = 'year';

@@ -793,6 +795,9 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
                 }
                 return array('name' => 'blob');
                 break;
+            case static::PHINX_TYPE_BIT:
+                return array('name' => 'bit', 'limit' => $limit ? $limit : 64);
+                break;
             case static::PHINX_TYPE_INTEGER:
[...]

PR #1248, opened 2017-12-01
Fix MySQL BIT Column Type (#1106)

shbmira:
https://github.com/cakephp/phinx/issues/1106


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -749,6 +751,8 @@ public function getSqlType($type, $limit = null)
                 }
 
                 return ['name' => 'blob'];
+            case static::PHINX_TYPE_BIT:
+                return ['name' => 'bit', 'limit' => $limit ? $limit : 64];
```
chinpei215: You could do `$limit ?: 64`.

shbmira: Of course, @chinpei215! Thank you!

dereuromark: is that available in 5.4?

shbmira: Yes, @dereuromark!
See: http://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -749,6 +751,8 @@ public function getSqlType($type, $limit = null)
                 }
 
                 return ['name' => 'blob'];
+            case static::PHINX_TYPE_BIT:
+                return ['name' => 'bit', 'limit' => $limit ?: 64];
```
chinpei215: Sorry for commenting again, but wouldn't it be better to use `1` instead `64` here? If you omit the size when you declare `bit` column in the MySQL, I think it will be `bit(1)`. Or would using maximum size is a usual thing in Phinx?

chinpei215:
Sorry if I am wrong, but file permissions seem to be changed 644 to 755 accidentally.

src/Phinx/Db/Adapter/MysqlAdapter.php
shbmira: Sorry, @chinpei215. Maybe I had a little misunderstanding here... But following STRING, CHAR, INTEGER definitions, I thoght this number was the maximum size (in bytes) associated with the type. 

And accordingly to MySQL documentation, BIT can vary from 1 to 64.

But I don't know now... What do you think?

And about file permissions (644). Is that right?

chinpei215: OK. Then I am OK that `$limit` defaults to 64.
Regarding file permissions, I think you can confirm it when you click the `Changes` tab of this PR. I think you can see 100644 → 100755 after the each file names.

dereuromark:
Yes, file permissions should be reverted.

shbmira:
Sorry, guys!
I'm veeeeeeeeeery late!
But I think it's alright now!

Could you take a look?

chinpei215:
Looks good to me.

src/Phinx/Db/Adapter/MysqlAdapter.php
caugner: FYI @chinpei215 @shbmira The options returned by `MySQLAdapter->getSqlType()` actually act as the default options.

This means:
1. `BIT` columns added by phinx will by default have a limit of 64, unless overridden by the developer.
2. A developer may still specify a limit > 64, but the migration will fail:
  `SQLSTATE[42000]: Syntax error or access violation: 1439 Display width out of range for column 'mybit' (max = 64)`

PR #1254, opened 2017-12-06
Added interval column type to PostgreSQL

ikagawa:
I often use interval type in PostgreSQL, so I need an interval type support.


chinpei215:
Thank you for your contribution. Could you please provide some test cases?

ikagawa:
I added test code for PostgreSQL interval type to existing test method.

chinpei215:
Thank you for adding tests. Looks good to me.

PR #1255, opened 2017-12-06
Improvement for composer.json

danielcosta:
Adding required PHP extensions to composer and suggesting PDO extensions for each optional adapter.

As explained on [composer's documentation here](https://getcomposer.org/doc/04-schema.md#package-links), we should require all dependencies on composer.json because not all systems have the common extensions installed by default:

> Note: It is important to list PHP extensions your project requires. Not all PHP installations are created equal: some may miss extensions you may consider as standard (such as ext-mysqli which is not installed by default in Fedora/CentOS minimal installation systems). Failure to list required PHP extensions may lead to a bad user experience: Composer will install your package without any errors but it will then fail at run-time. The composer show --platform command lists all PHP extensions available on your system. You may use it to help you compile the list of extensions you use and require. Alternatively you may use third party tools to analyze your project for the list of extensions used.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,9 @@
     }],
     "require": {
         "php": ">=5.4",
+        "ext-ctype": "*",
+        "ext-mbstring": "*",
+        "ext-pdo": "*",
         "symfony/console": "~2.8|~3.0|~4.0",
         "symfony/config": "~2.8|~3.0|~4.0",
         "symfony/yaml": "~2.8|~3.0|~4.0"
@@ -32,6 +35,12 @@
     "require-dev": {
         "phpunit/phpunit": "^4.8.26|^5.0",
         "cakephp/cakephp-codesniffer": "^3.0"
+    },
+    "suggest": {
+        "ext-pdo_mysql": "Needed to support MySQL adapter",
+        "ext-pdo_pgsql": "Needed to support PostgreSQL adapter",
+        "ext-pdo_sqlite": "Needed to support SQLite adapter",
+        "ext-pdo_sqlite": "Needed to support SQL Server/ SQL Azure adapter"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "ext-pdo_mysql": "Needed to support MySQL adapter",
         "ext-pdo_pgsql": "Needed to support PostgreSQL adapter",
         "ext-pdo_sqlite": "Needed to support SQLite adapter",
-        "ext-pdo_sqlite": "Needed to support SQL Server/ SQL Azure adapter"
+        "ext-pdo_sqlsrv": "Needed to support SQL Server/ SQL Azure adapter"
     },
     "autoload": {
         "psr-4": {


composer.json
```diff
@@ -33,6 +36,12 @@
         "phpunit/phpunit": "^4.8.26|^5.0",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
+    "suggest": {
+        "ext-pdo_mysql": "Needed to support MySQL adapter",
+        "ext-pdo_pgsql": "Needed to support PostgreSQL adapter",
+        "ext-pdo_sqlite": "Needed to support SQLite adapter",
+        "ext-pdo_sqlite": "Needed to support SQL Server/ SQL Azure adapter"
```
dereuromark: duplicate key here.

composer.json
```diff
@@ -25,6 +25,9 @@
     }],
     "require": {
         "php": ">=5.4",
+        "ext-ctype": "*",
```
dereuromark: ctype really needed for all db types?

danielcosta: It seems like this is actually a package dependency because when we try to create a new migration without Ctype the following error is shown:

```
/app $ vendor/bin/phinx create NewMigration
Phinx by CakePHP - https://phinx.org. 0.8.1


Fatal error: Uncaught Error: Call to undefined function Symfony\Component\Config\ctype_alpha() in /app/vendor/symfony/config/FileLocator.php:87
Stack trace:
#0 /app/vendor/symfony/config/FileLocator.php(42): Symfony\Component\Config\FileLocator->isAbsolutePath('phinx.php')
#1 /app/vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php(218): Symfony\Component\Config\FileLocator->locate('phinx.php', '/app', true)
#2 /app/vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php(236): Phinx\Console\Command\AbstractCommand->locateConfigFile(Object(Symfony\Component\Console\Input\ArgvInput))
#3 /app/vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php(93): Phinx\Console\Command\AbstractCommand->loadConfig(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#4 /app/vendor/robmorgan/phinx/src/Phinx/Console/Command/Create.php(154): Phinx\Console\Command\AbstractCommand->bootstrap(Object(Symfony\Component\Console\Input\ArgvInput), Objec in /app/vendor/symfony/config/FileLocator.php on line 87
```

composer.json
danielcosta: Fixed. Thank you

composer.json
ADmad: The PHP manual says ctype functions should be enabled by default http://php.net/manual/en/ctype.installation.php. Based on the comments only FreeBSD seems to have a problem.

dereuromark:
So are we moving forward with this?
It might break a few installs where this wasnt needed and still worked.
Suggest seems totally fine, the other is a bit to think about for the next patch release..

composer.json
dereuromark: So it would be safer to also put them into suggest for now IMO.

martijngastkemper:
Symfony Config seems to have solved this problem by adding a polyfill. https://github.com/symfony/config/blob/master/composer.json#L21

danielcosta:
> Symfony Config seems to have solved this problem by adding a polyfill. https://github.com/symfony/config/blob/master/composer.json#L21

Great! I'll close this issue since it's no longer relevant...

Issue #1013, opened 2017-01-09
MysqlAdapter: getColumns() doesn't return precision and scale for decimal


MarkoNV:
Migration:

```
$table = new Table('test', [], $this->adapter);
$table->addColumn('dec', 'decimal', array('precision' => 10, 'scale' => 2));
$table->save();
```
Decimal(10,2) column is successfully created, but:
`var_dump($table->getColumns()); `
returns:
```
array (size=2)
  0 => 
    object(Phinx\Db\Table\Column)[35]
      protected 'name' => string 'id' (length=2)
...
  1 => 
    object(Phinx\Db\Table\Column)[34]
      protected 'name' => string 'dec' (length=3)
      protected 'type' => string 'decimal' (length=7)
      protected 'limit' => **int 10**
      protected 'null' => boolean false
      protected 'default' => null
      protected 'identity' => boolean false
      protected 'precision' => **null**
      protected 'scale' => **null**
      protected 'after' => null
      protected 'update' => null
      protected 'comment' => null
      protected 'signed' => boolean true
      protected 'timezone' => boolean false
      protected 'properties' => 
        array (size=0)
          empty
      protected 'values' => null
```
_Since DECIMAL(M) is equivalent for DECIMAL(M,0), precision could be alias for limit, but scale is still missing. getPhinxType returns limit and scale, but it's not assigned to Column class if column type is decimal._ -> additional mistake, see edit.

Edit: Column definition for decimal is DECIMAL (precision, scale), but when parsing existing column data, it's parsed as DECIMAL (limit, precision). Same applies to SQLiteAdapter.

dereuromark:
Are you able to provide a pull request with your suggested changes?

ricardoapaes:
I have the same problem, has it already been resolved?

martijngastkemper:
PR #1258 doesn't solve the problem. I'm currently working on a PR to fix the parsing of `DECIMAL`. Hopefully no backwards compatibility problems 🤞

PR #1258, opened 2017-12-08
is_int check for precision & scale #1013

39mamon:
I found is column type is decimal bug. #1013 
Decimal options precision & scale is not reflected.

So I decimal options conditional judgment modify.

This bug at run cakephp3 migrations code.

```
        $table->changeColumn('item_code', 'decimal', [
            'precision' => 20,
            'scale' => 0,
        ]);
```

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -987,7 +987,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
 
         $def = '';
         $def .= strtoupper($sqlType['name']);
-        if ($column->getPrecision() && $column->getScale()) {
+        if (is_int($column->getPrecision()) && is_int($column->getScale())) {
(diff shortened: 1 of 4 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -987,7 +987,7 @@ protected function getColumnSqlDefinition(Column $column)
 
         $def = '';
         $def .= strtoupper($sqlType['name']);
-        if ($column->getPrecision() && $column->getScale()) {
+        if (is_int($column->getPrecision()) && is_int($column->getScale())) {
```
chinpei215: `!== null`  may be better than `is_int()`.

chinpei215:
Thank you for your contribution. Could you provide some test cases?

martijngastkemper:
I've created a PR (#1466) to populate the precision and scale so this PR isn't necessary anymore.

PR #1260, opened 2017-12-09
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertInternalType` and `assertNotInternalType` instead of `is_*` functions;
- `assertContains` instead of `in_array` function;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertAttributeEmpty`;
- `assertArrayHasKey` instead of `isset` function;
- `assertStringEqualsFile` when comparing a `string` and a file.


chinpei215:
Looks good to me.

PR #1263, opened 2017-12-14
Store version number in a separate file (fixes #1183)

### LINKED ISSUE #1183: phinx 0.9.1 introduces as 0.8.1 in console
Hi,
you've probably forgot to change version number in PhinxApplication class as it introduces in console as "Phinx by CakePHP - 0.8.1" but actually version installed is 0.9.1

olybri:
The version number displayed in the console currently does not match the actual version number since the v0.9.0 release. To fix that, I simply stored the version number in a separate file (as briefly suggested in #1183) so that  it would be easier to update at each release.

### DIFF
--- a/src/Phinx/Console/PhinxApplication.php
+++ b/src/Phinx/Console/PhinxApplication.php
@@ -32,6 +32,7 @@ use Phinx\Console\Command;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Yaml\Yaml;
 
 /**
  * Phinx console application.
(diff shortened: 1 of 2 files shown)


dereuromark:
:+1:

dereuromark:
@lorenzo Makes me wonder: Do we even need to maintain a version number in composer lock file days?...

broberts-mrmc:
It seems directly reading the composer.lock file would remove the need to have a separate “version file” whether YAML/PHP/text or whatever, but would also tighten the coupling between Phinx and composer by adding the composer.lock file specification/format as a dependency of Phinx and this feature. I’m not sure if that’s a “net” improvement – trading one dependency for another.

Also, wouldn’t reading composer.lock require logic to detect a global versus project-level installation of Phinx so that the correct .lock file is read?

dereuromark:
I meant not displaying at all. It is a dev tool. devs can look that up in easier qaxs otherwise.

broberts-mrmc:
Phinx is self-described as “ridiculously easy...migrations for everyone.” I think providing the version from the CLI is more consistent with that description than not.

I agree most users of Phinx are technically-oriented. However, most won’t – and shouldn’t need – crack open the source. That’s probably true for most packages.  It’s common for commands to provide a mechanism for reporting their version to users through, at least, an option. Often when there’s an issue of any kind, the first step is to check the version of the command, and it’s very useful, if not considerate of users, if identifying the version is NOT an exercise source code review.

broberts-mrmc:
Maybe a quick/easy/adequate solution is to provide a Phinx command/option like this:

    phinx -i
    phinx --info

Or adjust the existing:

    phinx -V
    phinx --version

So these are just a wrapper for:

    composer show robmorgan/phinx

Or maybe even take it a step further and parse the version from this command’s output and display it with other minimal information.

ADmad:
Why not just use the `version` config in `composer.json`?  Then we can have a release script which updates `composer.json` when tagging a new release like we have for CakePHP (which updates VERSION.txt).

ADmad:
Closing in favor of #1275

PR #1270, opened 2017-12-18
Adds dry run support for PDO inserts and bulkinserts

### LINKED ISSUE #1267: Add dry run support for the PDO insert function
Currently PdoAdapter#insert is running a prepared statement directly.
For dry runs that means the inserts are executed instead printed only.

My feature request (or is it a bug?) is to change the PDO insert and bulkinsert to print the SQL inserts in dry run mode.

I'm open to provide a PR for this.
One question for me is: since the PDO implementation uses prepared statements what should the output look like? The statement with placeholders and a dump of the values array? Or a ready to execute INSERT SQL statement?

weeman1337:
The PDO adapter insert and bulkinsert function now dump the statements in dry run mode.

This also removes the specific SQLite insert implementation.
The SQLite tests are still passing. So I think the special implementation isn't necessary.

Closes #1267


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -173,33 +173,105 @@ public function fetchAll($sql)
...
+    private function dryRunInsert(Table $table, $row)
+    {
+        $sql = sprintf(
+            'INSERT INTO %s ',
+            $this->quoteTableName($table->getName())
+        );
+
+        $columns = array_keys($row);
+        $sql .= '(' . implode(', ', array_map([$this, 'quoteColumnName'], $columns)) . ')';
+        $sql .= ' VALUES (' . implode(', ', array_map([$this, 'quoteValue'], $row)) . ');';
```
chinpei215: Please do not copy&paste. You can add a new private method and use it in `runInsert()` and `dryRunInsert()`. But please wait for a while before doing the work. I think we should discuss how to implement dry-run feature first.

chinpei215: Sorry, I misread. It was not a full copy&pasted.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -209,7 +281,7 @@ public function bulkinsert(Table $table, $rows)
         }
 
         $count_keys = count($keys);
-        $query = "(" . implode(', ', array_fill(0, $count_keys, '?')) . ")";
+        $query = '(' . implode(', ', array_fill(0, $count_keys, '?')) . ')';
```
chinpei215: I think the same thing can be said here. I think we could do like the following:
```php
if ($this->isDryRunEnabled()) {
    $values = array_map(function ($row) {
        return '(' . implode(', ', array_map([$this, 'quoteValue'], $row)) . ')';
    }, $rows);
    
    ...
} else {
    $values = [];
    foreach ($rows as $row) {
        foreach ($row as $v) {
            $values[] = $v;
        }
    }

    ...
}
```

weeman1337:
@chinpei215 I merged the methods, so there is no duplicate code any more.

chinpei215:
@weeman1337 Thank you. Looks good to me.

lorenzo:
Thanks!

PR #1275, opened 2017-12-21
Add version to composer.json.

ADmad:
Alternative to #1263


saeideng:
nice 
we can add version into cakephp/app/composer.json too :)

olybri:
I noticed that running Phinx from a Phar archive now fails to print the version since *composer.json* is not part of the archive. Adding *composer.json* into *box.json* should fix that.

dereuromark:
Are you able to quickly fix up a PR here?

olybri:
#1277 fixes this.

PR #1277, opened 2017-12-23
Include composer.json in Phar archive

olybri:
The displayed version number is now being read from composer.json, so this file should be part of the Phar archive as well.

### DIFF
--- a/box.json
+++ b/box.json
@@ -8,7 +8,8 @@
     "directories": ["src","app"],
     "files": [
         "LICENSE",
-        "phinx.yml"
+        "phinx.yml",
+        "composer.json"
     ],
     "finder": [
         {


dereuromark:
Thank you!

glensc:
but having `version` in composer.json is not recommended. newest composer even warns about that.
having git tag is sufficient for composer to detect versions.

https://getcomposer.org/doc/04-schema.md#version

for phar, if using [box](https://github.com/box-project/box2) to build it supports replacing placeholders.

example:
- https://github.com/eventum/cli/blob/0.1.0/eventum.php#L19 - define placeholder
- https://github.com/eventum/cli/blob/0.1.0/box.json#L40 - declare placeholder replacement

PR #1280, opened 2018-01-02
Use phpstan v9 and fix reported errors.

ADmad:


antograssiot:
@ADmad you forgot to update or push Travis config I think

ADmad:
@antograssiot Yeah i did forget to update travis config :P

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -981,9 +981,6 @@ protected function getIndexSqlDefinition(Index $index, $tableName)
             $indexName = $index->getName();
         } else {
             $columnNames = $index->getColumns();
-            if (is_string($columnNames)) {
```
ADmad: Can someone confirm that removing this check is safe? 

Docblock of `\Phinx\Db\Table\Index::getColumns()` says it returns `array` and such a `is_string()` isn't present in `MysqlAdapter`, so I think removing this check here and in `SqlServerAdapter` shouldn't be a problem.

ADmad:
[This](https://travis-ci.org/cakephp/phinx/jobs/324472303#L591) remaining failure is to due incorrect docblock in a symfony component. I had submitted a fix which was merged and should be available in their next bugfix release.

lorenzo:
Do you mind fixing this @ADmad?

ADmad:
@lorenzo Done.

PR #1282, opened 2018-01-04
Move from PHP 5.4 to PHP 7.0 for Docker

MasterOdin:
PHP 5.4 hit EOL over two years ago so it's probably not worth building Docker against it anymore. Moving to PHP 7.0 also has the added advantage of using PHPUnit 6.x development work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:5.4
+FROM php:7.0
 
 # system dependecies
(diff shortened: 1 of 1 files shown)


MasterOdin:
Not sure why Travis failed for PHP 5.5 and 5.6, though it's not like this affects Travis or even the codebase so it probably doesn't matter if they do pass (and would probably pass if just manually rerun).

PR #1284, opened 2018-01-12
Ensure SQLite alterations retain existing table triggers

nebez:
Modifying a SQLite table in any way that requires the use of a temporary table will result in the loss of all triggers that existed beforehand.

I realize that Phinx does not provide an interface for creating triggers. However, when using Phinx on a database that already had triggers, things end up getting destroyed in the process.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -358,7 +358,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
         );
 
         $this->execute($sql);
-
+        $this->migrateTriggers($tmpTableName, $tableName);
         $this->execute(sprintf('DROP TABLE %s', $this->quoteTableName($tmpTableName)));
     }
 
@@ -415,6 +415,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
         );
 
         $this->execute($sql);
+        $this->migrateTriggers($tmpTableName, $tableName);
         $this->execute(sprintf('DROP TABLE %s', $this->quoteTableName($tmpTableName)));
     }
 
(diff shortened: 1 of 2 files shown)


chinpei215:
Looks like Phinx does dangerous operations at the moment. Would the operations keep indexes and foreign keys correctly?

nebez:
Yes, it's both dangerous and it does keep indices/foreign keys properly.
Due to some limitations of SQLite, you need to move tables and data around (using temporary tables) to perform certain table/column operations.

See here for a temporary table creation to modify a column:
https://github.com/lxrco/phinx/blob/375e3d01e22dc3ebd189debb554ecb13aae18f7b/src/Phinx/Db/Adapter/SQLiteAdapter.php#L368-L405

nebez:
See here also for a StackExchange answer on the topic: https://stackoverflow.com/questions/805363/how-do-i-rename-a-column-in-a-sqlite-database-table#805508

The limitations of SQLite are widely known and the code is currently reflecting the appropriate process.
Let me know if there's any chance of getting this merged.

chinpei215:
Sorry for the late reply. I will confirm your changes on my local machine.

chinpei215:
I confirmed your changes on my local computer. Triggers was re-created as expected. But, as I thought, all indexes seem to be removed unexpectedly if some column is renamed. Would it keep indexes on your environment really?

Personally, I prefer a way shown in the second answer in the StackOverflow page you linked. It also seems dangerous though.

nebez:
@chinpei215 - Thanks for checking.
I think I'm misunderstanding. The link to the StackOverflow post is what happens in the existing code that I linked here: https://github.com/lxrco/phinx/blob/375e3d01e22dc3ebd189debb554ecb13aae18f7b/src/Phinx/Db/Adapter/SQLiteAdapter.php#L343

chinpei215:
I mean I prefer [this answer](https://stackoverflow.com/questions/805363/how-do-i-rename-a-column-in-a-sqlite-database-table#6684034) personally if it works. It can keep indexes, foreign keys, and triggers without re-inserting all records from old table to new table.

nebez:
Ah, I understand now. That is a pretty huge change to the SQLite adapter as a whole. I'm not comfortable making that change with the current level of test coverage in the SQLite implementation.

I'll go ahead and close up the PR.

PR #1289, opened 2018-01-18
allow timestamp field of length '0'. fixes #1287

### LINKED ISSUE #1287: Timestamp of length '0' broken for Postgres
When specifying a 'limit' (or 'length') for a column of type `timestamp` in Postgres, it typically works as intended and creates a timestamp with that many decimals of sub-second precision.

However, this is broken for a limit of 0. We would expect to create a column of type `TIMESTAMP(0)`, but it actually generates a field of type `TIMESTAMP` which by default has 6 decimals of precision.

It looks like an oversight in PostgresAdapter.php on line 926.

```
if ($column->getLimit() || isset($sqlType['limit'])) {
```

This condition probably intends to check if the result of `$column->getLimit()` is not null, but it misses the case for a 0 limit.

I will try to create a pull request for this issue tonight.

shamblesides:
Sorry for the lack of test cases; I couldn't figure out how to get them to work. If you want I can keep trying sometime soon.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -923,8 +923,8 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
                     $sqlType['srid']
                 );
             } elseif (!in_array($sqlType['name'], ['integer', 'smallint', 'bigint'])) {
-                if ($column->getLimit() || isset($sqlType['limit'])) {
-                    $buffer[] = sprintf('(%s)', $column->getLimit() ?: $sqlType['limit']);
+                if ($column->getLimit() !== null || isset($sqlType['limit'])) {
+                    $buffer[] = sprintf('(%s)', ($column->getLimit() !== null) ? $column->getLimit() : $sqlType['limit']);
                 }
             }
 


chinpei215:
@shamblesides You might want to add testTimestampWithPrecision() to PostgresAdapterTest. I think you can use testAddDecimalWithPrecisionAndScale() as a reference.
https://github.com/cakephp/phinx/blob/7072a54ef8ec4b7b9819e73d277a99d34f92e8d5/tests/Phinx/Db/Adapter/PostgresAdapterTest.php#L331-L351

shamblesides:
I tried to create a test case, but it's not behaving how I would expect. Here's the test I wrote:

```
    public function testAddTimestampWithPrecision()
    {
        $table = new \Phinx\Db\Table('table1', [], $this->adapter);
        $table->save();
        $table
            ->addColumn('ts1', 'timestamp', ['limit' => 1])
            ->save();
        $columns = $this->adapter->getColumns('table1');
        foreach ($columns as $column) {
            if ($column->getName() == 'ts1') {
                $this->assertEquals("1", $column->getLimit());
            }
        }
    }
```

However, the error:

```
1) Test\Phinx\Db\Adapter\PostgresAdapterTest::testAddTimestampWithPrecision
Failed asserting that null matches expected '1'.
```

Why is `getLimit()` returning `null` here?

chinpei215:
Sorry if I am wrong, but probably you need to change the following SQL to fetch the `datetime_precision` column:
https://github.com/cakephp/phinx/blob/7072a54ef8ec4b7b9819e73d277a99d34f92e8d5/src/Phinx/Db/Adapter/PostgresAdapter.php#L303-L309

And, probably you need to call either `setPrecision()` or `setLimit()` to set fetched results in `PostgresAdapter::getColumns()`. I don't know why Phinx has two similar methods. Would `getLimit()` be better?

shamblesides:
"Precision" probably makes more sense. What we're really trying to do is set the number of decimal places that the timestamp has. It should probably work just like the `decimal` type. I'm going to revise my contribution.

shamblesides:
Ultimately I think I'd have to go in and edit other database adapters and other classes to make this timestamp precision universally working and available and tested. Honestly I'm not sure if I'm up to that task right now. Maybe another time.

shamblesides:
Notably, SQLite doesn't seem to have a date type and I don't get how that's handled in Phinx.

chinpei215:
Probably we need to use regular expressions to fetch precision from the result of `PRAGMA table_info()` statement.

dereuromark:
Any update here? Only a test is missing.

shamblesides:
I think I got stuck on something related to testing. It's been a while so I'm afraid I don't remember what it was anymore. I also only have this working on Postgres, not on other databases, which I thought might be a barrier for merging this in.

dereuromark:
That is true, but even so better to have it working for one than for none probably.
Maybe someone can make a MySQL patch here on top.

dereuromark:
Also, any chance we can have a small test case?

dereuromark:
ping @shamblesides

smmd:
This change is not needed it anymore after this changes were merged https://github.com/cakephp/phinx/pull/1466 

Inspired on this I created a new PR that might be useful https://github.com/cakephp/phinx/pull/1631 I will be waiting for your opinions

cc @dereuromark

PR #1291, opened 2018-01-23
Bump year

cleptric:
Should we also add `Copyright (c) 2005-2018, Cake Software Foundation, Inc. (https://cakefoundation.org)` underneath?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 (The MIT license)
 
-Copyright (c) 2012-2017 Rob Morgan
+Copyright (c) 2012-2018 Rob Morgan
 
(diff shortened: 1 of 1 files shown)


chinpei215:
I think it should be:
```
Copyright (c) 2012-2017 Rob Morgan
Copyright (c) 2017-2018, Cake Software Foundation, Inc. 
```

@phpnut What should we do?

dereuromark:
2018 should be replace with "present" :) most importantly.

ravage84:
Agree with @chinpei215, though changing or showing the actual year is not that important as I showed some other place.

That would be OK with me, @dereuromark

dereuromark:
Let's please do that for the next one of these changes then!

PR #1193, opened 2017-09-22
Added full support of different schemas for Postgres.

corpsee:
Added full support of different schemas for Postgres (#828). Use-case:
```php
$this->getAdapter()->createSchema('schema1');

if (!$this->hasTable('schema1.example1')) {
    $table = $this->table('schema1.example1');
    $table
        ->addColumn('column1', 'integer')
        ->save();
}

$this->getAdapter()->createSchema('schema2');

if (!$this->hasTable('schema2.example2')) {
    $table = $this->table('schema2.example2');
    $table
        ->addColumn('column2', 'integer')
        ->save();

    $table
        ->addForeignKey('column2', 'schema1.example1', 'id')
        ->save();
}
```

Instead of now:
```php
$this->getAdapter()->createSchema('schema1');
$this->getAdapter()->setOptions(array_replace($this->getAdapter()->getOptions(), ['schema' => 'schema1']));

if (!$this->hasTable('example1')) {
    $table = $this->table('example1');
    $table
        ->addColumn('column1', 'integer')
        ->save();
}

$this->getAdapter()->createSchema('schema2');
$this->getAdapter()->setOptions(array_replace($this->getAdapter()->getOptions(), ['schema' => 'schema2']));

if (!$this->hasTable('example2')) {
    $table = $this->table('example2');
    $table
        ->addColumn('column2', 'integer')
        ->save();

    $this->execute('ALTER TABLE schema2.example2 ADD CONSTRAINT example2_column2_fkey FOREIGN KEY (column2) REFERENCES schema1.example1(id)');
}

$this->getAdapter()->setOptions(array_replace($this->getAdapter()->getOptions(), ['schema' => 'public']));
```

PR #1292, opened 2018-01-29
Make hasTable understand 'schema.tableName' notation.

winkbrace:
In a setup with one application using multiple mysql databases, I don't want to have to setup phinx for each database. Instead, all we have to do is make `hasTable` understand the "schema.table" notation and we're golden.


dereuromark:
Can this introduce a regression for tables that might contain a dot already in their table name?
Or is this not possible so far?

winkbrace:
My god, a dot in the table name allowed?!

So I guess that makes this change breaking for those users. Though honestly I don't expect there to be more than a handful.

The only way to be sure is to check the database meta tables then, but that can become expensive... hmm..

winkbrace:
Since we are already checking the information schema, adding support for tables with a dot in the name was not difficult or very expensive.

When hasTable finds a dot in the table name, it will first assume "schema.table" notation and if table is not found, fall through to looking up the complete table name in the main schema.

lorenzo:
Does this work in postgres as well? I'm not comfortable only adding support for mysql

corpsee:
@lorenzo [There is](https://github.com/cakephp/phinx/pull/1193) support schemas for PostgreSQL (Not only for `hasTable` but for all operations for tables with schema).

winkbrace:
Since we require this fork to use phinx, I had to rebase it on the latest stable release. So we have conflicts with master which I will resolve when you are ready to merge.

I will also look into the postgres changes and see what I have to add for mysql to make it schema aware everywhere.

dereuromark:
Are you still interested in fixing this up?

winkbrace:
I am happy to resolve the conflicts, but since I've moved to another project and am not using phinx anymore I suggest we merge the changes that I've made and that are enough for most use cases where multiple mysql databases are used.

At work we happily used my fork for over a year without any issues.

lorenzo:
@winkbrace if you can resolve the conflicts, that would be great

dereuromark:
New PR: https://github.com/cakephp/phinx/pull/1624

winkbrace:
Thanks Arjen <3

PR #1293, opened 2018-01-30
Make unit tests pass on php 7.2

winkbrace:
Make unit tests pass on php 7.2 by requiring minimum version of sebastian/comparator.


composer.json
```diff
@@ -32,6 +32,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.35|^5.7|^6.5",
+        "sebastian/comparator": ">=1.2.3",
```
ADmad: Just to confirm this package should be under `require-dev` and not `require`?

winkbrace: Yes, it's a dependency of phpunit (under sebastian's private github account)

ADmad: Okay, thanks.

PR #1296, opened 2018-02-05
Fix SQLite identity nullability

51systems:
Ensure primary key auto-incrementing columns are not null


dereuromark:
Looks good to me.

lorenzo:
Thanks!

PR #1298, opened 2018-02-06
Fix SQLite & MySQL identity definition incompatibility

### LINKED ISSUE #1297: SQLite & MySQL identity definition incompatible
If you are using the `identity` key to define an auto-increment column the syntax is incompatible between SQLite and MySQL.

In MySQL you _must_ define the `primary_key` in the table definition in addition to setting the `identity` key to true. For example:
```php
$table = $this->table('test', [
    'id' => false,
    'primary_key' => ['custom_id']
]);

$table
    ->addColumn('custom_id',
        'integer', [
            'signed' => false,
            'identity' => true
        ])
```

If you don't do this, the `PRIMARY KEY` is not set for the table.

However, in the case of the SQLite adapter, when you set `identity` to `true`, this defines the column as `PRIMARY KEY AUTOINCREMENT`, so if the column appears in the `primary_key` array, it tries to add the primary key multiple times, resulting in an exception (eg: `SQLSTATE[HY000]: General error: 1 table "test" has more than one primary key`)

51systems:
Modify createTable to remove any identity columns from the primary_key
list to ensure that the SQLite adapter doesn't attempt to define the
primary key multiple times.

Fixes #1297

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -187,6 +187,20 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
         $sql .= $this->quoteTableName($table->getName()) . ' (';
         foreach ($columns as $column) {
             $sql .= $this->quoteColumnName($column->getName()) . ' ' . $this->getColumnSqlDefinition($column) . ', ';
+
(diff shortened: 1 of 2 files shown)


dereuromark:
Looks good to me!

lorenzo:
Thanks!

PR #1305, opened 2018-02-21
allow custom id field to be unsigned

rommon:
currently a specified id field is not able to be unsigned.


dereuromark:
Looks good to me.

lorenzo:
Thanks!

Issue #1137, opened 2017-07-21
Get -e <environment> from phinx.php config


l0gicgate:
I've been trying to find a way to get the environment option from my phinx.php file and it seems rather complicated.

The only way right now is to export PHINX_ENVIRONMENT prior to running the phinx console and I don't want to have to do that.

Is there a simple way to do this?

lorenzo:
What's the use case for getting back the environment name?

l0gicgate:
So I can load the proper .env file which gives my dependencies a PDO to the right DB. 

During my migrations I have to stage fixtures in the database, my repositories require a connection to the right database and there's no way for me to know which environment was detected / which one it defaulted to. 

I can put a code sample together if you really need to visualize what I'm trying to do. 

Thanks

broberts-mrmc:
Pierre,

I've yet to contribute to the Phinx project (sad face) though I watch and utilize the project, so take this with a grain of salt. My point is, it's NOT an opinion or recommendation from a Phinx project lead or even contributor.

That said, I believe we've done something like this by leveraging our project/application's configuration loader in the Phinx source with one line of code. What we've done is include our application's configuration loader in Phinx's PHP config file, and set the various Phinx "default" database configuration options to items from our configuration loader. That is, our configuration loader, looks at information about its runtime environment (e.g. host/server, path of executing script, etc) to load appropriate DB settings among other things, which are assigned to the appropriate options in the Phinx config. I suppose you could actually instantiate a PDO object too.

Unless I'm misunderstanding, it seems a similar solution would also work in the situation you're describing. Does your app/code base have some kind of configuration loader?

This approach mostly eliminates the need for the Phinx command "-e" option, though we're looking at possibly leveraging it for a different purpose – to support multiple databases, not different instances or environments for/of the same database. However, we're still working out the details on this piece as there would really need to be a district set of migrations for each database, though, in our case, there's only one project to contain them.

Regards,

Ben Roberts
903-520-4774

On Jul 22, 2017, at 11:31 AM, Pierre Bérubé <notifications@github.com<mailto:notifications@github.com>> wrote:


So I can load the proper .env file which gives my dependencies a PDO to the right DB.

During my migrations I have to stage fixtures in the database, my repositories require a connection to the right database and there's no way for me to know which environment was detected / which one it defaulted to.

I can put a code sample together if you really need to visualize what I'm trying to do.

Thanks

l0gicgate:
[...]

PR #1308, opened 2018-02-24
Pass Environment Variable Down To Migrations

l0gicgate:
This PR adds functionality requested in #1137.

**Issue**
Sometimes within a migration you may want to have different behavior based on the environment you are in. For example, in my development environment I also stage fixtures in my database i.e add a default user to the development database but not in the testing database.

**Documentation**
- `AbstractMigration` and `MigrationInterface` have been modified with one new methods `::getEnvironment()`.
- The constructor signature of `AbstractMigration` has been modified so the migration gets instantiated with the `$environment` parameter. I don't think this will cause any issues to the end user as I assume users do not override the constructor, they simply extend the migration when creating their own. 

**Example Usage**
```php
use DI\Container;
use Dotenv\Dotenv;

abstract class SQLMigration extends AbstractMigration
{
    /**
     * @return bool
     */
    protected function shouldStage()
    {
        return $this->getEnvironment() !== 'testing' && $this->isMigratingUp();
    }
}
```

**Status**
- [X] Implement functionality
- [X] Refactor existing tests and add new coverage for additional feature
- [ ] Discussion and Code Review

### DIFF
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -196,16 +196,6 @@ abstract class AbstractMigration implements MigrationInterface
     /**
      * {@inheritdoc}
      */
-    public function setEnvironment($environment)
-    {
-        $this->environment = $environment;
-
-        return $this;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function setVersion($version)
     {
         $this->version = $version;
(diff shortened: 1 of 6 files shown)


src/Phinx/Migration/AbstractMigration.php
```diff
@@ -177,6 +185,24 @@ public function getName()
...
+     */
+    public function getEnvironment()
+    {
+        return $this->environment;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setEnvironment($environment)
```
lorenzo: Can you remove this? I don't think changing the environment at runtime is a good idea

l0gicgate: Done!

lorenzo:
Thanks!

Issue #1276, opened 2017-12-22
[MysqlAdapter] CREATE TABLE and Generated Columns


demyan112rv:
The method that creates the expression for the column does not support the generated columns. 
For instance: 

```
CREATE TABLE `my_table` (
    ...
    `generated_column` int(11) [GENERATED ALWAYS] AS (expression)
    [VIRTUAL | STORED] [NOT NULL | NULL]
    [UNIQUE [KEY]] [[PRIMARY] KEY],
    ...
) ENGINE=InnoDB
```

Can you add support for these columns?

[\Phinx\Db\Adapter\MysqlAdapter::getColumnSqlDefinition()](https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/MysqlAdapter.php#L989)
[Mysql doc](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html)

chinpei215:
I think #1175 will allow you to add those columns if it gets merged.

PR #1311, opened 2018-02-27
Add MySQL generated columns

hdogan:
Quick & dirty solution for #1276

This solution may be not necessary anymore if #1175 is accepted.

### DIFF
--- a/src/Phinx/Db/Table/Column.php
+++ b/src/Phinx/Db/Table/Column.php
@@ -122,6 +122,21 @@ class Column
     protected $values;
 
     /**
+     * @var boolean
+     */
+    protected $virtual = false;
+
+    /**
+     * @var boolean
+     */
+    protected $stored = false;
+
+    /**
+     * @var string
+     */
+    protected $expression;
+
+    /**
      * Sets the column name.
      *
      * @param string $name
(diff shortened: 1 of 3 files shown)


src/Phinx/Db/Table/Column.php
```diff
@@ -121,6 +121,21 @@ class Column
      */
     protected $values;
 
+    /**
+     * @var boolean
```
dereuromark: bool

dereuromark:
Looks OK to be merged after CS fixes.

dereuromark:
@lorenzo Do you prefer https://github.com/cakephp/phinx/pull/1175 ? Or shall we continue with this?

hdogan:
SQL Server has a similar featured called "computed column". See issue #1317 
I'm not exactly sure if #1175 solves those kind of problems. It would be nice to have some more examples of #1175 PR with generated/computed column scenario.

Or adapter class (or user) may inject custom/relaxed properties into column class then use it in own way. Like:

```php
// For MySQL generated columns
$table->addColumn('full_name', 'string', [
    'limit' => 100,
    'custom' => [
        'generated' => true,
        'stored' => true, // virtual if false or not defined
        'expression' => 'CONCAT(first_name, \' \', last_name)'
    ]
]);

// For SQL Server computed columns
$table->addColumn('full_name', 'string', [
    'limit' => 100,
    'custom' => [
        'computed' => true,
        'persisted' => true, // not persisted if false or not defined
        'expression' => 'CONCAT(first_name, \' \', last_name)'
    ]
]);
```

dereuromark:
The other PR was merged. I guess we can close?

designermonkey:
@dereuromark you've stated that 'the other pr was merged' but I'm not sure what you're referencing?

I have come across this looking for support for generated columns and I can't find anything about it in the codebase. Is there support added or not? If so how do I use it?

/cc @hdogan

dereuromark:
Wasnt it https://github.com/cakephp/phinx/pull/1175 ?

designermonkey:
How would that be used to create generated columns? Is the literal value just used as is? Some examples other than simple types might be useful in the docs, as I completely missed that this could be used for generated columns :/

dereuromark:
All right, maybe you can reopen a PR with the suggested changes and we can try to fit it in asap?

designermonkey:
I wouldn't know how to document this feature, that's why I'm looking for it and failing to spot how to do it.

If I could, I would.

PR #1313, opened 2018-02-27
Fix SQLite transaction handling

### LINKED ISSUE #1312: Sqlite PDO->inTransaction() returns wrong value
Due to the way that the SQLiteAdapter manages transactions (via execute rather than the provided PDO transaction methods), `$adapter->getConnection()->inTransaction()` will always return `false`.

51systems:
Modify SQLiteAdapter to use the transaction methods provided by PDO
rather than directly executing transaction commands against the connection.

This allows getConnection()->inTransaction() to return the proper value.

Fixes #1312


dereuromark:
Looks good to me.

lorenzo:
Thanks!

Issue #753, opened 2016-01-17
Feature Request: custom data types


bradynpoulsen:
I think it would be really useful to have a way to add a custom column name data type when using `addColumn()`. This would then make it so it could be aliased to an actual data type. 

For example with MySQL,

*without custom type*
```php
$table->addColumn('id', 'binary', ['limit' => 16]); // uuid v4
```

*with custom type*
```php
$table->addColumn('id', 'uuid');
```

bradynpoulsen:
Here is a possible flow:
- [ ] 1. Register `'uuid'` to my `UuidCustomType` class that implements `CustomTypeInterface` via
    - [ ] a. YAML entry
    - [ ] b. Inside a custom migration base class on construct
- [ ] 2. Implement `UuidCustomType` that aliases to the real `'binary'` type with `['limit' => 16]`

twoixter:
@bradynpoulsen This is actually something I was thinking about implementing. This semantic is very much like a User Defined Type or a Data Domain. I was skeptical at first because each database has a concept of UDT or Domain (E.G: [Check constraints](https://en.wikipedia.org/wiki/Check_constraint) and UDT in SQL Server) and AFAIK, don't supported in migrations. My concern was that if implemented, it would give the impression to extend the user defined type to the DB Adapter.

This is my suggestion for the flow. If interested, I might create a fork and start developing:

** Define a new "domain" for our data types in the phinx.yml configuration**

```yaml
# Domain defined in the config phinx.ml
# This would start a "domain" for our data types
domain:
    phone:    # New user defined data type name
        type: string   # Based on this base phinx data type
        options:       # With these options.
            length: 9
            null: true
    short_status:
        type: enum
        values: pending, started, finished
```

The intention is to use those as new data types, and they'd just use the values for `type` and `options`.

```php
    // Use our user defined types from config
    $table->addColumn("phone_number", "phone");
    $table->addColumn("task_status", "short_status");
```

Since those are really just aliases for the type and options, we could just add new options to override:

```php
    // Use our user defined types from config
    $table->addColumn("mobile_number", "phone", array("length" => 12, "null" => false));
```

Let me know what do you think.

bradynpoulsen:
That would certainly acheive the goals of my example in a much more isolated location instead of having to include a bunch of driver classes

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

twoixter:
[...]

PR #1320, opened 2018-03-06
Adds Custom Data Types (Data Domain)

twoixter:
As commented in #753 this PR adds **Custom Data Types** to phinx.

This feature works by adding a root config option to `phinx.yml` creating a data domain:

```yaml
# Add a root config option to "phinx.yml" to start a Data Domain
data_domain:
    phone_number:           # You can add different new data types
        type: string        # Use a base type as reference
        length: 20          # Using default column options
    address_line:
        type: string
        length: 150
    file:
        type: blob          # You can even extend special types
        limit: BLOB_LONG    # For MySQL DB. Uses MysqlAdapter::BLOB_LONG
    boolean:
        type: boolean       # Or customize native phinx types
        signed: false       # Customization of the boolean to be unsigned
    image_type:
        type: enum          # Enums can use YAML lists or a comma separated string
        values:
            - gif
            - jpg
            - png
```

You can then use the new data types in migrations:

```php
// Use our user defined types from config
$table->addColumn("customer_phone", "phone_number");
$table->addColumn("file_upload", "file", ["null" => true]);
```


src/Phinx/Config/Config.php
```diff
@@ -313,6 +313,18 @@ public function getTemplateClass()
         return $this->values['templates']['class'];
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getDataDomain()
+    {
+        if (!isset($this->values['data_domain'])) {
+            return false;
```
dereuromark: if this method returns always array, why not returning `[]` here?
Cleaner return codes.

src/Phinx/Config/ConfigInterface.php
```diff
@@ -116,6 +116,13 @@ public function getTemplateFile();
      */
     public function getTemplateClass();
 
+    /**
+     * Get the data domain array.
+     *
+     * @return array[]|false
```
dereuromark: This can then be just `array` again.

tests/Phinx/Config/ConfigTest.php
```diff
@@ -96,6 +96,24 @@ public function testHasEnvironmentMethod()
...
+        $this->assertInternalType('array', $config->getDataDomain());
+    }
+
+    /**
+     * @covers \Phinx\Config\Config::getDataDomain
+     */
+    public function testDoesNotThrowsWithEmptyDataDomain()
+    {
+        $config = new Config([]);
+        $this->assertFalse($config->getDataDomain());
```
dereuromark: Either empty array or null IMO. Never false as alternative (anti-pattern).

src/Phinx/Config/Config.php
twoixter: I was following the pattern from the methods above (`getTemplateFile`, `getTemplateClass`), but thinking again you're right. I'll make the return value always an array. Thank you!

src/Phinx/Db/Adapter/AbstractAdapter.php
```diff
@@ -195,6 +204,102 @@ public function setSchemaTableName($schemaTableName)
...
+        return $this->dataDomain;
+    }
+
+    /**
+     * Sets the data domain.
+     *
+     * @param array $dataDomain Array for the data domain
+     * @return $this
+     */
+    public function setDataDomain($dataDomain)
```
dereuromark: Should we typehint array here?

src/Phinx/Db/Adapter/AbstractAdapter.php
```diff
@@ -195,6 +204,102 @@ public function setSchemaTableName($schemaTableName)
...
+
+            // Do a simple replacement for the 'length' / 'limit' option and
+            // detect hinting values for 'limit'.
+            if (isset($options['length'])) {
+                $options['limit'] = $options['length'];
+                unset($options['length']);
+            }
+
+            if (isset($options['limit']) && !is_numeric($options['limit'])) {
+                if (defined('static::' . $options['limit'])) {
```
dereuromark: if (!defined) exception saves you a else here.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -156,6 +156,16 @@ public function setOutput(OutputInterface $output);
...
 
+    /**
+     * Returns a new Phinx\Db\Table\Column using the existent data domain.
+     *
+     * @param string $columnName The desired column name
+     * @param string $type The type for the column. Can be a data domain type.
+     * @param array $options Options array
+     * @return \Phinx\Db\Table\Column
+     */
+    public function getColumnForType($columnName, $type, $options);
```
dereuromark: array typehint on options?

tests/Phinx/Db/Adapter/DataDomainTest.php
```diff
@@ -0,0 +1,190 @@
...
+
+        $mysql_adapter = new MysqlAdapter(['data_domain' => $data_domain]);
+        $dd = $mysql_adapter->getDataDomain();
+
+        $this->assertInternalType('array', $dd['phone_number']['options']);
+        $this->assertEquals(19, $dd['phone_number']['options']['limit']);
+    }
+
+    /**
+     *
```
dereuromark: `@return void` missing

src/Phinx/Db/Adapter/AdapterInterface.php
twoixter: I think this is no bc break because it wasn't implemented before and the code is basically the same as when it was part of the original method if-else block.

However I can see it can be `protected` method anyway, since it is not intended to be called from public user side (only from concrete adapter classes).

tests/Phinx/Db/Adapter/DataDomainTest.php
twoixter: This is a test case. I've seen no test cases with type hinting or doc comments. (apart from phpunit `@expectedException`, `@covers`, etc). But I can add a `@return void` if you like. ;-)

dereuromark:
Whats the status on this?
Please close all conversations that are resolved.

dereuromark:
Any chance we can revive and merge this?

twoixter:
Yes @dereuromark , I'll try to resolve all merge conflicts and resolve any conflicts left asap.

twoixter:
@dereuromark I think I've resolved all issues! Please, let me know if there is something still unresolved.

dereuromark:
So based on your changes we can also ship the cake core type `binaryuuid` now?
See https://github.com/cakephp/phinx/issues/753#issuecomment-602057688

dereuromark:
We are releasing a new major (0.12) soon.
Based on your changes to interfaces ( src/Phinx/Config/ConfigInterface.php ), should we maybe merge this into 0.next branch?

twoixter:
> So based on your changes we can also ship the cake core type `binaryuuid` now?
> See [#753 (comment)](https://github.com/cakephp/phinx/issues/753#issuecomment-602057688)

I'm not sure if this is related. I mean, my changes are related to Phinx migrations on standard Phinx adapters @ `src/Phinx/Db/Adapter/*`. So I think you'll get the same error as in the #753 comment since it does not automatically integrates custom Cake types.

To add a new `binaryuuid` for Cake, I think you'll need to create a new data domain in the config file. For example:

```yaml
# Add a root config option to "phinx.yml" to start a Data Domain
data_domain:
    binaryuuid:       # Create the "binaryuuid" data type
        type: blob    # Use a base type as reference (Is Blob good for binaryuuid?)
        default: null
        null: true
```

Adding this new "data domain" into `phinx.yml` allows you use it as a new data type:

```php
$table->addColumn('uuid', 'binaryuuid');
```

I don't know if it's ok to automatically inject Cake types by default for folks (like me) that uses Phinx outside CakePHP projects.

twoixter:
> We are releasing a new major (0.12) soon.
> Based on your changes to interfaces ( src/Phinx/Config/ConfigInterface.php ), should we maybe merge this into 0.next branch?

Absolutely, please merge into the most relevant branch. Thanks!

dereuromark:
Can you switch the target branch to 0.next?

twoixter:
@dereuromark I just switched branches to `0.next` and then the Travis CI shows many errors from the tests. Did I just done it wrong or those tests are just "work in progress" from the `0.next` branch? ;-)

dereuromark:
It is possible
If those arent easily fixable, you can also make a changeset with only yours.
And we merge/fix the rest separately.

twoixter:
Hi @dereuromark !! I've reviewed the new errors and I think they're not related to changes from my branch.

They're easy to fix I think, but I find it weird to fix tests on this branch for other not related features.

So, what are the options? I don't know how to make a "changeset" :-D Bear with me, hehehe.

dereuromark:
Either fixing the other changes in another PR and we merge that first, then this one will be auto green
Or all in this one.
The result would be the same.

dereuromark:
Technically, those are not different features, but as you can see from
https://github.com/cakephp/phinx/branches
You now rebased in a way that it includes a forward port of master into 0.next.
This has to be done anyway, as we need to be up to date with bugfixes before we do
more features and releases on 0.next.

twoixter:
Thank you @dereuromark !! That might be it. Does it makes sense if I try to help merging `master` into `0.next' then?

dereuromark:
master is now merged into it. 
Are you able to re-apply your changes?

twoixter:
@dereuromark Done!

PR #1322, opened 2018-03-12
Improve error reporting when loading migration file

jbq:
I was writing a new migration, and just forgot a semicolon at the end of a statement.  It took half an hour for me to figure it out, as phinx does not report errors when loading a migration file, it just exits silently:

```
$ ./vendor/bin/phinx status -e current -vv
Phinx by CakePHP - https://phinx.org. 0.9.2

using config file ./phinx.php
using config parser php
using migration paths 
 - /var/www/html/db/migration
ordering by creation time
```

I added error reporting, and verbose logging:

```
$ ./vendor/bin/phinx status -e current -vv
Phinx by CakePHP - https://phinx.org. 0.9.2

using config file ./phinx.php
using config parser php
using migration paths 
 - /var/www/html/db/migration
ordering by creation time
Migration file db/migration/20180312095838_predef_questions.php
Valid migration file db/migration/20180312095838_predef_questions.php
Loading class PredefQuestions from db/migration/20180312095838_predef_questions.php

Parse error: syntax error, unexpected '$this' (T_VARIABLE) in /var/www/html/db/migration/20180312095838_predef_questions.php on line 50
```

This is a much better output to understand what's wrong.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -668,6 +668,10 @@ class Manager
         if ($this->migrations === null) {
             $phpFiles = $this->getMigrationFiles();
 
+            if (OutputInterface::VERBOSITY_VERBOSE <= $this->getOutput()->getVerbosity()) {
+                $this->getOutput()->writeln("Migration file " . implode(", ", $phpFiles));
+            }
+
             // filter the files to only get the ones that match our naming scheme
             $fileNames = [];
             /** @var \Phinx\Migration\AbstractMigration[] $versions */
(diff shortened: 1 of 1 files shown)


src/Phinx/Migration/Manager.php
```diff
@@ -720,6 +739,10 @@ public function getMigrations($environment)
                     }
 
                     $versions[$version] = $migration;
+                } else {
+                    if (OutputInterface::VERBOSITY_VERBOSE <= $this->getOutput()->getVerbosity()) {
+                        $this->getOutput()->writeln("Invalid migration file " . implode(", ", $phpFiles));
```
dereuromark: Should we use `<error>` for red and colorful response? Or keep plain one?

jbq: No this is just a logging statement, let's not use `<error>` here

lorenzo:
Thanks!

Issue #1301, opened 2018-02-09
MySQL Timestamp Precision


johnrobertporter:
The `timestamp `type in MySQL allows for precision values like `timestamp(4)`. Phinx seems unable to do this at present.

dereuromark:
Duplicate of https://github.com/cakephp/phinx/issues/1301 ?

dereuromark:
Are you able to provide a pull request here?

johnrobertporter:
Sadly not, I am under a lot of pressure to get too much work done. I would if I could.

Plus I don't know the internals of this code base and it looks very complicated.

PR #1323, opened 2018-03-15
Add ability to specify limit for timestamp/date/time columns for mysql

MasterOdin:
Implements #1003 and #1301, allowing someone to specify a limit for time, timestamp, and datetime types. Trying to use this on older versions of mysql (< 5.6.3) will throw a PDO exception. 

For local testing, I bumped the docker-compose to use mysql 5.6 instead of 5.5.

I've added a check on mysql version for the related tests so as to skip them on those MySQL versions (and added the necessary function to get database version for the PDOAdapter using http://php.net/manual/en/pdo.getattribute.php).


lorenzo:
Thanks!

PR #1327, opened 2018-03-23
Handle `schema` option defined in the Configuration

lukasz-kozik:
Currently I'm not able to write migration for PostgreSQL for specific schema without specifying it in the migration file. The problem is with migration status if I operate on multiple schemas. Seting up `search_path` right after connection was set solves the problem.


JayPHP:
Can we get some feedback on this please?

PR #1331, opened 2018-03-30
Fix CS sniffer sync.

dereuromark:
Refs https://github.com/cakephp/phinx/pull/1330 issues around CS

At least now we know why..
tons of "Missing parameter comment" still in there.
tbh I never really liked this strict one as it only lead to a lot of duplicating without much value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
 script:
     - if [[ $DEFAULT = 1 ]]; then vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover; fi
     - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.8 && vendor/bin/phpstan analyse -c phpstan.neon -l 4 src; fi
-    - if [[ $PHPCS = 1 ]]; then phpcs -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP --ignore=/_rootDirectories/ src/ tests/ app/; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -34,7 +34,7 @@ before_script:
 script:
     - if [[ $DEFAULT = 1 ]]; then vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover; fi
     - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.8 && vendor/bin/phpstan analyse -c phpstan.neon -l 4 src; fi
-    - if [[ $PHPCS = 1 ]]; then composer require --dev squizlabs/php_codesniffer:~3.1.1 && vendor/bin/phpcs -p -s src; fi
+    - if [[ $PHPCS = 1 ]]; then phpcs -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/ tests/ app/; fi
```
ceeram: cant do `then phpcs -p `
should be `then vendor/bin/phpcs -p `

ceeram: also probs will complain about cakephp standard not being available

dereuromark: Good point, I re-use the  composer command then.

JayPHP:
What's the status of this PR?

dereuromark:
Check the travis report. Too many issues regarding comments in doc blocks.
I suggest just silencing them as they also bear no real value here right now moving forward.

lorenzo:
Closing as the code has diverged significantly

PR #1332, opened 2018-04-02
allow override command name

FallDi:
Symfony command `Symfony\Component\Console\Command;` allows to override command name via constructor, but Phinx command do not allow it.

```
/**
 * @param string|null $name The name of the command; passing null means it must be set in configure()
 *
 * @throws LogicException When the command name is empty
 */
public function __construct(string $name = null)
{
    $this->definition = new InputDefinition();

    if (null !== $name || null !== $name = static::getDefaultName()) {
        $this->setName($name);
    }

    $this->configure();
}
```
Use-case:
Customer use Phinx commands only, in self big CLI application, and would like to add `migration:XXX` prefix to each command

### DIFF
--- a/src/Phinx/Console/Command/Breakpoint.php
+++ b/src/Phinx/Console/Command/Breakpoint.php
@@ -44,7 +44,7 @@ class Breakpoint extends AbstractCommand
 
         $this->addOption('--environment', '-e', InputOption::VALUE_REQUIRED, 'The target environment.');
 
-        $this->setName('breakpoint')
+        $this->setName($this->getName() ?: 'breakpoint')
             ->setDescription('Manage breakpoints')
             ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to set or clear a breakpoint against')
             ->addOption('--remove-all', '-r', InputOption::VALUE_NONE, 'Remove all breakpoints')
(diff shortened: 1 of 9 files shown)


rquadling:
My approach for this, when I need to include Phinx in my utils is ...

```
                    foreach ($this->phinxApplication->all() as $commandName => $command) {
                        // Ignore the phinx:help and phinx:list commands.
                        if (!in_array($commandName, ['help', 'list'])) {
                            $phinxCommands[] = $command->setName('phinx:'.str_replace(':', '-', $commandName));
                        }
                    }
```

`$this->phinxApplication` is an injected instance of `Phinx\Console\PhinxApplication`.

This gives me ...

```
 phinx
  phinx:breakpoint                            Manage breakpoints
  phinx:create                                Create a new migration
  phinx:init                                  Initialize the application for Phinx
  phinx:migrate                               Migrate the database
  phinx:rollback                              Rollback the last or to a specific migration
  phinx:seed-create                           Create a new database seeder
  phinx:seed-run                              Run database seeders
  phinx:status                                Show migration status
  phinx:test                                  Verify the configuration file
```

FallDi:
Yes you are right. It possible to inject name via `$command->setName`.

But Phinx command breaks backward compatibility with parent class `Symfony\Command` which allow set command name via constructor. This behaviour is unexpected.

JayPHP:
Thanks @FallDi !

PR #1340, opened 2018-04-29
Fix tests breaking due to double column type check in PostgresAdapter

SeinopSys:
This is a follow-up to #1175 which apparently started breaking tests. I think this might be enough to fix the issue.


lorenzo:
Thanks!

PR #1345, opened 2018-05-07
Documented the `getDependencies` method.

vimist:
Documented the `getDependencies` method with an example and a brief description of what it does and how to use it.


lorenzo:
Thanks!

PR #1350, opened 2018-05-13
Add option to addTimestamps with timezone column types

SeinopSys:
This PR adds a third argument to `Db\Table::addTimestamps` which lets the developer control whether the timestamp columns being added will retain the timezone information. I also fixed the indentation being all over the place in that method.

This is the cleanest solution I could think of to add support for changing this value, but I'm open to feedback.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -789,7 +789,8 @@ timezone enable or disable the ``with time zone`` option for ``time`` and ``time
 ======== ===========
 
 You can add ``created_at`` and ``updated_at`` timestamps to a table using the ``addTimestamps()`` method. This method also
-allows you to supply alternative names.
+allows you to supply alternative names. The optional third argument allows you to change the ``timezone`` option for the
+columns being added.
 
 .. code-block:: php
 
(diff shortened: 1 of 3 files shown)


lorenzo:
Do you mind adding a addTimestampsWithTimezone() method as an alias for passing the Boolean?

SeinopSys:
@lorenzo How's that look?

lorenzo:
Great!

I think we are only missing a simple test, if you know how to do it

SeinopSys:
@lorenzo Added a test. I wasn't sure if I should only check for the timestamp being true in the `testAddTimestampsWithTimezone` method or copy all checks over from `testAddTimestamps`, so I went with the latter, but I kinda feel it's pointless to test the same things twice. The end of the 2nd test could be reduced to ```php
        $this->assertEquals(true, $columns[0]->getTimezone());
        $this->assertEquals(true, $columns[1]->getTimezone());
```
in my opinion. Let me know if I should make that change,

lorenzo:
I think this is fine. Thanks for your great work!

Issue #1080, opened 2017-04-28
multi column addIndex with limit specifier


glensc:
i'm unable to produce index specifying length of individual fields. this is mysql table.

```mysql
CREATE TABLE `email_account` (
  `ema_folder` varchar(255) DEFAULT NULL,
  `ema_hostname` varchar(255) NOT NULL DEFAULT '',
  `ema_username` varchar(64) NOT NULL DEFAULT '',
  UNIQUE KEY `ema_username` (`ema_username`,`ema_hostname`(100),`ema_folder`(100))
);
```

i tried:

```php
->addIndex(['ema_username', 'ema_hostname', 'ema_folder'], ['unique' => true, 'limit' => 100])
```

however that tries to execute:

```mysql
UNIQUE KEY (`ema_username`,`ema_hostname`,`ema_folder`(100)),  KEY (`ema_prj_id`))
```

which in my case fails with mysql engine:

```
  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 1000 bytes  
```

glensc:
looks like the current `Index` structure does not allow per-column limit, so it's a bit bigger work to handle this.

https://github.com/robmorgan/phinx/blob/v0.8.0/src/Phinx/Db/Table/Index.php#L66

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

BLaDZer:
I think this should be closed cuz it fixed long time ago in v0.10.0

PR #1351, opened 2018-05-16
Allow multi column addIndex with limit specifier #1080

gamalan:
Allow per-column limit for index, only for mySQL (other DB adapter doesn't allow index length), example :
```php
                $table = $this->table( 'test_foo_bar' );
		$table->addColumn( 'column1', 'string' )
		      ->addColumn( 'column2', 'string' )
		      ->addColumn( 'column3', 'string' )
		      ->addIndex( [ 'column1', 'column2', 'column3' ], [ 'limit' => [0,10]] )
		      ->create();
```
Will resulting to multi column index, where column2 is having length of 10, column1 doesn't have any specific length, because it was defined 0. Meanwhile, column3 also doesn't have any specific length, due to undefined index limit.


lorenzo:
It is difficult to review your changes with all the extra white space added to it. Can you re-submit without all that?

gamalan:
Oh, wait, let me close it and fix it.

PR #1352, opened 2018-05-16
Allow multi column addIndex with limit specifier #1080

gamalan:
Allow per-column limit for index, only for mySQL (other DB adapter doesn't allow index length), example :
```php
                $table = $this->table( 'test_foo_bar' );
		$table->addColumn( 'column1', 'string' )
		      ->addColumn( 'column2', 'string' )
		      ->addColumn( 'column3', 'string' )
		      ->addIndex( [ 'column1', 'column2', 'column3' ], [ 'limit' => ['column2'=>10]] )
		      ->create();
```
Will resulting to multi column index, where column2 is having length of 10, column1 doesn't have any specific length, because it was defined 0. Meanwhile, column3 also doesn't have any specific length, due to undefined index limit.  

Edit :
- Change to column name when defining limit length


lorenzo:
I think something like

```
$table = $this->table( 'test_foo_bar' );
		$table->addColumn( 'column1', 'string' )
		      ->addColumn( 'column2', 'string' )
		      ->addColumn( 'column3', 'string' )
		      ->addIndex( [ 'column1', 'column2', 'column3' ], [ 'limit' => [''column1' => 0, 'column3' => 10]] )
		      ->create();
```

is more readable. Is it possible to do this instead?

gamalan:
@lorenzo yes, it is. let me update it.

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1035,7 +1032,25 @@ protected function getIndexSqlDefinition(Index $index)
             $def .= ' `' . $index->getName() . '`';
         }
 
-        $def .= ' (`' . implode('`,`', $index->getColumns()) . '`' . $limit . ')';
+        if (! is_array($index->getLimit())) {
```
dereuromark: CS: `if (!is_array(` etc.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -849,6 +849,48 @@ public function testAddIndexWithLimit()
...
+    }
+
+    public function testAddSingleIndexesWithLimitSpecifier()
+    {
+        $table = new \Phinx\Db\Table('table1', [], $this->adapter);
+        $table->addColumn('email', 'string')
+            ->addColumn('username', 'string')
+            ->save();
+        $this->assertFalse($table->hasIndex('email'));
+        $table->addIndex('email', ['limit' => [ 'email' => 3, 2 ]])
```
lorenzo: Is the number 2 at the end used for anything?

gamalan: No, just making sure it was running normally. When used like that, number 2 should be ignored.

lorenzo:
@gamalan Great work!

Do you mind adding something to the docs about this new feature?

gamalan:
@lorenzo Ok, where i should add the docs? Should I add after this part?

> The MySQL adapter also supports fulltext indexes. If you are using a version before 5.6 you must ensure the table uses the MyISAM engine.

lorenzo:
Thanks, this is great!

PR #1357, opened 2018-05-26
Added a Query builder through the cakephp/database package

lorenzo:
Support for a more feature reach query builder has been asked for
multiple times. Creating query builders is difficult, so better use
one that already exists.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -55,8 +55,6 @@ Phinx automatically creates a skeleton migration file with a single method:
              *    addIndex
              *    addForeignKey
              *
-             * Remember to call "create()" or "update()" and NOT "save()" when working
-             * with the Table class.
              */
             public function change()
             {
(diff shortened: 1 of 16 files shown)


lorenzo:
Merging this as there no feedback for over a week and would like to make a release soon

rquadling:
Any reason why the notes about `change()` have been removed from the documentation? Doesn't seem to relate to this feature.

docs/migrations.rst
```diff
@@ -55,8 +55,6 @@ Phinx automatically creates a skeleton migration file with a single method:
              *    addIndex
              *    addForeignKey
              *
-             * Remember to call "create()" or "update()" and NOT "save()" when working
```
rquadling: Why has this been removed?

docs/migrations.rst
```diff
@@ -125,13 +123,6 @@ Please be aware that when a ``change`` method exists, Phinx will automatically
 ignore the ``up`` and ``down`` methods. If you need to use these methods it is
 recommended to create a separate migration file.
 
-.. note::
```
rquadling: Why has this been removed?

rquadling:
Ah. I'm late on the merge, but still, those notes are important.

docs/migrations.rst
lorenzo: I forgot to remove them in my previous PR, this is not required anymore

docs/migrations.rst
lorenzo: Because it is not true anymore after my previous PR

rquadling: Ha! I'll shut up then!

Issue #1367, opened 2018-06-13
$this->dropTable('my_table');  does not work


F-JJTH:
Hello,
I'm trying to setup a simple migration script with up() and down() function.
The up() is working fine but the down() does not do anything.

### Short story:
My up() function is creating a table, after ./phinx migrate, then I can see my new table in my database => all is OK
My down function() is deleting the table with $this->dropTable('my_table');, then the table is still present in my database => it's not OK, I want my table to be dropped as I expect the dropTable() is doing that.

### Long story:
My migration script looks like:
```
use Phinx\Migration\AbstractMigration;

class CreateMyTable extends AbstractMigration
{
    /**
     * Migrate Up.
     */
    public function up()
    {
        $this->table('my_table', ['id' => false, 'primary_key' => 'C_Role', 'engine' => 'MyISAM'])
             ->addColumn('C_Role', 'integer', ['identity' => true])
             ->addColumn('T_Role', 'string', ['limit' => 45, 'default' => NULL, 'null' => true])
             ->addColumn('T_Role_Libe', 'string', ['limit' => 255, 'default' => NULL, 'null' => true])
             ->save();

        $rows = [
            [
              'C_Role' => 1,
              'T_Role' => 'CHT',
              'T_Role_Libe' => 'Foo',
            ],
            [
              'C_Role' => 2,
              'T_Role' => 'EXE',
              'T_Role_Libe' => 'Bar',
            ],
        ];

        $this->insert('my_table', $rows);
    }

    /**
     * Migrate Down.
     */
    public function down()
    {
        $this->dropTable('my_table');
    }
}
```

The command line result looks fine:
```
me@server:~$ ./phinx migrate
Phinx by CakePHP - https://phinx.org. 0.10.0

using config file ./bin/phinx.php
using config parser php
using migration paths 
 - /bdd/migrations
using seed paths 
 - /bdd/seeds
warning no environment specified, defaulting to: development
using database my_database

 == 20180612082039 CreateMyTable: migrating
 == 20180612082039 CreateMyTable: migrated 0.0723s

All Done. Took 0.0886s
me@server:~$ ./phinx rollback
Phinx by CakePHP - https://phinx.org. 0.10.0

using config file ./bin/phinx.php
using config parser php
using migration paths 
 - /bdd/migrations
using seed paths 
 - /bdd/seeds
warning no environment specified, defaulting to: development
using database my_database
ordering by creation time

 == 20180612082039 CreateMyTable: reverting
[...]

PR #1370, opened 2018-06-13
Making dropTable work again and deprecating it

lorenzo:
There is little point of having yet another way of dropping a table. Migrations should happen only through the Table class.

refs #1367


dereuromark:
It is a nice convenience method however, that doesnt need anything else.

lorenzo:
@dereuromark I'm questioning here the need to drop `NOW!!!`, by doing so you deprive phinx the ability to do a better job for you

PR #1371, opened 2018-06-13
Allow file suffix db adapter option for sqlite

### LINKED ISSUE #1245: SQLite create/dropDatabase
Both of these don't do what I'd expect:

createDatabase takes the name, and always appends `.sqlite3` as a file suffix. However, SQLite doesn't care what the file is called, many also use just use `.db` .
Suggestion: allow an optional suffix that replaces the .sqlite3 (second parameter for the function?)

dropDatabase is worse as in: It doesn't work.
The function does a simple `unlink` (also requiring the file suffix to be `.sqlite3`). However, this always fails (at least on Windows). I assume this is because phinx itself still has a file lock on the database while it does migrations?
Suggestion: Not sure if there's a way to solve this one, even? Can Phinx remove the file lock and destroy the database while it still is running?

markkimsal:
The file extension was hardcoded to ".sqlite3", this allows for
overriding that in the phinx.yml options for database adapters.

Partially fixes #1245

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -304,6 +304,7 @@ Declaring an SQLite database uses a simplified structure:
         development:
             adapter: sqlite
             name: ./data/derby
+            suffix: ".db"    # Defaults to ".sqlite3"
         testing:
             adapter: sqlite
             memory: true     # Setting memory to *any* value overrides name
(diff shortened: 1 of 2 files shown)


markkimsal:
There was one other issue in that the `createDatabase()` command would properly use the suffix to control "3rd party" databases, but not the one actively used by phinx.  The `connect()` method would use it's own naming standard to find or create a new file.

First, it would assume no file suffix, then it would check if a file with the suffix exists and use that file with the suffix.  This like an unnecessary and non-deterministic check now that the file extension can be explicitly controlled.

lorenzo:
great job!

PR #1372, opened 2018-06-13
Add deprecated warning to insert() and trigger_error on both

MasterOdin:
Building off https://github.com/cakephp/phinx/issues/1367#issuecomment-396966171 and the previous commits of today on dropTable.

This adds a deprecated tag on insert() as well (to match dropTable's) and has both methods trigger a `E_USER_DEPRECATED` warning so as to be explicit about this going away (and to follow PHP's model of having deprecated functions issuing said warning).

I also added version information to the deprecated tag to show when it was deprecated. You may also want to add info when the functions will be removed.

The releases page probably should also be updated to reflect that these two functions are deprecated.

### DIFF
--- a/tests/Phinx/Migration/AbstractMigrationTest.php
+++ b/tests/Phinx/Migration/AbstractMigrationTest.php
@@ -190,7 +190,15 @@ class AbstractMigrationTest extends TestCase
         $table = new Table('testdb', [], $adapterStub);
 
         $migrationStub->setAdapter($adapterStub);
-        @$migrationStub->insert($table, ['row' => 'value']);
+        if (PHP_VERSION_ID < 70000) {
+            $class = \PHPUnit_Framework_Error_Deprecated::class;
+        }
+        else {
+            $class = \PHPUnit\Framework\Error\Deprecated::class;
+        }
+        $class::$enabled = false;
+        $migrationStub->insert($table, ['row' => 'value']);
+        $class::$enabled = true;
     }
 
     public function testInsertString()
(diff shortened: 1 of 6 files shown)


tests/Phinx/Migration/AbstractMigrationTest.php
```diff
@@ -190,7 +190,33 @@ public function testInsert()
...
+            ->getMock();
+        $adapterStub->expects($this->once())
+            ->method('bulkinsert');
+
+        $migrationStub->setAdapter($adapterStub);
+        @$migrationStub->insert('testdb', ['row' => 'value']);
+    }
+
+    /**
+     * @expectedException \PHPUnit\Framework\Error\Deprecated
```
lorenzo: Seems like this class does not exist for older phpunit :(

I'm ok with not having this test, as it does not provide a lot of value to us

MasterOdin:
Got the tests to pass, but stickler-ci complains about the `@` usage, but it's necessary so that the deprecated warnings don't get pasted into the console when running the test suite. I could disable error_reporting and log_errors around the block as well as make PHPUnit ignore deprecated warnings, but that seems overkill for something that will just get removed at some point. Could also just toss out the tests I suppose.

tests/Phinx/Migration/AbstractMigrationTest.php
MasterOdin: Yeah, had to move this check inside of the test to detect PHP version to use proper class name. Works now.

lorenzo:
Thanks so much for helping@MasterOdin, I think we can disregard stickler-ci for the time being as the code and those tests is going away in the next release.

PR #1375, opened 2018-06-16
Warn when up or down will be ignored

### LINKED ISSUE #1329: When up() and change() is present, up() is silently ignored
My workmate just ran into a case where change() was present and up(), if this happens only change() is executed and up() ignored - silently.

I'm OK that phinx is just running change(), but before running it at all, it should check if up() or down() is present as well and throw an exception and not run the migrations silently without telling you that something was *not* executed.

@netstyler

markkimsal:
The presence of the change() method ignores individual up() or down()
methods, but there is no warning to end users.  The preFlightCheck for
AbstractMigration allows the opportunity to warn about potential
conflicts.

Fixes #1329


lorenzo:
Thanks!

burzum:
Thanks for the PR, but just a warning? IMHO the migration should not run at all without fixing the issue?

PR #1376, opened 2018-06-17
Remove up/down functions from abstract

MasterOdin:
With the change from #1375, all migrations that only use change throw a warning regardless of whether or not the migration file had an up/down function. This is because those functions always exist due to the AbstractMigration file. Given that the change function is not in the AbstractMigration (or MigrationInterface), to keep things consistent (and not show this warning always), can just remove the functions.

### DIFF
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -112,20 +112,6 @@ abstract class AbstractMigration implements MigrationInterface
     /**
      * {@inheritdoc}
      */
-    public function up()
-    {
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function down()
-    {
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function setAdapter(AdapterInterface $adapter)
     {
         $this->adapter = $adapter;
--- a/src/Phinx/Migration/MigrationInterface.php
+++ b/src/Phinx/Migration/MigrationInterface.php
@@ -55,20 +55,6 @@ interface MigrationInterface
     const DOWN = 'down';
 
     /**
-     * Migrate Up
-     *
-     * @return void
-     */
-    public function up();
-
-    /**
-     * Migrate Down
-     *
-     * @return void
-     */
-    public function down();
-
-    /**
      * Sets the database adapter.
      *
      * @param \Phinx\Db\Adapter\AdapterInterface $adapter Database Adapter
--- a/tests/Phinx/Migration/AbstractMigrationTest.php
+++ b/tests/Phinx/Migration/AbstractMigrationTest.php
@@ -7,20 +7,6 @@ use PHPUnit\Framework\TestCase;
 
 class AbstractMigrationTest extends TestCase
 {
-    public function testUp()
-    {
-        // stub migration
-        $migrationStub = $this->getMockForAbstractClass('\Phinx\Migration\AbstractMigration', ['mockenv', 0]);
-        $this->assertNull($migrationStub->up());
-    }
-
-    public function testDown()
-    {
-        // stub migration
-        $migrationStub = $this->getMockForAbstractClass('\Phinx\Migration\AbstractMigration', ['mockenv', 0]);
-        $this->assertNull($migrationStub->down());
-    }
-
     public function testAdapterMethods()
     {
         // stub migration


lorenzo:
This makes a lot of sense!

I was thinking if we could instead offer 2 diffrent abstract classes, one with `change` and another with the up/down methods. So that it is clear that we support both modes but they are incompatible. What do you think?

rquadling:
The `https://github.com/cakephp/phinx/blob/master/src/Phinx/Migration/Migration.template.php.dist` template no longer mentions of `up()`/`down()`, but there will be plenty of migrations in the wild that do use `up()`/`down()` and have never had a `change()` anywhere near them.

I've now got over 50 warnings from my pipeline relating to a few cases where we have used `change()` (yes, `50` _AND_ `a few` in the same sentence ... 4 years of development gives a LOT of migrations).

The vast majority of our migrations don't use `change()`.

MasterOdin:
Making two different abstract migration classes seems like it'd also be confusing to end users in knowing they have to change the method (from change to up/down) and the abstract migration they have to use, and of course it doesn't stop people from copy/pasting an old migration, changing the method, and then getting the warning confusingly because they didn't realize they needed to change the abstract method.

lorenzo:
Right, I keep forgetting that php has not compilation step were you get the warnings first. I would not like to get rid forever of `up` and `down`, so I do think we need a way to support both but make it clear that you cannot combine them

MasterOdin:
You don't actually need the functions in the interface/abstract to actually use them though? And given that the function checks (and calling for up/down) are done dynamically based on naming from (https://github.com/cakephp/phinx/blob/master/src/Phinx/Migration/MigrationInterface.php#L42-L55), I'm not sure you need to have the concrete functions defined beforehand. There's not a lot of value of having an interface that all implementing classes will probably just have empty stubs for as the concrete implementations may use just the change, just the up, just the down, or both up/down and never all three. I think just documenting it is probably good enough (add in the class comment on AbstractMigration or MigrationInterface about the functions and in the site documentation).

rquadling:
If the signature was `function change(MigrationDirection $direction){}`, where `MigrationDirection` was an enum class (I use `\Eloquent\Enumeration\AbstractEnumeration` for my enum classes), then there would be a migration path from `up()`/`down()` to `change()`  for those that needed to clean up old migrations.

Maybe a better fix would be to see what the calling class of the `up()` and `down()` method is. If it is the base abstract class, then this is NOT a user implemented method and so no warning is necessary.

MasterOdin:
The purpose of `function change(MigrationDirection $direction)` being that I could then actually completely drop up/down as the conversion of say:
```
class Test extends AbstractMigration
{
    /**
     * Migrate Up.
     */
    public function up()
    {
        $this->table('my_table', ['id' => false, 'primary_key' => 'C_Role', 'engine' => 'MyISAM'])
            ->addColumn('C_Role', 'integer', ['identity' => true])
            ->addColumn('T_Role', 'string', ['limit' => 45, 'default' => NULL, 'null' => true])
            ->addColumn('T_Role_Libe', 'string', ['limit' => 255, 'default' => NULL, 'null' => true])
            ->save();

        $rows = [
            [
                'C_Role' => 1,
                'T_Role' => 'CHT',
                'T_Role_Libe' => 'Foo',
            ],
            [
                'C_Role' => 2,
                'T_Role' => 'EXE',
                'T_Role_Libe' => 'Bar',
            ],
        ];

        $this->insert('my_table', $rows);
    }

    /**
     * Migrate Down.
     */
    public function down()
    {
        $this->table('my_table')->drop()->save();
    }
}
```

to be:
```
class Test extends AbstractMigration
{
    public function change(MigrationDirection $direction)
    {
        $this->table('my_table', ['id' => false, 'primary_key' => 'C_Role', 'engine' => 'MyISAM'])
            ->addColumn('C_Role', 'integer', ['identity' => true])
            ->addColumn('T_Role', 'string', ['limit' => 45, 'default' => NULL, 'null' => true])
            ->addColumn('T_Role_Libe', 'string', ['limit' => 255, 'default' => NULL, 'null' => true])
            ->create();
        if ($direction === MigrationDirection::UP) {
            $rows = [['C_Role' => 1], ['C_Role' => 2]];
            $this->insert('my_table', $rows);
        }   
    }
}
```

rquadling:
With the `change()` method having the `$direction`, I can refactor my current migration templates and base easily enough.

MasterOdin:
Yeah, I'm in support of adding that parameter and doing so wouldn't break anyone's code (and I'll look into making a PR for it as it resolved my only pain point). It doesn't resolve the issue with the warning though.

rquadling:
The pre-flight check should halt the migration if there are some failed conditions.

MasterOdin:
I think it might make sense to have that be a separate issue to be done in a separate PR given that we've got this PR and #1379 with conflicting solution on this. I maintain that having an interface where any abstract implementation of it will minimally define the functions to be empty makes them a somewhat pointless addition since the whole point I guess is to be able to give them a docstring but you could probably achieve a similar effect via a class docstring. Doubly so given that you have a method to generate a shell of a migration for you (with a change function).

lorenzo:
thanks @MasterOdin

Issue #574, opened 2015-06-17
Dropping FK without constraint name causes unwanted dropping of PK in Postgres


Stene3:
When trying to drop the foreign key without passing the constraint name, all constraints connected with the specified columns are dropped. Works for foreign key constraints, but primary constraints are dropped too.

When listing all constraints for the current table and specified column, there should be another check for the constraint type of "FOREIGN KEY".

dereuromark:
Are you able to provide a pull request with your suggested changes?

djcvijic:
I'm having this issue as well, so I assume it's not a negligibly rare one.

Since I'm not a contributor here and don't know what's the process for submitting a pull request, I'll share my proposed fix here. In `PostgresAdapter.php`:

```
protected function getDropForeignKeyByColumnsInstructions($tableName, $columns)
    {
        $instructions = new AlterInstructions();
        $parts = $this->getSchemaName($tableName);
        foreach ($columns as $column) {
            $rows = $this->fetchAll(sprintf(
                "SELECT CONSTRAINT_NAME
                FROM information_schema.KEY_COLUMN_USAGE
                WHERE TABLE_SCHEMA = %s
                AND TABLE_NAME IS NOT NULL
                AND TABLE_NAME = %s
                AND COLUMN_NAME = %s
                AND POSITION_IN_UNIQUE_CONSTRAINT IS NOT NULL
                ORDER BY POSITION_IN_UNIQUE_CONSTRAINT",
                $this->getConnection()->quote($parts['schema']),
                $this->getConnection()->quote($parts['table']),
                $this->getConnection()->quote($column)
            ));
            foreach ($rows as $row) {
                $newInstr = $this->getDropForeignKeyInstructions($tableName, $row['constraint_name']);
                $instructions->merge($newInstr);
            }
        }
        return $instructions;
    }
```

I've added `AND POSITION_IN_UNIQUE_CONSTRAINT IS NOT NULL` to the WHERE clause, because according to [this](https://www.postgresql.org/docs/8.2/static/infoschema-key-column-usage.html), the POSITION_IN_UNIQUE_CONSTRAINT should be NOT NULL if and only if the constraint is a foreign key constraint.

Let me know if you'll implement this, or how I could contribute the change myself.

Thanks!

lorenzo:
the process of submitting a pull request is to just submit the pull request and have all test passing, hopefully after adding a new test to verify that your fix actually works.

Could you help us by doing that?

[...]

PR #1378, opened 2018-06-17
ISSUE-574 Dropping FK without constraint name…

djcvijic:
Dropping FK without constraint name causes unwanted dropping of PK in Postgres

Issue: https://github.com/cakephp/phinx/issues/574


lorenzo:
Thanks!

Issue #1038, opened 2017-02-15
Postgresql - seeding boolean column gets: Invalid text representation for false value


sergioalcaraz:
Hi, I get this error when run seeder:

`Error: [PDOException] SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type boolean: ""`

Under:
- PHP 7.0.13
- Phinx version 0.6.5
- CakePHP 3.3.15
- Postgresql 9.5

The error only occurs when the value is false
`['in_offer' => false]`

Migrations for the column is:
```
$table->addColumn('in_offer', 'boolean', [
    'default' => false,
    'null' => false,
]);
```

SQL from pgadmin
```
ALTER TABLE public.products ADD COLUMN in_offer boolean;
ALTER TABLE public.products ALTER COLUMN in_offer SET NOT NULL;
ALTER TABLE public.products ALTER COLUMN in_offer SET DEFAULT false;
```

mariorez:
I get the same error running the seeder command in MySQL adapter:
```
[PDOException (HY000)]                                                                                
SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '' for column 'is_user_admin' at row 1
```
The error only occurs when the value is `false`.

```
Exception trace:
 () at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php:392
 PDOStatement->execute() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php:392
 Phinx\Db\Adapter\PdoAdapter->insert() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Db/Table.php:651
 Phinx\Db\Table->saveData() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Db/Table.php:639
 Phinx\Db\Table->update() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Db/Table.php:675
 Phinx\Db\Table->save() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Seed/AbstractSeed.php:197
 Phinx\Seed\AbstractSeed->insert() at /my-app-path/db/seeds/Seeder003UserCompanyGroup.php:43
 Seeder003UserCompanyGroup->run() at /my-app-path/vendor/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php:154
[...]

PR #1380, opened 2018-06-19
ISSUE-1038 Postgresql - seeding boolean column

djcvijic:
Postgresql - seeding boolean column gets: Invalid text representation for false value

https://github.com/cakephp/phinx/issues/1038


lorenzo:
thanks!

PR #1382, opened 2018-06-19
Add firing of events when migration and rollback happen

sane4ek-2:
Sometimes we want do some things regularly after migrations or rollbacks. Before that changes we couldn't do that. But now we can make our own event subscribers and they will be called when one of the two commands call. Also we can refuse from subscribers and don't write them.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -379,6 +379,7 @@ abstract class AbstractCommand extends Command
      * Load all event subscribers
      *
      * @throws \ReflectionException
+     * @return void
      */
     protected function loadEventSubscribers()
     {
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -38,6 +38,10 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 /**
  * Abstract command, contains bootstrapping info
@@ -72,6 +76,11 @@ abstract class AbstractCommand extends Command
     protected $manager;
 
     /**
+     * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
+     */
+    protected $dispatcher;
+
+    /**
      * {@inheritdoc}
      */
     protected function configure()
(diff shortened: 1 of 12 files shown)


src/Phinx/Console/CommandEvents.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Phinx\Console;
+
+class CommandEvents
```
rquadling: I can certainly see use cases for phinx.command.[migrate|rollback|seed|breakpoint].[started|completed|failed]

Well. Maybe not so much breakpoint, but keep the pattern as then you don't have "special" behaviour.

Also, (maybe overkill), would firing an event for each migration/seed be possible?

phinx.command.[migrate-single|rollback-single|seed-single].[started|completed|failed]

The event data for the singles (not a great name admittedly, so be inventive!!!) would need to include the migration name.

One use case for this for me would be that I could activate an "after" migration once a "before" migration has been executed.

sane4ek-2: > I can certainly see use cases for phinx.command.[migrate|rollback|seed|breakpoint].[started|completed|failed]

I do that.

> Also, (maybe overkill), would firing an event for each migration/seed be possible?
> phinx.command.[migrate-single|rollback-single|seed-single].[started|completed|failed]
> The event data for the singles (not a great name admittedly, so be inventive!!!) would need to include the migration name.

If be honest I don't see any use case right now. I suggest not to develop that at present time.

> One use case for this for me would be that I could activate an "after" migration once a "before" migration has been executed.

You're right. Let's do it.

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -362,4 +374,31 @@ protected function getSeedTemplateFilename()
...
+        }
+        $finder = new Finder();
+        $listeners = $finder->in($paths)->name('*.php')->files();
+
+        /** @var SplFileInfo $listener */
+        foreach ($listeners as $listener) {
+            require_once $listener->getRealPath();
+
+            $refClass = new \ReflectionClass($listener->getBasename('.php'));
+            if (!$refClass->implementsInterface('\Symfony\Component\EventDispatcher\EventSubscriberInterface')) {
```
dereuromark: Would be better to use ::class usage here.

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -362,4 +374,31 @@ protected function getSeedTemplateFilename()
...
+    protected function loadEventSubscribers()
+    {
+        $paths = $this->getConfig()->getSubscriberPaths();
+        if (!$paths) {
+            return;
+        }
+        $finder = new Finder();
+        $listeners = $finder->in($paths)->name('*.php')->files();
+
+        /** @var SplFileInfo $listener */
```
dereuromark: FQCN missing

dersam:
I'd love to be able to define events around migrations - I have a few use cases that would be solved by an event system. What still needs to be done to get this to a point where it would be mergeable? I'd be willing to finish it off if @sane4ek-2 isn't free to complete it.

sane4ek-2:
@dersam it would be great if you could complete those changes. I forget about this PR and don't have enough time.

dereuromark:
ping @dersam @sane4ek-2

othercorey:
Closing due to no activity.

PR #1387, opened 2018-06-23
Add setPrecisionAndScale

garas:
Precision and scale usually used together. Add setter to set both parameters at once, so they look like in SQL datatype `DECIMAL(11, 3)`. This also helps prevent from mixing precision and scale.

### DIFF
--- a/src/Phinx/Db/Table/Column.php
+++ b/src/Phinx/Db/Table/Column.php
@@ -372,6 +372,21 @@ class Column
     }
 
     /**
+     * Sets the column precision and scale for decimal.
+     *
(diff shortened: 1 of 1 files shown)


lorenzo:
Do you mis adding the missing comments?

garas:
Whole file is missing parameter descriptions.
I've added descriptions.

lorenzo:
@garas since I personally find confusing f what part is the scale and what is the precision, could you include a short example in the method description showing which is which using the DECIMAL(10,3) example?

lorenzo:
Thanks!

Issue #802, opened 2016-03-04
Cannot update table comment


WillGibson:
It's quite possible I'm doing something wrong, but I cannot get a table comment to update with this code...

```
$table = $this->table('some_table');
$table->setOptions(['comment' => 'No longer in use by the Something app']);
$table->save();
```

Is it me, or is this a bug?

catharsisjelly:
Not a bug I think just not an implemented feature https://github.com/robmorgan/phinx/blob/0.5.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L251 allows you to add options which get done during table creation but there is no ALTER for just the table as far as I can see

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

WillGibson:
@dereuromark I can't just now, all my spare time is being sucked up by looking for a new job, which is good really, I think.

othercorey:
Added in https://github.com/cakephp/phinx/pull/1393

Issue #335, opened 2014-10-16
Add a primary key on an existing table


liawlim:
Hello guys,

How we can add a primary key on an existing table ?

I try to do this but the table dont be updated :

``` php
    public function up()
    {
        $this->table('toto', array('id' => true))
            ->save();
    }
```

But Table::update() method was call and this method didn't create primary key ...

sergejkaravajnij:
I have exactly the same task and not found any solution except using ALTER TABLE ... :(
But in this case I couldn't use _change()_ method.

DerMika:
You should be able to create a primary key by creating an index of type "PRIMARY" just as you can create a unique key. However, the current implementation only looks at the UNIQUE type. 

I was expecting `$table->addIndex(['someId', 'someOtherId'], ['type' => 'PRIMARY'])` to work, but I found out that any index type besides 'unique' is ignored.

With that said: here is my vote for fixing this bug.

berarma:
+1

daniel-gomes-sociomantic:
+1

purpleweb:
+1

burzum:
Can we add a method `addPrimaryKey()` :)

lorenzo:
Thats. correct. Thanks for pointing that out @djcvijic

PR #1393, opened 2018-06-26
Implement Table::change

djcvijic:
Addresses requests for being able to modify the primary key, as well as the table comment and other table options. Implemented and tested for all adapters. Check migrations.rst for explanation of how it works. Feedback is welcome!

Affected issues:
https://github.com/cakephp/phinx/issues/335
https://github.com/cakephp/phinx/issues/802

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -539,6 +539,55 @@ Option    Description
 comment   set a text comment on the table
 ========= ===========
 
+Changing Table Options
+~~~~~~~~~~~~~~~~~~~~~~
+
+To change the options on an existing table, use the ``change()`` method.
+Options ``id`` and ``primary_key`` are used to determine the primary key on the table, same as when creating a new table.
+Also see `Creating a Table`_ for lists of additional supported options for each adapter.
+
+In the following example, a table is initially created with the default primary key, no table comment, and two columns.
+Then another column is added, and the ``change()`` method is used to remove the default primary key,
+set the 'new_id' column to be the new primary key, and add a new comment on the table.
+
+.. code-block:: php
+
+        <?php
+
+        use Phinx\Migration\AbstractMigration;
+
+        class MyNewMigration extends AbstractMigration
+        {
+            /**
+             * Migrate Up.
+             */
+            public function up()
+            {
+                $users = $this->table('users');
+                $users
+                    ->addColumn('username', 'string', ['limit' => 20])
+                    ->addColumn('password', 'string', ['limit' => 40])
+                    ->save();
+
+                $users
+                    ->addColumn('new_id', 'integer', ['null' => false])
+                    ->change([
+                        'id' => false,
+                        'primary_key' => 'new_id',
+                        'comment' => 'Some comment',
+                    ])
+                    ->save();
+            }
+
+            /**
+             * Migrate Down.
+             */
+            public function down()
+            {
+
+            }
+        }
+
 Valid Column Types
 ~~~~~~~~~~~~~~~~~~
 
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -348,6 +348,15 @@ interface AdapterInterface
     public function createTable(Table $table, array $columns = [], array $indexes = []);
 
     /**
+     * Creates the specified database table.
+     *
+     * @param \Phinx\Db\Table\Table $table Table
+     * @param array $newOptions New options for the table
+     * @return void
+     */
+    public function changeTable(Table $table, array $newOptions);
+
+    /**
      * Truncates the specified table
      *
      * @param string $tableName
@@ -389,6 +398,16 @@ interface AdapterInterface
      * @return bool
      */
     public function hasIndexByName($tableName, $indexName);
+
+    /**
+     * Checks to see if the specified primary key exists.
+     *
+     * @param string   $tableName
+     * @param string[] $columns    Column(s)
+     * @param string   $constraint Constraint name
+     * @return bool
+     */
+    public function hasPrimaryKey($tableName, $columns, $constraint = null);
 
     /**
      * Checks to see if a foreign key exists.
--- a/src/Phinx/Db/Adapter/AdapterWrapper.php
+++ b/src/Phinx/Db/Adapter/AdapterWrapper.php
@@ -236,7 +236,7 @@ abstract class AdapterWrapper implements AdapterInterface, WrapperInterface
     }
 
     /**
-     * @inheritDoc
+     * {@inheritdoc}
      */
     public function toggleBreakpoint(MigrationInterface $migration)
     {
@@ -246,7 +246,7 @@ abstract class AdapterWrapper implements AdapterInterface, WrapperInterface
     }
 
     /**
-     * @inheritDoc
+     * {@inheritdoc}
      */
     public function resetAllBreakpoints()
     {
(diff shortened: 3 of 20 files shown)


docs/migrations.rst
```diff
@@ -539,6 +539,55 @@ Option    Description
...
+            {
+                $users = $this->table('users');
+                $users
+                    ->addColumn('username', 'string', ['limit' => 20])
+                    ->addColumn('password', 'string', ['limit' => 40])
+                    ->save();
+
+                $users
+                    ->addColumn('new_id', 'integer', ['null' => false])
+                    ->change([
```
lorenzo: I am wondering if we could have specific methods, instead having an array interface. Something like

```
$users->changePrimaryKey()
$users->changeComment()
```

Internally those method could just use `change()`. What do you think?

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -298,6 +298,103 @@ public function createTable(Table $table, array $columns = [], array $indexes =
...
+                    "Tried to create primary key column %s for table %s, but that column already exists",
+                    $this->quoteColumnName($newOptions['primary_key']),
+                    $this->quoteTableName($table->getName())
+                ));
+            }
+
+            $column = new Column();
+            $column
+                ->setName($newOptions['primary_key'])
+                ->setType('integer')
```
lorenzo: How do you know the type is integer?

lorenzo: This makes me think that perhaps changing the primary key should provide a column object instead

docs/migrations.rst
djcvijic: It came across my mind too. On one hand, change() provides a similar interface to Migration::table(), allowing users to change these options in a similar way as when creating the table, and consistency like that is always nice. On the other hand changePrimaryKey and changeComment would allow the user to be more deliberate and precise, make the methods easier to find, and probably leading to fewer errors too. Another option would be to just provide both interfaces, but I haven't thought much about it yet.

I somehow decided on change() initially, until I got feedback from the owners. So it's your call in the end, once you're sure about the approach let me know and I'll rework it in the next few days.

lorenzo: I thought about it for a bit and I think having separate methods is better, can you go that way?

djcvijic: Can do! I'll get to it during the weekend probably.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -347,6 +347,24 @@ public function hasTable($tableName);
...
+    public function changePrimaryKey(Table $table, $newColumns);
+
+    /**
+     * Changes the comment of the specified database table.
+     *
+     * @param Table $table Table
+     * @param string|null $newComment New comment string, or null to drop the comment
+     * @return void
+     */
+    public function changeComment(Table $table, $newComment);
```
lorenzo: This function and the one above should be moved to `DirectActionInterface`. The idea is to keep `AdapterInterface` small and all those direct actions should live in the tother interface: https://github.com/djcvijic/phinx/blob/74341755a952804e8698b18d0eee9c76c193a117/src/Phinx/Db/Adapter/DirectActionInterface.php

src/Phinx/Db/Adapter/AdapterWrapper.php
```diff
@@ -349,6 +349,22 @@ public function createTable(Table $table, array $columns = [], array $indexes =
...
+    {
+        $this->getAdapter()->changePrimaryKey($table, $newColumns);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function changeComment(Table $table, $newComment)
+    {
+        $this->getAdapter()->changeComment($table, $newComment);
```
lorenzo: This method and the one above should be guarded by an `if` condition, checking that the adapter is `DirectActionInterface`

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -298,6 +298,63 @@ public function createTable(Table $table, array $columns = [], array $indexes =
...
+    protected function getChangePrimaryKeyInstructions(Table $table, $newColumns)
+    {
+        $instructions = new AlterInstructions();
+
+        // Drop the existing primary key
+        $primaryKey = $this->getPrimaryKey($table->getName());
+        if (!empty($primaryKey['columns'])) {
+            foreach ($primaryKey['columns'] as $column) {
+                $sql = sprintf(
+                    'MODIFY %s INT NOT NULL',
```
lorenzo: I still don't get how hardcoding `INT` here is safe. Can you explain?

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -248,6 +248,44 @@ public function createTable(Table $table, array $columns = [], array $indexes =
...
+     * {@inheritdoc}
+     */
+    protected function getChangePrimaryKeyInstructions(Table $table, $newColumns)
+    {
+        $instructions = new AlterInstructions();
+
+        // Drop the existing primary key
+        $primaryKey = $this->getPrimaryKey($table->getName());
+        if (!empty($primaryKey)) {
+            $this->dropPrimaryKey($table->getName(), $primaryKey);
```
lorenzo: Can this be done through `$instructions`? It's not nice that it is being done immediately right now. The simplest way is to make `dropPrimaryKey` return another `AlterInstructions` object you can merge here. Then you can rename the function to `getDropPrimaryKeyInstructions`

src/Phinx/Db/Adapter/AdapterInterface.php
djcvijic: Will do!

But I guess I'll do so still not understanding what the difference is between those interfaces. AdapterInterface seems to be more for higher-level operations and non-modifying operations in general, whereas DirectActionInterface seems to have operations that modify existing tables in some way. And the explanation in a comment on DirectActionInterface does not help much either.
`Represents an adapter that is capable of directly executing alter instructions, without having to plan them first.`

If future contributors need to understand this better, it might be worth to clarify the difference.

src/Phinx/Db/Adapter/AdapterWrapper.php
djcvijic: I'm actually considering removing them from AdapterWrapper for consistency, since now that I'm looking at the difference between this class and DirectActionInterface, I'm noticing that this class doesn't have any of the other table modifications anyway.
Do you agree or should I strive to keep them as suggested?

src/Phinx/Db/Adapter/SQLiteAdapter.php
djcvijic: I remember it was painful to have to write it this way, but I was having some difficulty adding it to $instructions... I'll try again in this iteration since the code is simpler now too, and I'll ask here if I have trouble again.

Note to self: it's the same in SqlServerAdapeter!

djcvijic:
I appreciate the feedback!
And I look forward to contributing more, since I already have my eye on several other issues from the pile, that I'll get around to whenever I get enough free time :)

src/Phinx/Db/Adapter/AdapterInterface.php
lorenzo: To be honest `DirectActionInterface` are methods I don't want t all exposed, but they already existed and many depend on them, so I had to keep them.

The general idea is that all changes to the database should be planned, by crating `Action` classes. All methods in  `DirectActionInterface` directly modify the database, without going through a planning and validation step, so they cannot be optimised either.

src/Phinx/Db/Adapter/AdapterWrapper.php
lorenzo: I agree!

djcvijic:
@lorenzo Ping!

Issue #1392, opened 2018-06-26
Add MySQL FULLTEXT indexes one by one


garas:
InnoDB doesn't support adding multiple FULLTEXT at once.

Phinx v0.10 started grouping ALTER statements.

Phinx shouldn't group adding FULLTEXT indexes, should execute separate queries.

```php
$this->table('ttt')
    ->addIndex('b', [
        'type' => Index::FULLTEXT,
    ])
    ->addIndex('c', [
        'type' => Index::FULLTEXT,
    ])
    ->update();

// workaround - run separately each addIndex+update
```

```sql
CREATE TABLE `ttt` (
  `a` int,
  `b` text,
  `c` text
);


ALTER TABLE `ttt` 
    ADD FULLTEXT KEY (`b`),
    ADD FULLTEXT KEY (`c`);
-- Error #1795 - InnoDB presently supports one FULLTEXT index creation at a time


-- Phinx should execute separate queries
ALTER TABLE `ttt` ADD FULLTEXT KEY (`b`);
ALTER TABLE `ttt` ADD FULLTEXT KEY (`c`);
```

phpMyAdmin exports such tables as
```sql
ALTER TABLE `items`
  ADD PRIMARY KEY (`id`),
  ADD KEY `category_id` (`category_id`),
  ADD KEY `user_id` (`user_id`);
ALTER TABLE `items` ADD FULLTEXT KEY `name` (`name`);
ALTER TABLE `items` ADD FULLTEXT KEY `description` (`description`);
```

lorenzo:
Interesting, I did not know that. Would you like to submit a pull request to fix this?

garas:
I'll try

PR #1394, opened 2018-06-27
Add FULLTEXT indexes on MySQL one at a time

garas:
SQLSTATE[HY000]: General error: 1795 InnoDB presently supports one FULLTEXT index creation at a time

Ref #1392


lorenzo:
Thanks a lot!

PR #1396, opened 2018-06-29
Boolean fields couldn't have default NULL

garas:
`$table->addColumn('column1', 'boolean', ['default' => null, 'null' => true])` was added as `column1 tinyint(1) NOT NULL DEFAULT 0`


lorenzo:
Thanks!

PR #1397, opened 2018-07-02
Fix the addForeignKey() example.

ZebraNorth:
The addForeignKey() example doesn't work with MySQL because it uses the "SET NULL" option, but the field is not nullable.


lorenzo:
Thanks!

Issue #752, opened 2016-01-17
The command "php vendor/bin/phinx create MyNewMigration" is not working.


Brian1111:
It just shows the following.  I tried it on the latest version, i.e., version 0.5.1, and on version 0.4.4.  I am getting the following with both instead of getting a migration class.  I could get "phinx.yml" file; I entered the info for pgsql database.  I also could get a migrations folder which is empty.

   dir=$(d=${0%[/\\]*}; cd "$d"; cd '../robmorgan/phinx/bin' && pwd)

   # See if we are running in Cygwin by checking for cygpath program
   if command -v 'cygpath' >/dev/null 2>&1; then
           # Cygwin paths start with /cygdrive/ which will break windows PHP,
           # so we need to translate the dir path to windows format. However
           # we could be using cygwin PHP which does not require this, so we
           # test if the path to PHP starts with /cygdrive/ rather than /usr/bin
           if [[ $(which php) == /cygdrive/* ]]; then
                   dir=$(cygpath -m $dir);
            fi
    fi

   dir=$(echo $dir | sed 's/ /\ /g')
   "${dir}/phinx" "$@"

rquadling:
The content of ``vendor/bin`` are runnables. No need to use the PHP loader.

Use ``vendor/bin/phinx create xxxx`` where xxxx is the name of the migration you want.

Brian1111:
@rquadling I noticed that the tables were created in the directory db/migrations.  However, there is an empty migrations directory just after root directory; should I delete that?  The content of  phinx file is the same as above; I guess it should be like that.

Thanks for your help

rquadling:
The location of the migrations is softcoded. You can set it in your config file (``phinx.yml``, ``phinx.php``, etc.)

Brian1111:
Table 'users' and its columns are not being migrated to postgres database.  It says: "All Done. Took 0.0367s", but it doesn't show any message indicating migration.  Also, my database table is empty; no database table with the name 'users' is created.

rquadling:
Can you show your migration script?

Brian1111:
Here is the script, originally it was different, then I wrote the way to you wrote on your website:

    public function up()
    {
      $users = $this->table('users');
      $users->addColumn('first_name', 'string', array('limit' => 30))
            ->addColumn('last_name', 'string', array('limit' => 30))
            ->addColumn('email', 'string', array('limit' => 100))
            ->addColumn('password', 'string', array('limit' => 40))
[...]

PR #1398, opened 2018-07-03
Creating new Migration - without php in command // Update Docs

100Errors:
Hey,
as far as I know, is it not possible to create migrations with `php vendor/bin/phinx create MyNewMigration` but with `vendor/bin/phinx create MyNewMigration`

#752


lorenzo:
Thanks!

PR #1410, opened 2018-07-16
Introduce deprecation and todo annotations in migration template

rquadling:
Introduce deprecation of up() and down() and the support for $direction in change().
Introduce a todo for the users to refactor their existing migrations.

### DIFF
--- a/src/Phinx/Migration/Migration.template.php.dist
+++ b/src/Phinx/Migration/Migration.template.php.dist
@@ -3,6 +3,13 @@ $namespaceDefinition
 
 use $useClassName;
 
+/**
+ * $className migration
+ *
+ * @deprecated up() This method is deprecated and will be removed in a future release. Use change($direction).
+ * @deprecated down() This method is deprecated and will be removed in a future release. Use change($direction).
+ * @todo Refactor existing migrations from up()/down() to change($direction)
+ */
 class $className extends $baseClassName
 {
     /**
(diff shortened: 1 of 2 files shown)


src/Phinx/Migration/Migration.template.php.dist
```diff
@@ -3,6 +3,13 @@ $namespaceDefinition
 
 use $useClassName;
 
+/**
+ * $className migration
+ *
+ * @deprecated up() This method is deprecated and will be removed in a future release. Use change($direction).
```
dereuromark: Does this work? Or will IDEs show all class usage as deprecated? Then we better use an inline deprecation:

    * Deprecated up() ...

rquadling: Aha! Yes. The intent is to show only `up()` and `down()` are deprecated, not the migration itself.

I wonder if just passing `$direction` to `change()` would be enough first. Add a deprecation notice of `up()` and `down()` to the release notes and a new upgrade doc (next release) that documents a what's required to convert existing migrations from `up()`/`down()` to `change($direction)`, though it is pretty obvious.

dereuromark:
@lorenzo Is this fine to merge?

dereuromark:
Looking into the original issue, I do have to say that I agree with @lorenzo here
https://github.com/cakephp/phinx/issues/1411#issuecomment-405255135

>  I personally find it more clear to split the migration in two steps, if I know my steps cannot be automatically done for me in change

Indeed it seems that leaving them in allows a speaking way to have up/down used where change() cannot work without having to build in cyclic complexity (if/else) into the main change() method etc.

If there is no real drawback why not just keeping them?

PR #1416, opened 2018-08-01
Fix formatting in Foreign Keys section

reassured-jy:
A missing backtick was messing up the formatting at one point.


dereuromark:
Thank ypu.

PR #1419, opened 2018-08-01
fixe stdClass bug in MysqlAdapter

bernard-ng:
fixe bug, when $column is a stdClass instead of an Array


bernard-ng:
When using the PDO driver with the following attribute:
(PDO :: ATTR_DEFAULT_FETCH_MODE, PDO :: FETCH_OBJ);

and that in the phinx configuration we use a connection to a pdo instance with this attribute, this generates some bugs: "can not use stdClass as an Array" I thought it would affect two or three phinx files but I realize that no..
my pull request was a fast fixed, I cross that we must review that (the attributes of the instance of pdo when using phinx with a connection)

lorenzo:
A fix would be to set the correct attribute when creating the adapter object

PR #1425, opened 2018-08-10
Sq lite alter columns

JamesRItchie88:
Ability to add multiple columns whilst using sqlite adapter. unit tests added ^^


lorenzo:
Thanks so much!

PR #1432, opened 2018-09-04
Double support for Mysql

martijngastkemper:
The `MysqlAdapter` doesn't support the double column type. Phinx can't create a column with that type, but when Phinx is added to a project with existing tables it's possible.


martijngastkemper:
This PR is replaced by #1436 . That one uses `Literal` which is [the documented way](http://docs.phinx.org/en/latest/migrations.html?highlight=literal#custom-column-types-default-values) to handle DBMS specific columns.

PR #1433, opened 2018-09-06
Fix DSN parsing

watermark:
When only using a DSN, I received the error that no adapter was specified.  Found the DSN parser was never called, so the adapter was never set.


lorenzo:
any chance you can add a test so this doesn't break again?

PR #1434, opened 2018-09-06
remove unused use statement for Phinx\Db\Action\DropColumn;

watermark:
Phinx\Db\Action\DropColumn doesn't exist, remove the use statement in Phinx\Db\Table


lorenzo:
thanks!

PR #1436, opened 2018-09-11
Return Literal when column is not a Phinx type

martijngastkemper:
This is an alternative for PR #1432 and adds support for all relevant DBMS adapters.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -943,7 +943,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
      * @param string $sqlTypeDef
      * @throws \RuntimeException
      * @internal param string $sqlType SQL type
-     * @returns string Phinx type
+     * @return array Phinx type
      */
     public function getPhinxType($sqlTypeDef)
     {
@@ -1039,8 +1039,12 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
                     break;
             }
 
-            // Call this to check if parsed type is supported.
-            $this->getSqlType($type, $limit);
+            try {
+                // Call this to check if parsed type is supported.
+                $this->getSqlType($type, $limit);
+            } catch (\RuntimeException $e) {
+                $type = Literal::from($type);
+            }
 
             $phinxType = [
                 'name' => $type,
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -933,7 +933,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
 
                 return ['name' => 'year', 'limit' => $limit];
             default:
-                throw new \RuntimeException('The type: "' . $type . '" is not supported.');
+                throw new UnsupportedColumnTypeException('Column type "' . $type . '" is not supported by MySQL.');
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1036,8 +1036,12 @@ public function getPhinxType($sqlTypeDef)
                     break;
             }
 
-            // Call this to check if parsed type is supported.
-            $this->getSqlType($type, $limit);
+            try {
+                // Call this to check if parsed type is supported.
+                $this->getSqlType($type, $limit);
+            } catch (\RuntimeException $e) {
```
martijngastkemper: I don't like this try catch, but `getSqlType` in other adapters throw a `RuntimeException`. Changing this behaviour in the `MysqlAdapter` isn't consistent so I decided to use this way.

lorenzo: Yeah, it would be nicer if the exception was more specific

lorenzo:
This is a great solution! 

Any change you can add this to the other adapters?

martijngastkemper:
PostgreSQL already had support for `Literal`. I've added it to SQLlite and SQLServer.

martijngastkemper:
codecov/patch and Scrutinizer give weird results. Could it be that some force push commands break them? I had to use it some times to test the SqlServer adapter.

src/Phinx/Db/Adapter/MysqlAdapter.php
martijngastkemper: I'll add a more specific exception.

src/Phinx/Db/Adapter/UnsupportedColumnTypeException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Phinx\Db\Adapter;
+
+/**
+ * Exception thrown when a column type doesn't match a Phinx type.
+ *
+ * @author Martijn Gastkemper <martijn@grrr.nl>
+ */
+class UnsupportedColumnTypeException extends \Exception {
```
lorenzo: Probably extending `RuntimeException` is a good idea to keep other things working as they used to

martijngastkemper:
@lorenzo Took some commits to fix the exception messages, but that's done. Now working on a unrelated but [very strange line in the Postgresql adapter](https://github.com/cakephp/phinx/pull/1436/files#diff-7ab560398b9e11a6639f36d0856ef7f2R1005). The function always returns a string , but in one case it returns an array. The function is used by `getColumns` in the same class which expects a string. But after this change unit tests are failing. So something is going wrong. Maybe you can to elucidate this behaviour. Hopefully I can finish this PR tonight (and do some squashing...).

martijngastkemper:
The result of `getPhinxType` is used [on this line in `PostgressAdapter`](https://github.com/martijngastkemper/phinx/blob/mysql-getcolums-literal-support/src/Phinx/Db/Adapter/PostgresAdapter.php#L416). That will fail because a column type can only by a string or `Literal`.

Right now I'm going to look into this problem.

martijngastkemper:
PR #511 introduced support for PostgreSQL smallint column type. `Literal` is added for these kind of PR's. In this case it's possible to add support for smallint to all adapters. MySQL, SQLite and SqlServer support this column type aswel. Adding support for smallint to all adapters is in my opinion the best solution. Adds a new feature, doesn't break backwards compatibility and it's not very hard to accomplice.

martijngastkemper:
To finish this PR small integer support must be added to all adapters. See PR #1438 .

lorenzo:
just need to resolve the conflicts and we are good to go!

martijngastkemper:
I'll fix it tonight!

martijngastkemper:
@lorenzo I've resolved the conflicts and squashed the ugly commits. The codecov check didn't pass but when looking at the details nothings shows up. Can you fix that or tell me what I have to do to let the check pass?

lorenzo:
@martijngastkemper Thanks for all your work! I think it is safe to ignore codecov in this case

PR #1437, opened 2018-09-13
Make removeIndex work with one column as a string

martijngastkemper:


src/Phinx/Db/Action/DropIndex.php
```diff
@@ -54,13 +54,13 @@ public function __construct(Table $table, Index $index)
      * arguments.
      *
      * @param Table $table The table where the index is
-     * @param array $columns the indexed columns
+     * @param string|array $columns the indexed columns
      * @return DropIndex
      */
-    public static function build(Table $table, array $columns = [])
+    public static function build(Table $table, $columns = [])
```
lorenzo: Can you leave this as-is and make the casting in the calling sites instead?

src/Phinx/Db/Table.php
```diff
@@ -407,10 +407,10 @@ public function addIndex($columns, array $options = [])
     /**
      * Removes the given index from a table.
      *
-     * @param array $columns Columns
+     * @param string|array $columns Columns
      * @return \Phinx\Db\Table
      */
-    public function removeIndex(array $columns)
+    public function removeIndex($columns)
```
lorenzo: This is ok, just need to cast to array when building the action

lorenzo:
Thanks again!

PR #1438, opened 2018-09-13
Add small integer support to all Mysql, SqlServer and Sqlite

martijngastkemper:
Because `PostgresAdapter` already had crappy support for small integer it's added to the other adapters aswel. This makes it possible to refactor the crappy code in `PostgresAdapter`.


lorenzo:
Awesome, thanks!

PR #1441, opened 2018-09-26
Fixed pdo exception on reserved words columns

velser:
When column name is for example "limit" then we have a problem with pdo as it's recognised as reserved word. This fix adds to all column names back quote so it's recognised as column name

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -1171,7 +1171,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
             $def .= '(' . $column->getPrecision() . ',' . $column->getScale() . ')';
         }
         if (($values = $column->getValues()) && is_array($values)) {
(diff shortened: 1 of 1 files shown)


rquadling:
Is there not a quote column type method that should be used in case quote char is not ```'`'```?

velser:
Sorry I didn't understand your question.. You mean does sqlite has any of methods which will add quotes if needed for internal process ?

rquadling:
https://www.sqlite.org/lang_keywords.html says that standard SQL is to use double quotes (I didn't know that).

velser:
Heh, never used doubles as always used mysql.. :)) I will check if it works with double quote, also i will check if passing name with double quotes in single quotes will work. If yes(double around single) i will add trim first, if no just double quotes around

velser:
Nope, if calling  

> ->addColumn('"limit"'

 it stils doesn't work for enum's. But i found already defined quoting method which uses ` . Changing it to use " breaking some other stuff i guess as migrations do not pass.. ANyways sqlite supports ` and it's in method, we just quote that place for enums and everything is fine

PR #1443, opened 2018-09-27
Allow primary key change on non-integer columns

velser:
Problem exists when we want to change primary key of non-integer column. Match doesn't work so no primary key added. With this change if "creation" SQL is not with INTEGER then we parse all data from column, for example my_primary VARCHAR(100) NOT NULL will be changed to my_primary VARCHAR(100) PRIMARY KEY. (now it does not preg_match, would preg_match my_primary VARCHAR NOT NULL and change to INTEGER AUTOINCREMENT..)

Pseudo-reproduction:
1st migration:
$this->table('sometable', ['primary_key' => 'abc']);

2nd:
$this->table('sometable')->changePrimaryKey('dfg'); <- dfg is varchar

This is backwards compatible change if integers is passed without limitations and autoincrement is still there (probably it was intended?)

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -866,7 +866,20 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
         $tableName = $table->getName();
         $instructions->addPostStep(function ($state) use ($column) {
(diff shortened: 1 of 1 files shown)


djcvijic:
(Note: I am not a maintainer, I just follow the project)

I appreciate this! You could probably do even better by adding a single test case, based on SQLiteAdapterTest::testChangePrimaryKey, just with the columns having a type different than 'integer', it should just take 5 minutes.

lorenzo:
Could you please add a test for this?

velser:
Yes sure, just too busy right now..

djcvijic:
I've forked velser's repo and added the test, but it's not adding the changes to this pull request, I may have to open another one...

PR #1444, opened 2018-09-27
Quote column names when running CHECK command

velser:
Copy of: cakephp#1441 just made separate branch so other fixes wouldn't conflict comparing forked repo..

Quoting column name when doing check command for example if doing enum in sqlite and column name is reserved words such as "limit"


lorenzo:
Thanks!

PR #1451, opened 2018-10-13
Fix Composer warnings in Docker

martijngastkemper:
- unzip command not available
- execute as root not allowed


lorenzo:
Thanks :)

PR #1454, opened 2018-10-22
Update Dockerfile to use php:7.1 and update composer dependencies

djcvijic:
* when running phpunit in docker, it asks for php 7.1, so changed php version in dockerfile to 7.1
* after rebuilding the container, phpunit gives warnings about column counts, which appear in phpunit 7.*, so added <7.0 to phpunit version spec
* after composer update, everything works correctly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.0
+FROM php:7.1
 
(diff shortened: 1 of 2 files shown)


lorenzo:
Thank you!

PR #1456, opened 2018-10-24
Update table examples to use change method

MasterOdin:
Updates the docs to use the change method instead of the up/down paradigm within the docs as that's the recommended way to use Phinx.


lorenzo:
Thanks!

Issue #963, opened 2016-09-29
Status code in bash when migration fails 


Mykolaichenko:
Hello, i have some bad issue with phinx migrate. 
We have deploy pipeline, that including one step with migrations. In our case **migration is fails**, but **phinx reported to shell that all is ok** - status code 0. 

Please, review this: 

```
 == 20160912134015 Un6064UsersContract: migrating


  [PDOException]
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for
  the right syntax to use near '
              from users WHERE company_id IS NOT NULL and delete_date = 0
              GROUP BY com' at line 2
```

```
echo $?
0
```

Please, fix this troubles, because all who use phinx in deploy process can break own production environment. 
Thanks a lot.

airdrummingfool:
Until this is fixed, here's a possible workaround: after running the migrations, run `phinx status` and check the exit code. If it is non-zero, then there are still down or missing migrations, which means that your migrations did not complete successfully.

Note that if an exception is thrown when checking `status`, the exit code will also be `1` (see #806).

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

othercorey:
Added in https://github.com/cakephp/phinx/pull/1462

PR #1462, opened 2018-11-15
Add Exit Status 1 on Exception

m4rcelpfeifer:
As in #963 described, the bash currently gets the status code 0 if the migration fails. With this PR it will get the status code 1.

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -113,14 +113,19 @@ EOT
             $output->writeln('<comment>warning</comment> performing fake migrations');
         }
 
-        // run the migrations
-        $start = microtime(true);
-        if ($date !== null) {
-            $this->getManager()->migrateToDateTime($environment, new \DateTime($date), $fake);
-        } else {
-            $this->getManager()->migrate($environment, $version, $fake);
+        try {
+            // run the migrations
+            $start = microtime(true);
+            if ($date !== null) {
+                $this->getManager()->migrateToDateTime($environment, new \DateTime($date), $fake);
+            } else {
+                $this->getManager()->migrate($environment, $version, $fake);
+            }
+            $end = microtime(true);
+        } catch (\Throwable $e) {
+            $output->writeln('<error>' . $e->getMessage() . '</error>');
+            return 1;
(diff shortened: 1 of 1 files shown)


src/Phinx/Console/Command/Migrate.php
```diff
@@ -113,14 +113,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        try {
+            // run the migrations
+            $start = microtime(true);
+            if ($date !== null) {
+                $this->getManager()->migrateToDateTime($environment, new \DateTime($date), $fake);
+            } else {
+                $this->getManager()->migrate($environment, $version, $fake);
+            }
+            $end = microtime(true);
+        } catch (\Throwable $e) {
```
piotr-cz: As Phinx is still compatible with [PHP v5.6](https://github.com/cakephp/phinx/blob/v0.10.6/composer.json#L28) which doesn't have a Throwable class, there should be catch for an Exception:

```suggestion
        } catch (\Exception $e) {
            $output->writeln('<error>' . $e->getMessage() . '</error>');
            return 1;
        } catch (\Throwable $e) {
```

m4rcelpfeifer: You are right, thanks.

src/Phinx/Console/Command/Migrate.php
```diff
@@ -113,14 +113,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            // run the migrations
+            $start = microtime(true);
+            if ($date !== null) {
+                $this->getManager()->migrateToDateTime($environment, new \DateTime($date), $fake);
+            } else {
+                $this->getManager()->migrate($environment, $version, $fake);
+            }
+            $end = microtime(true);
+        } catch (\Exception $e) {
+            $output->writeln('<error>' . $e->getMessage() . '</error>');
```
lorenzo: Not having the stack trace would be definitely annoying

lorenzo: Can you show a screenshot of how it looks like for you in the CLI?

lorenzo:
Thanks!

PR #1463, opened 2018-11-15
Enable PHP7.3

dereuromark:


dereuromark:
interesting fail

dereuromark:
Any idea of a fix here?

garas:
PHP 7.3 is linked with newer Sqlite.

Sqlite doesn't support column comments Phinx emulates comments by appending `/* comment */` to column definition.

New Sqlite probably now handles `/* */` differently.

I suggest to skip adding emulated comment in Sqlite. 

Column changing already refuses that, it should be the same when adding column.
https://github.com/cakephp/phinx/blob/76c96c851b737f93598d7ae7765a5cd2cc99ae7a/src/Phinx/Db/Adapter/SQLiteAdapter.php#L286-L289

dereuromark:
Help is greatly appreciated.

dereuromark:
2nd part sounds good.
@lorenzo what do you think?

lorenzo:
hmm, how should be getting the comments in sqlite now?

JKingweb:
@dereuromark Can you run the checks again? The test should pass, now.

dereuromark:
The comment issue is still there:
```
-''another default''
+''another default' /* another comment */'
```

MasterOdin:
You probably need to merge master first.

lorenzo:
👏

PR #1466, opened 2018-11-21
Fix decimal info getColumns()

### LINKED ISSUE #1013: MysqlAdapter: getColumns() doesn't return precision and scale for decimal

martijngastkemper:
Fixes #1013 and replaces PR #1258

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -401,7 +401,8 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
                    ->setDefault($columnInfo['Default'])
                    ->setType($phinxType['name'])
                    ->setSigned(strpos($columnInfo['Type'], 'unsigned') === false)
-                   ->setLimit($phinxType['limit']);
+                   ->setLimit($phinxType['limit'])
+                   ->setScale($phinxType['scale']);
 
             if ($columnInfo['Extra'] === 'auto_increment') {
                 $column->setIdentity(true);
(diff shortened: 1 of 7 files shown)


lorenzo:
Thank you, and sorry for noticing this too late

PR #1473, opened 2018-12-19
Docs about seed: Call to save() method replaced for saveData()

georgezim85:
Avoid error: Base table or view already exists: 1050 Table <table_name> already exists

### DIFF
--- a/docs/seeding.rst
+++ b/docs/seeding.rst
@@ -132,13 +132,13 @@ within your seed class and then use the `insert()` method to insert data:
 
                 $posts = $this->table('posts');
                 $posts->insert($data)
-                      ->save();
+                      ->saveData();
             }
         }
 
 .. note::
 
-    You must call the `save()` method to commit your data to the table. Phinx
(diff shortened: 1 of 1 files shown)


lorenzo:
hmm, this should not be the case. Can you explain how this can be reproduced?

dereuromark:
Any update here? ping @georgezim85

georgezim85:
> hmm, this should not be the case. Can you explain how this can be reproduced?

Sure. If you have a table already created and so you want to populate using the seeder you are creating, you cant't do it using the "save()" method, because it will try to create de table again, so we must use saveData() to just insert the data.

georgezim85:
> > hmm, this should not be the case. Can you explain how this can be reproduced?
> 
> Sure. If you have a table already created and so you want to populate using the seeder you are creating, you cant't do it using the "save()" method, because it will try to create de table again, so we must use saveData() to just insert the data.

I am referring to MySql databases.

lorenzo:
I see, this looks good to me now. Thanks

PR #1478, opened 2019-01-02
Update LICENSE

cleptric:
Shall we also add `Cake Software Foundation, Inc. (https://cakefoundation.org)` ?


dereuromark:
I dont think we can easily change it.
We could add it at some point however.

PR #1479, opened 2019-01-03
Simplify logic around default primary key name when creating table

MasterOdin:
Makes it so that regardless if you're setting `id` or not, it'll use the same if block of code, instead of previously having two identical blocks except one used the string 'id' and the other `$options['id']`


dereuromark:
@lorenzo Good to go?

lorenzo:
Thanks!

PR #1485, opened 2019-01-22
Close #1481 verify if set onlye scale

### LINKED ISSUE #1481: Precision and Scale bug mysql
When I create table with column decimal, for example:

```php
->addColumn(
                'id_usuario_inclusao', 
                'decimal', [
                    'default' => null,
                    'null' => false,
                    'precision' => 30,
                    'scale' => 0
                ])
```

the method protected function getColumnSqlDefinition() in  Phinx\Db\Adapter\MysqlAdapter, no set precision and scale because the verification `if ($column->getPrecision() && $column->getScale()) {` not works.

If I set scale how true, 1 and more it works, for example:
```php
->addColumn(
                'id_usuario_inclusao', 
                'decimal', [
                    'default' => null,
                    'null' => false,
                    'precision' => 30,
                    'scale' => true
                ])
```
My suggest is change the verification `if` to:

```php
        } elseif ($column->getScale()) {
            $def .= '(' . 10 . ',' . $column->getScale() . ')';
        } elseif ($column->getPrecision()) {
            $def .= '(' . $column->getPrecision() . ',' . 0 . ')';
        }
```

HDias:
Verification if set only Scale to MysqlAdapter

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1117,8 +1117,15 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $sqlType = $this->getSqlType($column->getType(), $column->getLimit());
             $def = strtoupper($sqlType['name']);
         }
+        
         if ($column->getPrecision() && $column->getScale()) {
-            $def .= '(' . $column->getPrecision() . ',' . $column->getScale() . ')';
+            $def .= sprintf('(%d, %d)', $column->getPrecision(), $column->getScale());
+        } elseif ($column->getScale()) {
+            $message = sprintf(
+                "'scale' option must be used together with 'precision' option on column '%s'",
+                $column->getName()
+                );
+            throw new \InvalidArgumentException($message);
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';
         }


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1117,8 +1117,15 @@ protected function getColumnSqlDefinition(Column $column)
             $sqlType = $this->getSqlType($column->getType(), $column->getLimit());
             $def = strtoupper($sqlType['name']);
         }
+        
         if ($column->getPrecision() && $column->getScale()) {
```
garas:
```suggestion
        if ($column->getPrecision()) {
```

It should be only precision check here, so it works with scale set to any number, 0, or not set.

I gave correct condition in first comment, but I made this mistake in second comment https://github.com/cakephp/phinx/issues/1481#issuecomment-454351168

HDias:
I found de the problem, will send new pull request

garas:
Add commits to the same branch `patch-2` of your fork and they will appear here. You don't need open new PR.

dereuromark:
Any chance we can have a small test case?

dereuromark:
ping @HDias

AgelxNash:
```

dereuromark:
ping

othercorey:
Closing due to no activity.

PR #1486, opened 2019-01-22
Fixed #1481 Bug when getScale

### LINKED ISSUE #1481: Precision and Scale bug mysql

HDias:
When add COlumn:
```php
'decimal', [
                    'precision' => 30,
                    'scale' => 0,
                ]
``` 
in line 1120 no set DECIMAL(30, 0) because the getScale() == '0'

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1117,7 +1117,7 @@ class MysqlAdapter extends PdoAdapter implements AdapterInterface
             $sqlType = $this->getSqlType($column->getType(), $column->getLimit());
             $def = strtoupper($sqlType['name']);
         }
-        if ($column->getPrecision() && $column->getScale()) {
+        if ($column->getPrecision() && $column->getScale() >= 0) {
             $def .= '(' . $column->getPrecision() . ',' . $column->getScale() . ')';
         } elseif (isset($sqlType['limit'])) {
             $def .= '(' . $sqlType['limit'] . ')';


dereuromark:
A small test case would be helpful here.

MasterOdin:
Would it make sense to handle the case of just setting precision (which MySQL allows) by doing something like:
```
        if ($column->getPrecision() && $column->getScale() >= 0) {
        }
        elseif ($column->getPrecision()) {
            $def .= '(' . $column->getPrecision() . ')';
        } 
```

garas:
Yes, `if ($column->getPrecision())` handles case when scale=0, as `DECIMAL(10, 0)` is same as `DECIMAL(10)`.

Check `$column->getScale() >= 0` only prevents from negative values, and I think it is too deep validation, otherwise it should be checked `P > 0`, `P > S` etc.

I suggest you to close this PR and continue on https://github.com/cakephp/phinx/pull/1485 as it is closer to solution. Accept my _Suggested change_, also fix whitespace issue

dereuromark:
Closing then in favor of https://github.com/cakephp/phinx/pull/1485

Issue #1490, opened 2019-01-28
Add Composer dependencies for required PHP extensions


halfer:
I just ran `phinx migrate` in an Alpine/PHP environment, and found that it blew up because the `json` extension was not installed. Can a PHP extension dependency be added to `composer.json`? This will ensure Composer fails gracefully based on an unavailable dependency.

I believe this is done using `ext-json` in the `require` section.

If a core dev is :+1: then ping me and I will raise a PR.

lorenzo:
I'm ok with this

halfer:
Great. Are there any other exts that are needed? I can add them at the same time.

dereuromark:
Only add whats required. Your IDE (e.g. phpstorm) will usually tell you in the files.

halfer:
@dereuromark: I'm on NetBeans, it's not quite that fancy :grin:. No matter, it was only if someone knew some extensions off the top of their head. I'll get `json` done for now.

dereuromark:
PR open

PR #1491, opened 2019-01-28
Tell Composer the JSON extension is required

halfer:
As per #1490, this asks Composer to check whether the JSON extension is used. It seems to be used in the Phinx core to read its own `composer.json` file.

Not all PHP installations - particularly Alpine - can guarantee that this extension is loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,3 @@ phpunit.xml
 
 # sqlite test database
 phinx_testing.sqlite3
-/nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ phpunit.xml
 
 # sqlite test database
 phinx_testing.sqlite3
+/nbproject/private/
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,3 +1,8 @@
+build:
+  environment:
+    pecl_extensions:
+      - json
+
 checks:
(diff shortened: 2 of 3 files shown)


halfer:
Ah, initially I couldn't imagine that my new dependency would have stopped Scrutinizer from working. However, this is probably the case [because of this](https://scrutinizer-ci.com/docs/tools/php/php-analyzer/guides/composer_dependencies).

I wonder if there is a way to create a lock file in a test folder, to be copied into place just for Scrutinizer?

dereuromark:
As soon as you do composer update or install it will automatically have a lock file for CI or scrut.

halfer:
Do you mean the build failure in this PR can be ignored? :smiley:

dereuromark:
It might just need a different setup process. Not sure how we can achieve it, though.

halfer:
I'll try some Scrutinizer config to add the required extension into PHP.

If that does not work, I will try generating a lock file, but rename it to something safe, and only move it into place in Scrutinizer.

dereuromark:
Shall we proceed with this? @lorenzo

MasterOdin:
@halfer @dereuromark this PR can be closed after #1653 was merged as the composer.json file is no longer read.

halfer:
Thanks @MasterOdin, good work :+1:

PR #1492, opened 2019-01-28
Just testing if a Composer lock allows the JSON ext to work (not for merging)

halfer:
This is a test for PR 1491 and can be ignored. I am experimenting to see if adding the `composer.lock` file allows the build to go through. If it does, Scrutinizer can probably be told to copy it in place from a spare/test directory, rather than committing it where it is.


halfer:
I'm trying a few things here to bring in a lock file from another location, all very much to no avail. If anyone chances upon this, and is familiar with how to run "before" commands in the `checks` phase only, please shout up.

dereuromark:
Closing in favor of https://github.com/cakephp/phinx/pull/1491 - we hopefully can merge that now soon.

PR #1497, opened 2019-02-05
Add upgrade guide for 0.10

julesjcraske:
This is not a complete upgrade guide, this is only the things I have had to change when upgrading. 

These may not be things you wanted to change in which case I could raise issues.


lorenzo:
Sentences should not be left incomplete, that would make it confusing to others

PR #1501, opened 2019-02-15
Update documentation on seeding

chillichicken:
It wasn't obvious that dependencies won't be considered when running specific seed classes (-s)

### DIFF
--- a/docs/seeding.rst
+++ b/docs/seeding.rst
@@ -99,6 +99,11 @@ current seeder:
                 // `ShopItemSeeder` have been run.
             }
         }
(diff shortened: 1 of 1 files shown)


docs/seeding.rst
```diff
@@ -99,6 +99,11 @@ current seeder:
                 // `ShopItemSeeder` have been run.
             }
         }
+        
+.. note::
+
+    Dependencies are only considered when executing all seed classes (default behaviour).
```
dereuromark: behavior

PR #1505, opened 2019-02-28
docs: add MySQL `bit` type

caugner:
#1248 added support for the MySQL column type `bit`,
but the type was not yet mentioned in the docs.

In addition, this PR aligns the currently inconsistent sections *Working With Tables* and *Working With Columns* w.r.t. the columns supported by the MySQL and Postgres adapter.


caugner:
@lorenzo Any chance to get these small docs improvements merged? :)

PR #1510, opened 2019-03-20
Mongo adapter

garipovazamat:
Add adapter for mongo with suggestion to install mongodb/mongodb.


dereuromark:
This seems to be missing a test.

PR #1517, opened 2019-04-07
Let column `updated_at` auto use current time fill

dream-mo:
I now have a requirement that I want the `updated_at` column to be automatically populated with the current time when executing the update SQL statement, instead of populating this field every time I specify the update time in the code, which will make the code cleaner. The `updated_at` and `created_at` fields are managed by the mysql database. In order to minimize source code changes, I just added a method to achieve the above functions. Use this method to ensure that the user explicitly uses the mysql database and that the version should be greater than or equal to 5.6. The source code did not use the attribute "URRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP" for `updated_at` before,i guess it should be officially considered compatibility issues and not processed.

For example:

Add attributes to the updated_at column after mysql5.6:

`updated_at` timestamp NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP

When you update a record, updated_at will automatically fill in the update time.

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -540,19 +540,62 @@ class Table
      */
     public function addTimestamps($createdAt = 'created_at', $updatedAt = 'updated_at', $withTimezone = false)
     {
+        return $this->innerAddTimestamps($createdAt, $updatedAt, $withTimezone);
+    }
+
+    /**
+     * Add timestamp columns created_at and updated_at to the table.By default, `update_at` adds attributes 'CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'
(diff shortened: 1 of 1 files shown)


rquadling:
Can you get the fractional seconds in here? I don't use the DBAL aspect of Phinx and have to go with

      createdAt   TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
      updatedAt   TIMESTAMP(6) NULL     DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP(6),

dereuromark:
ping @dream-mo

othercorey:
Closing due to no activity.

Issue #1520, opened 2019-04-19
Phinx is slowing down every project that has it as a requirement


alexskra:
**I thought Phinx as a dependency would be fine**
So my thought was that adding Phinx as a dependency would be a non-issue. I'd only use it once for migration on deployments and since autoloading is a thing it wont ruin the service by tanking performance. I think that's a fair assumption.

**But dependencies can be evil**
But now in v0.10 Phinx has added cakephp dependencies that does something very stupid: The dependencies are adding files as requirements.

**Why is that bad?**
The code in the required files are executed on every single request. It's absolutely pointless and can have a significant effect on the performance of lightweight applications.(reduces performance to ~82 % in my situation)

https://getcomposer.org/doc/04-schema.md#files

> If you want to require certain files explicitly on every request then you can use the files autoloading mechanism.

Here's the thing. Phinx is usually used from the CLI, and rarely from inside code itself. But these dependencies results in phinx dependency code getting executed on every single page load even if it has nothing to do with phinx. That is bad and can cause performance issues.

**Possible solutions**
Some possible solutions is to remove dependencies that doesn't respect performance, find something better to depend on or rewrite Phinx so it doesn't need dependencies that affects performance.

Dependencies that are causing problems:

- cakephp/collection
- cakephp/database (requires cakephp/core that causes the problem)

broberts-mrmc:
This is not an experience I’ve had. Does your app’s bootstrapping process instantiate Phinx?

aleskr:
Well, that is how it works. All you have to do is to include the composer autoloader, you don't have to do anything with sphinx, just adding it as a dependency is enough, and that's why it's a problem.

josegonzalez:
How are you gauging the performance change? Do you have a reproduction here we can run?

aleskr:
You can measure it however you want, either by benchmarking the whole stack or just profiling a run.

1. Create a PHP file and require the autoloader generated by composer.
```php
<?php
require 'vendor/autoload.php';
```
2. Measure/profile or whatever.
3. Install Phinx.
4. Congratulations, you now have additional code that is executed on every request!

broberts-mrmc:
[...]

PR #1522, opened 2019-04-20
Fix box configuration to work with humbug/box

jawngee:
With regards to #1520:

Considering humbug/box (https://github.com/humbug/box) is now the actively maintained app for building phars, I updated the box.json configuration so that working phars can be generated.  

The existing box.json was not working at all as humbug/box excludes composer.json from the root unless you explicitly set a flag to tell it not to.

### DIFF
--- a/box.json
+++ b/box.json
@@ -8,8 +8,7 @@
     "directories": ["src","app"],
     "files": [
         "LICENSE",
-        "phinx.yml",
-        "composer.json"
+        "phinx.yml"
     ],
     "finder": [
         {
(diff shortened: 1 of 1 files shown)


lorenzo:
Thanks!

lorenzo:
What do you suggest our next step should be?

theofidry:
Hello, just saw this issue after a report from a user:

- There is some instructions to install phinx somewhere still referencing Box2
- It is not clear where/how box is downloaded which makes confusing which Box version this configuration is used with
- If you are running the latests versions of Box3, you should have a couple of recommendations regarding redundant configuration settings and you could probably trim it down even further by taking advantage of some new Box3 features, namely that most files can be inferred from the Composer configuration

Issue #1468, opened 2018-11-28
Rollback problems with index and foreign key adding in same change method


egorbwork:
When I tried to upgrade phinx to version 0.10.0. Following case is not any more possible to rollback:
```
public function change()
{
        $table = $this->table('my_table');
        $table->addForeignKey('entity_id', 'my_another_table', 'id', [
            'delete' => ForeignKey::CASCADE,
            'update' => ForeignKey::CASCADE,
            'constraint' => 'my_another_table_foreign_key'])
            ->addIndex(['entity_id'], ['unique' => true])
            ->update();
}
```

The response from console is:
` SQLSTATE[HY000]: General error: 1553 Cannot drop index 'entity_id': needed in a foreign key constraint  `

Is there any solution for this?

ArtemGoutsoul:
the order of dropping column, foreign key and index in rollback is wrong, it's a clear regression

lorenzo:
would you mind showing the generated SQL that triggers the problem?

egorbwork:
What I get on -vv: 
```
START TRANSACTION
 -- Altering table my_table
ALTER TABLE `my_table` DROP INDEX `entity_id`
```

And after:
`SQLSTATE[HY000]: General error: 1553 Cannot drop index 'entity_id': needed in a foreign key constraint `

ArtemGoutsoul:
Any workarounds for now?

ArtemGoutsoul:
> 2 migrations: one for the index and other for the foreign key.

This would require refactoring all past migrations, which is not viable...

PR #1523, opened 2019-04-24
Fix reversible migrations when dropping FK and index

### LINKED ISSUE #1426: Foreign key drop / add (with the same name) in a single run results in mysql error
With the introduction of the "Plan" dropping / adding a foreign key with the same name (e.g. in order to change ondelete behavior) results in an error in mysql - apparently it cannot handle it when it's compiled to a single query.
```sql
ALTER TABLE `module_challenges_taskslots` DROP FOREIGN KEY fk_module_challenges_taskslots_module_challenges_missions1, ADD  CONSTRAINT `fk_module_challenges_taskslots_module_challenges_missions1` FOREIGN KEY (`id_mission`) REFERENCES `module_challenges_missions` (`id`) ON DELETE CASCADE;
```
`[23000][1022] Can't write; duplicate key in table '#sql-1_3'`
In phinx <0.10.x this was split to two separate queries, afair.

Should AddForeignKey and DropForeignKey be grouped separately in \Phinx\Db\Plan\Plan::gatherConstraints (with drop being executed first of the two), perhaps?

an example failing (working before the upgrade to 0.10.x) migration:
```php
<?php

class ChallengeTaskSlotAdjustForeignKey extends AbstractMigration
{
	public function up()
	{
		$this->table('module_challenges_taskslots')
			// modify onDelete from SET NULL to CASCADE
			->dropForeignKey('id_mission')
			->addForeignKey('id_mission', 'module_challenges_missions', 'id', ['delete' => 'CASCADE', 'constraint' => 'fk_module_challenges_taskslots_module_challenges_missions1'])
			->update();
	}

	public function down()
	{
		$this->table('module_challenges_taskslots')
			// modify onDelete from CASCADE to SET NULL
			->dropForeignKey('id_mission')
			->addForeignKey('id_mission', 'module_challenges_missions', 'id', ['delete' => 'SET NULL', 'constraint' => 'fk_module_challenges_taskslots_module_challenges_missions1'])
			->update();
	}
}

```

### LINKED ISSUE #1514: DropForeignKey and RemoveColumn in the same migration can conflict
I recently started the process of upgrading from v0.9.2 to v0.10.6, and discovered a weird interaction (in MySQL): Phinx now generates two queries to drop the FK and the corresponding column, but tries to drop the column before the FK. This results in `SQLSTATE[HY000]: General error: 1553 Cannot drop index 'page_id': needed in a foreign key constraint`

It's pretty easy to work around by tweaking the migration, but it would be nice if the two queries ran in the correct order by default so that  wouldn't be necessary.

Migration (this worked in 0.9.2 since it was a bunch of queries):

```php
    public function up()
    {
        $this->table('pages')
            ->dropForeignKey('state')
            ->removeColumn('state')
            ->update();
    }
```

Schema pre-migration (irrelevant columns removed for brevity):

```sql
CREATE TABLE `pages` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `state` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
  PRIMARY KEY (`id`),
  KEY `state` (`state`),
  CONSTRAINT `pages_ibfk_2` FOREIGN KEY (`state`) REFERENCES `scrape_states` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

Dry-run SQL output:

```sql
START TRANSACTION
ALTER TABLE `pages` DROP COLUMN `state`
ALTER TABLE `pages` DROP FOREIGN KEY pages_ibfk_2
COMMIT
INSERT INTO `phinxlog` (`version`, `migration_name`, `start_time`, `end_time`, `breakpoint`) VALUES ('20180315164720', 'RemoveOldColumnsFromPage', '2019-03-28 19:13:57', '2019-03-28 19:13:57', 0);
```

lorenzo:
closes #1514 and #1468

Also fixes a problem when dropping and adding the the same foreign key in the same migration file, which closes #1426


src/Phinx/Db/Plan/Plan.php
```diff
@@ -206,6 +216,90 @@ protected function forgetTable(Table $table, $actions)
...
+                $newAlter->addAction($action);
+            });
+
+        return $newAlter;
+    }
+
+    /**
+     * Deletes any DropIndex actions for the given table and exact columns
+     *
+     * @param Table $table The table to find in the list of actions
```
dereuromark: Your IDE could be configured to auto fqcn here in doc blocks. Not sure if our sniffer for 3.x can fix this yet on his own.

lorenzo: I thin in phinx in general we are not using fqcn

dereuromark: Probably just a matter of proper coding standards being applied via auto-fixer.
Can also be done laters.

src/Phinx/Db/Plan/Plan.php
```diff
@@ -171,18 +172,27 @@ public function executeInverse(AdapterInterface $executor)
...
                 $this->constraints = $this->forgetTable($action->getTable(), $this->constraints);
                 $this->indexes = $this->forgetTable($action->getTable(), $this->indexes);
             }
         }
+
+        // Dropping indexes used by foreign keys is a conflict, but one we can resolve
+        // if the foreign key is also scheduled to be dropped. If we can find sucha a case,
+        // we force the execution of the index drop after the foreign key is dropped.
+        $this->constraints = collection($this->constraints)
+             ->map(function ($alter) {
```
garas:
```suggestion
             ->map(function (AlterTable $alter) {
```

lorenzo: I’m on the road again, do you mind making the suggestions here and I will merge them using github?

garas:
I've tried with more complex example with BelongsToMany tables, but it fails.

FK on `articles.featured_photo_id` should be dropped before dropping table `photos`.

```php
    public function change()
    {
        $this->table('photos')
            ->addColumn('name', 'string')
            ->create();

        $this->table('articles_photos')
            ->addColumn('article_id', 'integer')
            ->addIndex(['article_id'])
            ->addForeignKey('article_id', 'articles', 'id')
            ->addColumn('photo_id', 'integer')
            ->addIndex(['photo_id'])
            ->addForeignKey('photo_id', 'photos', 'id')
            ->create();

        $this->table('articles')
            ->addColumn('featured_photo_id', 'integer')
            ->addIndex(['featured_photo_id'])
            ->addForeignKey('featured_photo_id', 'photos', 'id')
            ->update();
    }
```

lorenzo:
@garas that would be a separate issue, this one is about conflicting drop index. Can you open another ticket?

lorenzo:
@garas also, please add in which way it fails and the dry run sql if possible.

garas:
I've actually started working on this issue too. Test cases are my weak area, so it didn't finished with PR.

My example fails when rollbacking.

My fix for my sample that worked was simply to move FK to the end of sequence.

```php
    protected function updatesSequence()
    {
        return [
            $this->tableUpdates,
            $this->indexes,
            $this->tableMoves,
            $this->constraints,
        ];
    }
```

Without tests however I'm not sure if it affects other cases.

lorenzo:
@garas submit a pull request with your change and let Travis run the tests for you :)

lorenzo:
@garas A problem I imagine could happen with your change is renaming tables. With that sequence of changes, you will first rename and then change the constraints, which may contain the wrong table names.

falkenhawk:
@lorenzo any chance it fixes also this issue? https://github.com/cakephp/phinx/issues/1426

lorenzo:
@falkenhawk no, this only deals with conflicting index drops and fk drops, but I think I have time to look into that issue today or tomorrow

lorenzo:
@falkenhawk added a fix for your issue here too

lorenzo:
Seems like it is good to merge now :)

Issue #596, opened 2015-07-09
changeColumn touchs other options 


salarmehr:
Consider if want to change length of a nullable column. using `['length' => MysqlAdapter::INT_TINY]` make the column `not null`. I have to specify `'null'=>true]` options to prevent that. I think Phinx should not change unspecified options.

Decadence:
Similar situation with comment. If it is not specified, comment will be deleted.

rquadling:
A reasonable solution would be to get the column description from the DB and populate the change column params with that, and then override them with the user supplied ones.

Not all columns provide/require the same options (http://docs.phinx.org/en/latest/migrations.html#valid-column-options), so each adaptor / column combo would need to be tuned accordingly.

As changColumn is not reversible, the up and down would/should operate in the same way, only changing the attributes that need changing, rather than supplying those that are not changed, but need to be included.

As changing a column is handled by `\Phinx\Db\Table::changeColumn()`, I wonder how much work to interact with `\Phinx\Db\Adapter\AdapterInterface::getColumns()` to get the current setup, and then override it with the user supplied one.

Care to have a go at creating a PR for this?

I'd be happy to look it over/help you through it?

rquadling:
The adapter level `getColumns()` do not SEEMS to work consistently with regard to comments (first glance at least), so, if it is an issue, would need fixing also.

styks1987:
This is not optimized but this is what I am doing at the moment since if changeColumn is given a Column Class it will not create a new Column instance.

    $table->changeColumn('user_id', $this->getColumn('user_id', ['after' => 'id']));

    private function getColumn($columnName, $options)
    {
        $columns = new \Cake\Collection\Collection($this->table('Orders')->getColumns());

        $column = $columns->filter(function ($column) use ($columnName){
            return $column->getName() == $columnName;
        })->first();

        $column->setOptions($options);
        return $column;
    }

dereuromark:
Are you able to provide a patch with your suggested changes as PR?

dereuromark:
Check out https://github.com/cakephp/migrations/pull/941 - if thats sth to be ported to phinx directly,
I think the separate method allows BC here for the attributes.

Issue #123, opened 2013-08-24
changeColumn() allow null


joshribakoff:
``` php
$this->changeColumn('website', null, ['default'=>'null'])
```

This should keep the column type as is but just makes it null-able.  Right now you have to specify the current column type for 2nd parameter even if its not changing. I should make it so that passing empty string or null keeps it as is.

Its also finicky about 'default' being set to the string 'null' and not the actual value null.

moskrc:
+1

Isinlor:
+1

AgentStephens:
+1

robmorgan:
I like this idea

igorsantos07:
Man, this is the dream of every database user: not having to remember the column type if you want to just change an option. hahaha

JREAM:
+1

simkimsia:
+1

marcosdalte:
:+1:

vitorccsiqueira:
+1

TeyimPila:
+1

LauLogisch:
+1

dereuromark:
Anyone wants to make a PR here?

dereuromark:
> Its also finicky about 'default' being set to the string 'null' and not the actual value null.

That is expected.

But other fields not specific should remain the same, I agree on that.

dereuromark:
Check out https://github.com/cakephp/migrations/pull/941 - if thats sth to be backported to phinx directly,
I think the separate method allows BC here for the attributes.
The type "null" change is a no brainer either way :)

PR #1524, opened 2019-04-26
Fixed Bug #596, #123 - changeColumn resets the nullable option if not specified

saitejaprasadam:
Fetching column options before performing changeColumn Action

### DIFF
--- a/src/Phinx/Db/Action/ChangeColumn.php
+++ b/src/Phinx/Db/Action/ChangeColumn.php
@@ -72,7 +72,7 @@ class ChangeColumn extends Action
      * @param mixed $options Additional options for the column
      * @return ChangeColumn
      */
-    public static function build(Table $table, $columnName, $type = null, $options = [], $adapter)
+    public static function build(Table $table, $columnName, $type = null, $options = [], $adapter = null)
     {
         $column = new Column();
         $column->setName($columnName);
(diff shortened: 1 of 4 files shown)


src/Phinx/Db/Action/ChangeColumn.php
```diff
@@ -72,11 +72,12 @@ public function __construct(Table $table, $columnName, Column $column)
      * @param mixed $options Additional options for the column
      * @return ChangeColumn
      */
-    public static function build(Table $table, $columnName, $type = null, $options = [])
+    public static function build(Table $table, $columnName, $type = null, $options = [], $adapter = null)
```
lorenzo: This is not ok, options should be passed to the object and not fetched here.

saitejaprasadam: when a column change is performed, if the null option not specified it will replace the existing value with the default value from Phinx\Db\Action\Column.php class.

Currently, am fetching existing options of the column from Database in Phinx\Db\Action\Column.php class (isn't it right??). Can I fetch the existing options of the column from Database in Phinx/Db/Action/ChangeColumn.php

lorenzo: No, the part of fetching data inside this class is the part I'm objecting to. You should fetch it from the outside and pass the data as an argument

saitejaprasadam: Got it @lorenzo, For confirmation, can I fetch the existing options of the column from Database in Phinx/Db/Action/ChangeColumn.php?

lorenzo: No, you should not fetch the from that class, they should be passed as arguments.

dereuromark:
Whats the bugfix status here?

PR #1493, opened 2019-01-30
Add double type

adamwulf:
This continues the PR from #1208. I've resolved the conflicts for merging with master.

This feature request has come up numerous times, from the following tickets: #498, #1025, #722, #668, and likely elsewhere.

This branch picks up from where @ntzm left off, it simply merges his existing branch into the current master.

PR #1527, opened 2019-05-06
Add double to valid column types

nlubisch:
As introduced with #1493 und released with 0.10.7 using `double` is possible.
This PR adds it to the valid column types list.


lorenzo:
thank you!

PR #1529, opened 2019-05-07
[SqlServer] Fix SqlServer syntax after making precision alias of limit

### LINKED ISSUE #1528: SQL Server syntax broken after making precision alias of limit
SQL Server adapter broken after #1466 

`SqlServerAdapter::getColumnSqlDefinition()` returns limit inline with precision and scale when adding decimal type columns.

misotolar:
Fixes #1528


lorenzo:
Thanks!

PR #1533, opened 2019-05-15
Various SQLite fixes:

JKingweb:
- Use suffix option when dropping a database
- Handle dropping a memory database
- Allow the use of the suffix and memory options in tests
- Fix trivially failing tests in Windows
- Make memory option a proper boolean check
- Slight coverage improvement by simplifying trivial string-to-array conversions

Tests still fail in Windows when using a file-backed database (you can't unlink an open database in Windows), but I have more important problems I want to tackle first. ;)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -73,7 +73,7 @@ public function connect()
             $options = $this->getOptions();
 
             // if port is specified use it, otherwise use the MySQL default
-            if (isset($options['memory'])) {
+            if (isset($options['memory']) && $options['memory']) {
```
dereuromark: Why that? If so, you could use !empty() as one call.

JKingweb: Sure. `empty()` isn't in my mental toolset, so I didn't think of it. I'll make the change.

dereuromark: I think I just had a similar issue with getenv() being false instead of null ( https://github.com/cakephp/cakephp/pull/13242 ), running into similar issues if you use too strict check on the null type.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -641,11 +641,7 @@ protected function getIndexes($tableName)
      */
     public function hasIndex($tableName, $columns)
     {
-        if (is_string($columns)) {
-            $columns = [$columns]; // str to array
-        }
-
-        $columns = array_map('strtolower', $columns);
+        $columns = array_map('strtolower', (array) $columns);
```
dereuromark: no space after casting, see CS.

JKingweb: Will fix. Is this documented anywhere?

lorenzo:
Thanks, this is really good!

JKingweb:
> Thanks, this is really good!

Thanks. More to come, if you're interested.

lorenzo:
100% interested

Issue #1535, opened 2019-05-16
SQLite: hasTable() fails when specifying a schema


JKingweb:
The `SQLiteAdapter::hasTable()` method fails when provided with a schema in the table name, including the `main` and `temp` schemata.

PR #1537, opened 2019-05-16
Fix for issue #1535

JKingweb:
Enhances `SQLiteAdapter::hasTable()` so that it can find tables in any already-attached database, including temporary tables when explicitly asked for. 

Also fixes a quoting issue the original implementation had: table names containing `'` would previously cause the method to error out. This required changing the `PdoAdapter::quoteValue()` method visibility to `protected`.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -266,6 +266,17 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
     }
 
     /**
+     * Quotes a database string.
+     *
+     * @param string $value  The string to quote
+     * @return mixed
+     */
+    protected function quoteString($value)
+    {
+        return $this->getConnection()->quote($value);
+    }
+
+    /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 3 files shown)


JKingweb:
How do I find out what Stickler is actually complaining about? I've tried running phpcs locally and it only complains about something I didn't write.

lorenzo:
@markstory do you know what is going on with sticklerci here?

markstory:
@lorenzo It is attempting to apply phpcbf changes but can't because the changes are coming from a fork. I will get it updated so that fixer failure doesn't stop other errors from being reported.

lorenzo:
I'll merge this now and solve the issues later. Thanks for the contribution!

lorenzo:
Oops, after a second look at the tests, it is clear that there are tests failing. It is not just style issues. Reverting!

lorenzo:
Can you please send a new pull request with passing tests?

JKingweb:
They had passed on my end in the first place, so it may take a bit of fiddling. I'd imagine it's a cross-version compatibility thing. I'll have a look, though.

lorenzo:
thanks!

PR #1542, opened 2019-05-28
New hasTable implementation for SQLite

### LINKED ISSUE #1535: SQLite: hasTable() fails when specifying a schema

JKingweb:
A second attempt at fixing `SQLiteAdapter::hasTable()`.

Notably, it is able to find tables in any schema rather than only main, and is also able to find tables when the main schema is explicitly specified.

Like [SQLite itself](https://sqlite.org/lang_naming.html), if no schema is specified it will search all schemata until the table is found. This is unlike the other adapters, but is internally consistent.

This fixes #1535.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -266,6 +266,17 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
     }
 
     /**
+     * Quotes a database string.
+     *
+     * @param string $value  The string to quote
+     * @return string
+     */
+    protected function quoteString($value)
+    {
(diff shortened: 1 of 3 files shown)


lorenzo:
Thanks!

JKingweb:
Would you prefer a series of small fixes like this, or one big change-set? It turns out I'll probably end up doing a pretty significant rewrite of the SQLiteAdapter class at this point. I can prepare pull requests as I fix things, or just one at the end. Either way I'll have lots of tests (340 and counting (just for SQLite) so far) so that you can be confident it's correct.

lorenzo:
Smaller is better, makes it easier to review

PR #1545, opened 2019-05-30
Fixes to SQLite index handling

### LINKED ISSUE #1544: SQLite: Index handling is completely broken
Phinx is currently absolutely hopeless when finding, listing, re-creating, or dropping indices for an SQLite database. The root cause is [a particular line](https://github.com/cakephp/phinx/blob/2bdeffc5b828baf79fa4d2761aa153e8f2972fb9/src/Phinx/Db/Adapter/SQLiteAdapter.php#L676) which overwrites an array key in a loop instead of pushing to the end of the array. Thus, whatever last index SQLite reports internally is all Phinx ends up knowing about. It also doesn't distinguish between incidental indices created by constraints (which are an implementation detail) and actual indices which can be dropped  and need to be re-created when altering a table.

JKingweb:
Enhances `SQLiteAdapter::hasIndex()` and `SQLiteAdapter::hasIndexByName()` and related functionality in the following ways:

- Correctly enumerates a tables indices when searching for a match
- Matches named indices case-insensitively
- Matches indices by column only if _all_ columns of the index are present in the query array
- Adjusts the `Table::removeIndex()` and `Table::removeIndexByName()` handlers so that they too benefit from these improvements

I suspect there are also bugs in table re-creation related to indices, but I'm trying to avoid digging into the alteration functionality for now.

Fixes #1544.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -170,9 +170,10 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
     /**
      * @param string $tableName Table name
+     * @param boolean $quoted Whether to return the schema name and table name escaped and quoted. If quoted, the schema (if any) will also be appended with a dot
      * @return array
      */
-    protected function getSchemaName($tableName)
+    protected function getSchemaName($tableName, $quoted = false)
     {
         if (preg_match("/.\.([^\.]+)$/", $tableName, $match)) {
             $table = $match[1];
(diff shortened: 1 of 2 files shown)


garas:
I think indices `(a, b)` and `(b, a)` are different and should not match.

`hasIndex()` should match indices by columns only if _all_ columns of the index are present in the query array and in the  _same_ order.

`MysqlAdapter::hasIndex()` has such implementation.

JKingweb:
There's certainly an argument to be made for order mattering: commonly an (a,b,c) index is not useful for a (b,c) query, but a (c,b,a) index would be, so ordering can matter depending on why you're asking for a given index. 

Neither the PostgreSQL nor SQL Server adapters care about order, by the way. I'm not opposed to changing it, but it's not obvious to me what the right or least surprising behaviour actually is.

JKingweb:
Actually, no, it's pretty clear what the least surprising behaviour is now that I think about it.

If both (a,b,c) and (c,b,a) exist and one asks about index (c,b,a), removes (c,b,a) and then asks about it again, Phinx would still answer true, which would be silly. I'll change it, then.

lorenzo:
Thanks for your great work!

PR #1548, opened 2019-05-31
New hasPrimaryKey and hasForeignKey implementations for SQLite

### LINKED ISSUE #1538: SQLite: hasPrimaryKey() fails variously
The `SQLiteAdapter::hasPrimaryKey()` method, (implemented largely by `SQLiteAdapter::getPrimaryKey()`) is only able to find very specifically-formatted primary keys and has no support for composite keys at all.

### LINKED ISSUE #1539: SQLite: hasForeignKey() fails variously
Similar to issue #1538, the `SQLiteAdapter::hasForeignKey` method fails in a varied number of ways:

- It can return `true` when asking about a composite key while only simple keys exist on the same columns
- It can return `true` when asking about a key while only a composite key containing a superset of the queried columns exists
- Identifiers are treated as case-sensitive when SQLite does not do so in the course of normal operation
- Any table definition which does not use uppercase `REFERENCES` will yield `false`
- It will answer `true` for a table does does not contain a key if a temporary table of the same name does
- It always returns `false` if the table name contains a schema
- It always returns `false` if the table name starts with the substring `sqlite` _not_ followed by an underscore
- It probably fails spectacularly if a trigger attached to a table includes a string literal with the substring `REFERENCES`, but I have not confirmed this

JKingweb:
Corrected implementations of `SQLiteAdapter::hasPrimaryKey()` and `SQLiteAdapter::hasForeignKey()`. They were changed in the following ways:

- Avoids haphazard SQL scraping
- Reports correctly on composite primary keys
- Reports correctly on composite foreign keys
- Reports correctly on subsets of composite keys
- Fully case-insensitive
- Fails noisily when a named constraint is queried

Fixes #1538 and  fixes #1539.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -811,13 +811,11 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
         $columns = array_map('strtolower', (array)$columns);
         $primaryKey = array_map('strtolower', $this->getPrimaryKey($tableName));
 
-        if (array_diff($primaryKey, $columns)) {
+        if (array_diff($primaryKey, $columns) || array_diff($columns, $primaryKey)) {
             return false;
-        } elseif (array_diff($columns, $primaryKey)) {
-            return false;
-        } else {
-            return true;
         }
+        
+        return true;
     }
 
     /**
@@ -855,13 +853,10 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
         foreach ($foreignKeys as $key) {
             $key = array_map('strtolower', $key);
-            if (array_diff($key, $columns)) {
+            if (array_diff($key, $columns) || array_diff($columns, $key)) {
                 continue;
-            } elseif (array_diff($columns, $key)) {
-                continue;
-            } else {
-                return true;
             }
+            return true;
         }
 
         return false;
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -784,70 +804,67 @@ protected function getDropIndexByNameInstructions($tableName, $indexName)
...
+        $primaryKey = array_map('strtolower', $this->getPrimaryKey($tableName));
 
-        return empty($missingColumns);
+        if (array_diff($primaryKey, $columns)) {
+            return false;
+        } elseif (array_diff($columns, $primaryKey)) {
+            return false;
+        } else {
+            return true;
+        }
```
lorenzo:
```php
if (array_diff($primaryKey, $columns)) {
    return false;
}
if (array_diff($columns, $primaryKey)) {
    return false;
}
return true;
```
please

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -784,70 +804,67 @@ protected function getDropIndexByNameInstructions($tableName, $indexName)
...
-        return !array_diff((array)$columns, $foreignKeys);
+        foreach ($foreignKeys as $key) {
+            $key = array_map('strtolower', $key);
+            if (array_diff($key, $columns)) {
+                continue;
+            } elseif (array_diff($columns, $key)) {
+                continue;
+            } else {
+                return true;
+            }
```
lorenzo:
```php
if (array_diff($key, $columns) || array_diff($columns, $key)) {
    continue;
} 
return true;
```

please

src/Phinx/Db/Adapter/SQLiteAdapter.php
JKingweb: I removed the else, and also did an or comparison as in the other case.

lorenzo:
Thanks agains for such a great job!

Issue #1406, opened 2018-07-12
DSN not interpreted when using migration command (use Manager instead ?)


skwi:
When using the command line to start a migration on an environment configured with DSN, I get the following error : 
```
Could not determine database name! Please specify a database name in your config file
```
It seems that the `Migrate` command relies on the instruction `$this->getConfig()->getEnvironment($environment);` to get the environment options which does not relies on the `Environment` class, where the `parseAgnosticDsn` can transform the DSN to a proper environment configuration.

The `$envOptions = $this->getManager()->getEnvironment($environment)->getOptions();` is able to fetch the proper configuration.

I can write the PR to use the manager in the command but before doing so I prefer checking with you that it's the proper way to do it and make sure I'm not missing something.

lorenzo:
A PR would be great!

djcvijic:
I also came across this error today.

@skwi I tried your approach, and while the DSN does get parsed now, tokens no longer get replaced! It's as if Config::replaceTokens() is never called, even though it seems like it should be called even before Manager is instantiated...

I have a hunch this is connected to https://github.com/cakephp/phinx/pull/1433 but I ran out of time to try to combine these two solutions for now. Will keep exploring at some point.

TL;DR do not implement this fix yet, it is an incomplete solution.

alejandro-falkowski-gelato:
I see this is still an issue. Do you guys have an ETA when this will be released?

PR #1549, opened 2019-05-31
migrate and rollback commands use the env options, which parses the dsn

watermark:
Use the parsed DSN for the migrate and rollback command #1406

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -85,7 +85,8 @@ EOT
             $output->writeln('<info>using environment</info> ' . $environment);
         }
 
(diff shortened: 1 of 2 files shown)


lorenzo:
Tests are failing, can you please fix those?

watermark:
I updated the code to account for a null possibility, but the tests are still failing.  I think this may be a failing of the mock object (which I'm not very familiar with).  A function that should never return null is returning null due to the mocking.

lorenzo:
I think the problem it that you needed to update the tests, it is safe to assume the manager is passed.

dereuromark:
ping @watermark 

Are you still interested in finishing this? Would be awesome.
Otherwise, we have to close.

othercorey:
Closing due to no activity.

PR #1551, opened 2019-06-05
Overhaul of type system for SQLite

### LINKED ISSUE #1547: Problems with SQLite type conversion
There are a few relatively minor problems with how Phinx types are converted to SQL types and back again when using SQLite:

- Certain types (`date`, `time`, `datetime`, `timestamp`, `boolean`) have the wrong affinity (`NUMERIC`) when they should have `TEXT` (or `INTEGER` for `boolean`s) affinity
- Types are converted lossily when they could be converted losslessly
- The `varbinary` type is needlessly rejected
- The `decimal` type is accepted despite SQLite not supported decimal arithmetic
- The `enum` type is accepted despite enum emulation being incomplete and flaky
- Geospacial types are accepted despite SpaciaLite's method of operation being entirely dissimilar from PostGIS'.

Of course, rejecting types that were not rejected before would be an incompatible change, but I'm of the opinion it's better that users are not encouraged to make assumptions about equivalence.

JKingweb:
This change takes a different approach to Phinx-to-SQL type conversion when dealing with SQLite. For those not aware, SQLite has an unusual type system which allows a column to contain values of any type, and the types specified for a column function only as hints to [a column's affinity](https://sqlite.org/datatype3.html#type_affinity), which can be one of `INTEGER`, `FLOAT`, `TEXT`, `BLOB`, or `NUMERIC`. This affinity is used for type coercion during inserts and comparisons.

Other than how a given type string affects affinity (and one special case), SQLite doesn't care whether you type your column as `VARCHAR`, or `TEXT`, or `FALAFEL`, and the limit/precision/scale parameters are ignored entirely.

I have attempted therefore to make Phinx's type conversion adhere to the following principles:

- All Phinx types should translate to reasonable SQLite affinities
- Types as written to the database should mirror Phinx's own names as much as possible
- Types for which SQLite does not have adequate actual support should be rejected
- Any limits or scales are passed through as specified, except for `TINYINT(1)`

The third point is possibly contentious, as it is in some cases not backwards-compatible. In particular, I have chosen to reject `enum` for the following reasons:

- The current emulation of MySQL enums does not translate integer values to the strings matching the relevant index
- The current emulation does not read back out the allowed values, and this would be difficult to do
- Check constraints can be disabled, possibly leading to inconsistency anyway
- There are various surprising behaviours of MySQL enums which are also not emulated

I have also elected to reject `decimal` because SQLite does not support exact real numbers, and anyone specifically requesting exact numbers would not want to silently receive inexact numbers instead.

Further I have rejected most other complex types as their semantic compatibility with SQLite is not well defined. Exceptions include the various date types (SQLite does have various date-manipulation functions built in, and the likes of `CURRENT_TIMESTAMP` are supported), `json` and `jsonb` (JSON is well defined, and SQLite's JSON extension is sometimes available by default), and `uuid` (because UUIDs are usually opaque).

There are also other mildly backwards-imcompatible changes. Namely, various special limits on varchars or integers no longer yield a different type or null limit, with the exception of `TINYINT(1)`, which will yield a boolean with null limit.

Finally, I have adjusted a few of the existing tests in this area so that they pass, but frankly I am not confident in the ability of the existing test suite to actually catch regressions here, and I don't yet understand the way alterations are done well enough to have confidence no regressions will result. Someone familiar with the adapter's workings should probably evaluate this change's implications to the table-altering routines.

Fixes #1547.


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -1241,9 +1254,6 @@ protected function getColumnSqlDefinition(Column $column)
         if ($column->getPrecision() && $column->getScale()) {
             $def .= '(' . $column->getPrecision() . ',' . $column->getScale() . ')';
         }
-        if (($values = $column->getValues()) && is_array($values)) {
-            $def .= " CHECK({$this->quoteColumnName($column->getName())} IN ('" . implode("', '", $values) . "'))";
-        }
```
lorenzo: What happened to the support for `CHECK`?

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -1062,130 +1103,102 @@ protected function getDropForeignKeyByColumnsInstructions($tableName, $columns)
...
+            $type = $match[1];
+            $typeLC = strtolower($type);
+            $affinity = isset($match[2]) ? $match[2] : '';
+            $limit = isset($match[3]) && strlen($match[3]) ? (int)$match[3] : null;
+            $scale = isset($match[4]) && strlen($match[4]) ? (int)$match[4] : null;
+            if (isset(self::$supportedColumnTypes[$typeLC])) {
+                // the type is an explicitly supported type
+                $name = $typeLC;
+            } else {
+                switch ($typeLC) {
```
lorenzo: There is already too much nesting at this point, can you try to reduce it? One idea is breaking the function in smaller different functions.

src/Phinx/Db/Adapter/SQLiteAdapter.php
JKingweb: The check constraint is the mentioned enum emulation.

src/Phinx/Db/Adapter/SQLiteAdapter.php
JKingweb: I've replaced the switch with an array lookup.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -45,6 +45,67 @@
...
+        'mediuminteger' => self::PHINX_TYPE_INTEGER,
+        'bigint' => self::PHINX_TYPE_BIG_INTEGER,
+        'tinytext' => self::PHINX_TYPE_TEXT,
+        'mediumtext' => self::PHINX_TYPE_TEXT,
+        'longtext' => self::PHINX_TYPE_TEXT,
+        'tinyblob' => self::PHINX_TYPE_BLOB,
+        'mediumblob' => self::PHINX_TYPE_BLOB,
+        'longblob' => self::PHINX_TYPE_BLOB,
+        'real' => self::PHINX_TYPE_FLOAT,
+        'numeric' => self::PHINX_TYPE_FLOAT,
```
garas: DECIMAL and NUMERIC are aliases in MySQL, PostgreSQL and SQL Server.

It should map to `self::PHINX_TYPE_DECIMAL`, or removed as you've removed support for DECIMAL.

JKingweb: It does, which is why I had the alias. On the other hand, `NUMERIC` affinity is distinct from `FLOAT` affinity, so it would probably do more harm than good to have that alias. I'll remove it.

lorenzo:
Do you mind finding the listing errors?

JKingweb:
I'm not sure what you mean.

lorenzo:
Nevermind,  it is a rate limiting error in stickler. Thanks for the changes once again!

PR #1553, opened 2019-06-06
FIX updated_at in addTimestamps method

DvDMH:


src/Phinx/Db/Table.php
```diff
@@ -550,7 +550,8 @@ public function addTimestamps($createdAt = 'created_at', $updatedAt = 'updated_a
              ])
              ->addColumn($updatedAt, 'timestamp', [
                  'null' => true,
-                 'default' => null,
+	             'default' => 'CURRENT_TIMESTAMP',
```
lorenzo: I think that for backwards compatibility, it should not have a default.

rquadling:
Is there a way to get the fractional seconds in here?

dereuromark:
ping @DvDMH

dereuromark:
Seems to be a duplicate of https://github.com/cakephp/phinx/pull/1499/files
Either way we proceed (this or the other), they probably need a small test case.

othercorey:
Closing due to no activity.

PR #1554, opened 2019-06-09
New implementation of getColumns for SQLite

### LINKED ISSUE #1550: SQLite: getColumns() could be better
While `SQLiteAdapter::getColumns()` is mostly okay, it has two major problems:

1. The process for determining the identity column is aggressively simplistic and can trivially be fooled even using Phinx itself
2. Default-value expressions are not parsed, leading to inconsistent results compared to PostgreSQL and MySQL

Fixing the latter properly will fix the block on PR #1463.

JKingweb:
This change constitutes a new implementation of `SQLiteAdapter::getColumns()`. The method itself is not significantly changed, but it now calls out to two new methods: one which correctly identify a table's identity column (if any), and another which correctly interprets default-value expressions. Specifically, the following expressions are recognized and converted to applicable PHP types:

- String literals
- Integers and hexadecimal numbers
- Floats
- `NULL`
- `CURRENT_DATE`, `CURRENT_TIME`, and `CURRENT_TIMESTAMP`
- `TRUE` and `FALSE` (for SQLite 3.24 and later)

Anything else (including blob literals, for now) is wrapped in an instance of the new `Expression` class, a companion to the existing `Literal` class.

~Note that the existing tests for `getColumns()` have been retained as they appear to provide coverage for `getPhinxType()` and `getSqlType()`. Merging PR #1551 would make these tests obsolete. Note also that merging PR #1551 would require a change at line 542 in this patch, as indicated by the `FIXME`.~

Merging this patch should allow PR #1463 to be merged.

Fixes #1550.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -216,60 +216,6 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
     }
 
     /**
-     * Searches through all available schemata to find a table and returns an array
-     * containing the bare schema name and whether the table exists at all.
-     * If no schema was specified and the table does not exist the "main" schema is returned
-     *
-     * @param string $tableName The name of the table to find
-     * @return array
-     */
-    protected function resolveTable($tableName)
-    {
-        $info = $this->getSchemaName($tableName);
-        if ($info['schema'] === '') {
-            // if no schema is specified we search all schemata
-            $rows = $this->fetchAll('PRAGMA database_list;');
-            // the temp schema is always first to be searched
-            $schemata = ['temp'];
-            foreach ($rows as $row) {
-                if (strtolower($row['name']) !== 'temp') {
-                    $schemata[] = $row['name'];
-                }
-            }
-            $default = 'main';
-        } else {
-            // otherwise we search just the specified schema
-            $schemata = (array)$info['schema'];
-            $default = $info['schema'];
-        }
-
-        $table = strtolower($info['table']);
-        foreach ($schemata as $schema) {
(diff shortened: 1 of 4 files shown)


JKingweb:
Note that diff coverage is actually 100%. Some of it is only testable with SQLite 3.24, which was first bundled with PHP 7.3.0.

lorenzo:
Can you solve the merge errors, please?

JKingweb:
Absolutely. I don't have time presently, but should be able to do so sometime today.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -57,6 +58,29 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
     protected $suffix = '.sqlite3';
 
+    /** Indicates whether the database library version is at least the specified version
+     *
+     * @param string $ver The version to check against e.g. '3.28.0'
+     * @return boolean
+     */
+    public function databaseVersionAtLeast($ver)
```
lorenzo: This function seems to be repeating logic from `version_compare` https://php.net/version_compare

JKingweb: That's debatable: SQLite version numbers don't have the same rules as PHP version numbers. It would perform the same functionality, though, I suppose.

lorenzo: I think your implementation is the same as `version_compare`, probably will be more clear to use that func

JKingweb: I agree. I'll change it.

lorenzo:
Thanks again, you are doing a great job!

PR #1555, opened 2019-06-10
Fix #1474

### LINKED ISSUE #1474: Postgres: truncate doesn't reset auto increment
The truncate on Postgres's BDD doesn't reset the auto increments by default. For this, we have to add 'RESTART IDENTITY' to the query:

`TRUNCATE TABLE table_name RESTART IDENTITY`

The current behavior of truncate doesn't reset auto increment and there is no option to force it. The only way is with a raw sql script.

Should be truncate with reset auto increments the default behavior?

Thanks

rakshazi:
Hello,
By default, phinx postgres adapter does not reset autoincrement, but other adapters (like mysql) do that by default. That PR adds `RESTART IDENTITY` to truncate command in postgres adapter


lorenzo:
Thanks!

PR #1556, opened 2019-06-12
 Fix table truncation for SQLite

JKingweb:
Pursuant to #1474 and #1555, this patch applies an equivalent fix to `SQLiteAdapter::truncateTable()`.

The diff is a bit confused because most of `hasTable()` has been moved to a more diligent `resolveTable()` method which can be used to determine which schema a table is actually found in. This is necessary to delete from the correct `sqlite_sequence` table when no schema is specified.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -306,7 +306,7 @@ class SQLiteAdapter extends PdoAdapter implements AdapterInterface
 
         $table = strtolower($info['table']);
         foreach ($schemata as $schema) {
-            if ($schema === 'temp') {
+            if (strtolower($schema) === 'temp') {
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -305,19 +315,27 @@ public function hasTable($tableName)
...
-        return false;
+        return ['schema' => $defaultSchema, 'table' => $info['table'], 'exists' => false];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasTable($tableName)
+    {
+        return $this->resolveTable($tableName)['exists'];
```
lorenzo: I believe this is php 7 only syntax and we still support 5.7

lorenzo:
Thanks again!!

PR #1557, opened 2019-06-13
Decouple DirectActionInterface from PdoAdapter

JKingweb:
This patch moves the `DirectActionInterface` implementation of `PdoAdapter` out to a trait, `AbstractDirectActionTrait`. 

The rationale for this change is simple: SQLite is not actually capable of most direct actions, whilst unquestionably being a PDO-based adapter. The capability for direct actions has been shoehorned into the SQLite adapter, but it makes the code harder to read, harder to maintain, and ultimately makes the actions themselves less efficient than they could be.

For now the SQLite adapter is unchanged; this merely ends the assumption that all PDO-based adapters are capable of direct actions.


src/Phinx/Db/Adapter/AbstractDirectActionTrait.php
```diff
@@ -0,0 +1,423 @@
...
+use Phinx\Db\Table\Column;
+use Phinx\Db\Table\ForeignKey;
+use Phinx\Db\Table\Index;
+use Phinx\Db\Table\Table;
+use Phinx\Db\Util\AlterInstructions;
+
+/**
+ * A generic implementation of DirectActionInterface
+ *
+ * @author Rob Morgan <robbym@gmail.com>
```
lorenzo: can you remove this?

src/Phinx/Db/Adapter/AbstractDirectActionTrait.php
```diff
@@ -0,0 +1,423 @@
...
+trait AbstractDirectActionTrait
+{
+    /**
+     * Executes all the ALTER TABLE instructions passed for the given table
+     *
+     * @param string $tableName The table name to use in the ALTER statement
+     * @param AlterInstructions $instructions The object containing the alter sequence
+     * @return void
+     */
+    abstract protected function executeAlterSteps($tableName, AlterInstructions $instructions);
```
lorenzo: This an many other functions in this file are not part of `DirectActionInterface`

lorenzo: oh, I see why you had to do this. It's alright.

src/Phinx/Db/Adapter/AbstractDirectActionTrait.php
```diff
@@ -0,0 +1,423 @@
...
+     * @param Table $table Table
+     * @param string|null $newComment New comment string, or null to drop the comment
+     * @return AlterInstructions
+     */
+    abstract protected function getChangeCommentInstructions(Table $table, $newComment);
+
+    /**
+     * {@inheritdoc}
+     */
+    public function executeActions(Table $table, array $actions)
```
lorenzo: I would prefer if this function lived in another file. Seems to be very misplaced here

JKingweb: Perhaps a `DirectPlanExecutorTrait` file? I wasn't happy with its inclusion, either.

lorenzo: What’s the problem with having it in the PDO adapter?

JKingweb: The plan executor relies on the underlying implementations of the direct actions throughout. It is, in essence, a plan executor based on a series of direct actions.

lorenzo: No, it is calling `executeAlterSteps` which is different from Direct actions in that the alter steps usually comes from a planner, and are not executed by hand. That's why I don't think those correspond to `DirectActionsInterface`.

An alternative is to make another interface having `executeAlterSteps` and friends and making DirectActions extend it.

dereuromark:
ping @JKingweb

othercorey:
Closing due to no activity.

PR #1558, opened 2019-06-18
Added the possibility to change the default SRID for geometry type

LucasPadilha:
Main problem was I couldn't change the default SRID sent to geometry type columns.

Currently when using PostgresAdapter plus geometry type columns, all of them will have SRID 4326 and that can't be changed.

My PR adds a new option (srid) while adding columns so user can choose between leaving the default unchanged or not.

```php
  $table->addColumn('column_name', 'geometry', ['srid' => 4674])->save();
```

P.S.: Didn't write PHP Unit tests because this would need PostGIS extension enabled on Postgres.

### DIFF
--- a/src/Phinx/Db/Table/Column.php
+++ b/src/Phinx/Db/Table/Column.php
@@ -112,6 +112,11 @@ class Column
     protected $encoding;
 
     /**
+     * @var integer
+     */
+    protected $srid = null;
+
+    /**
      * @var array
      */
     protected $values;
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Table/Column.php
```diff
@@ -111,6 +111,11 @@ class Column
      */
     protected $encoding;
 
+    /**
+     * @var integer
```
dereuromark: |null missing?

src/Phinx/Db/Table/Column.php
```diff
@@ -605,6 +610,29 @@ public function getEncoding()
...
+    {
+        $this->srid = $srid;
+
+        return $this;
+    }
+
+    /**
+     * Gets the column SRID.
+     *
+     * @return int
```
dereuromark: Can also be |null, right?

dereuromark:
Apart from my comments this looks good to me.
Ideally, we have a small test case for this added functionality, however.

PR #1560, opened 2019-06-28
Introduce `--set` and `--unset` options for breakpoints.

rquadling:
This enhances the current breakpoint feature and provides greater control
over the state of a breakpoint, without the need to have any prior
knowledge of the state of the breakpoints.

### DIFF
--- a/tests/Phinx/Migration/ManagerTest.php
+++ b/tests/Phinx/Migration/ManagerTest.php
@@ -5656,7 +5656,7 @@ class ManagerTest extends TestCase
         $this->assertFalse($adapter->hasTable('user_logins_foo_bar'));
     }
 
-    public function testBreakpointsOperateAsExpected()
+    public function testBreakpointsTogglingOperateAsExpected()
     {
         if (!TESTS_PHINX_DB_ADAPTER_MYSQL_ENABLED) {
             $this->markTestSkipped('Mysql tests disabled. See TESTS_PHINX_DB_ADAPTER_MYSQL_ENABLED constant.');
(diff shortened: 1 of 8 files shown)


tests/Phinx/Migration/ManagerTest.php
```diff
@@ -5703,17 +5729,110 @@ public function testBreakpointsOperateAsExpected()
...
+                if (!is_numeric($column)) {
+                    $this->assertEquals($value, $resetVersions[$originalVersionKey][$column]);
+                }
+            }
+        }
+
+        // Wait until the second has changed.
+        $now = time();
+        while ($now == time()) {
+        }
```
garas: Don't waste CPU. `usleep()` for 0.1s between checks or calculate remaining microseconds.

Maybe add helper method as you use this block multiple times.

rquadling: As the phinxlog only holds the start_time to 1s resolution, we need to wait until the next second so that we can demonstrate that changing the breakpoint in the phinxlog doesn't also change the start_time which has (for some reason) `ON UPDATE CURRENT_TIMESTAMP()`.

rquadling: But sleeping for 1s makes a lot more sense.

garas: It is better to 10 * (check time + wait 0.1s) than check time 123456 times using 100% CPU.

garas: Also, I you can use [time_sleep_until()](https://www.php.net/manual/en/function.time-sleep-until.php)

rquadling: The main issue is that the code has to wait 1s regardless of how we wait. The 1s delay is best as even though PHP has waited until the next second, any deviation between the clocks between PHP and mysql could mean that I am requesting changes to the phinxlog in the same 1s from mysql's perspective. By waiting 1s, we have waited 1s, so unless time runs differently in mysql, we will always be in the next second and so the test can adequately prove we are not altering the start_time because we are setting the breakpoint.

lorenzo:
this looks great, thanks!

rquadling:
Excellent. I'd like to get this released. Is the process the same as in the wiki?

lorenzo:
@rquadling yes, same process, although instead of writing a changeling file, we write the changes in the release notes in GitHub.

rquadling:
Excellent. Will release now.

rquadling:
I've updated the `gh-pages` branch, but it doesn't seem to be reflected in http://docs.phinx.org/ which has a headline version of `v0.10.6` which matches the last change to `conf.py`, but not the previous changes in the `gh-pages` branch.

lorenzo:
@rquadling I'll try ti fix that

rquadling:
Cheers. If it is a process thing, can you add it to the wiki. I never did work out how gh-pages worked. Also, I thought the documentation was/is within CakePHP now?

lorenzo:
We moved the docs back to readthedocs, because they were hard to find by many.
@rquadling you did the changes to gh-pages correctly, the code there concerns this page which is currently showing the new version https://phinx.org

PR #1566, opened 2019-07-09
Fix inserting multiple rows without zero key

### LINKED ISSUE #1565: Bug: inserting multiple rows fails when keys are modified
I just encountered an issue with a migration. On inserting multiple rows of data with `$this->table('my_table')->insert($data);` the migration failed with multiple error notices: "PHP Notice:  Array to string conversion in /var/www/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php on line 306" and finally "SQLSTATE[42S22]: Column not found: 1054 Unknown column '2' in 'field list'."

Some debugging lead me to this line: https://github.com/cakephp/phinx/blob/7187e558ab70d5ac2197d197234ce6a5faf5cfbf/src/Phinx/Db/Table.php#L591. Turns out that the check to determine whether to insert multiple rows or a single row expects the key of the first item to be zero. Since I was building my array dynamically and ran it through array_map and array_filter, the first item happened to have 2 as it's key. Running my array through array_values before passing it to the insert function solves my issue, but it would be nice if this is done in the insert method (`$data = array_values($data);` would do the trick).

skript-cc:
Fixes #1565

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -588,7 +588,8 @@ class Table
     public function insert($data)
     {
         // handle array of array situations
-        if (isset($data[0]) && is_array($data[0])) {
+        $firstKey = array_shift(array_keys($data));
+        if ($firstKey !== null && is_array($data[$firstKey])) {
             foreach ($data as $row) {
                 $this->data[] = $row;
             }
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -588,7 +588,8 @@ class Table
     public function insert($data)
     {
         // handle array of array situations
-        $firstKey = array_shift(array_keys($data));
+        $keys = array_keys($data);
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Table.php
```diff
@@ -588,7 +588,8 @@ public function addTimestampsWithTimezone($createdAt = null, $updatedAt = null)
     public function insert($data)
     {
         // handle array of array situations
-        if (isset($data[0]) && is_array($data[0])) {
+        $firstKey = array_shift(array_keys($data));
```
lorenzo: You can only pass variables to `array_shift` If you pass a function call, it results in an error

skript-cc: Hmm, oke. Didn't saw any errors while running the tests. Whatever, easy change.

lorenzo:
Thanks!

PR #1573, opened 2019-07-17
Add semicolon for transaction sql lines for --dry-run

### LINKED ISSUE #1567: Dry run - missing semicolons in some some cases
IMO I'd be useful if --dry-run option generated working SQL script. I does not for now - I've noticed missing semicolons after:

- START TRANSACTION
- COMMIT
- ALTER TABLE

MasterOdin:
Fixes #1567. 

Using `--dry-run` was lacking the semi-colon at the end of any of the commands for transactions for MySQL, PostgreSQL, and SQLServer. This PR adds the missing semi-colons, and add tests to ensure proper behavior.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -114,7 +114,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
      */
     public function beginTransaction()
     {
-        $this->execute('BEGIN');
+        $this->execute('BEGIN;');
     }
 
     /**
@@ -122,7 +122,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
      */
     public function commitTransaction()
     {
-        $this->execute('COMMIT');
+        $this->execute('COMMIT;');
     }
 
     /**
@@ -130,7 +130,7 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
      */
     public function rollbackTransaction()
     {
-        $this->execute('ROLLBACK');
+        $this->execute('ROLLBACK;');
     }
 
     /**
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -235,23 +235,21 @@ class PostgresAdapter extends PdoAdapter implements AdapterInterface
         }
 
         $sql .= ')';
+        $this->execute($sql);
 
         // process column comments
         if (!empty($this->columnsWithComments)) {
             foreach ($this->columnsWithComments as $column) {
-                $sql .= $this->getColumnCommentSqlDefinition($column, $table->getName());
+                $this->execute($this->getColumnCommentSqlDefinition($column, $table->getName()));
             }
         }
 
         // set the indexes
         if (!empty($indexes)) {
             foreach ($indexes as $index) {
-                $sql .= $this->getIndexSqlDefinition($index, $table->getName());
+                $this->execute($this->getIndexSqlDefinition($index, $table->getName()));
             }
         }
-
-        // execute the sql
-        $this->execute($sql);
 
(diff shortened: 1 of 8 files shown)


lorenzo:
I think it would make more sense to add the semicolon in the printer, that way it would be impossible to break again in the future by adding more commands.

lorenzo:
@MasterOdin apparently you removed some `;` that are necessary, look at the failing tests.

MasterOdin:
@lorenzo The build failures have been fixed.

The biggest change is that I broke up the postgres adapter's createTable so that each sql string that gets generated gets executed one at a time instead of potentially grouping the create table, indexes, and column comments into one query and table comment into another query which seemed internally inconsistent. I think this also probably makes the dry-run output a bit clearer as well as to what's going on, but can revert if desired and just add the `;` as was originally there and have it all be one query again.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1572,6 +1572,29 @@ public function testDumpBulkinsert()
...
+        $table->addColumn('column1', 'string')
+            ->addColumn('column2', 'integer')
+            ->addColumn('column3', 'string', ['default' => 'test'])
+            ->save();
+        $this->adapter->commitTransaction();
+        $this->adapter->rollbackTransaction();
+
+        $actualOutput = $consoleOutput->fetch();
+        $this->assertStringStartsWith("START TRANSACTION;\n", $actualOutput, 'Passing the --dry-run doesn\'t drump the transaction to the output');
+        $this->assertStringEndsWith("COMMIT;\nROLLBACK;\n", $actualOutput, 'Passing the --dry-run doesn\'t drump the transaction to the output');
```
garas: _dump_ here and few more tests

MasterOdin: I'm not sure what you mean by this? Do you want me to test over the entire output message, and then should the other dump tests be changed to do that as well?

And for more tests, what did you have in mind that you'd like to see tested? This test (plus the dump ones before it) should cover the simple bases of "multiple executes in one query" and then the various other adapter tests cover that my removing the semi-colons from the functions hasn't broken anything.

dereuromark: You wrote "drump" there, a typo.

MasterOdin: Oh, totally misread the intention of that statement, sorry! Fixed the typo.

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -234,25 +234,23 @@ public function createTable(Table $table, array $columns = [], array $indexes =
...
-                $sql .= $this->getColumnCommentSqlDefinition($column, $table->getName());
+                $this->execute($this->getColumnCommentSqlDefinition($column, $table->getName()));
             }
         }
 
         // set the indexes
         if (!empty($indexes)) {
             foreach ($indexes as $index) {
-                $sql .= $this->getIndexSqlDefinition($index, $table->getName());
+                $this->execute($this->getIndexSqlDefinition($index, $table->getName()));
```
lorenzo: This is not ok, one of the improvement made before is to not execute side-effects before all the logic has been run first. I'd suggest buffering the SQL into an array and then execute each of the entries in the array separately.

lorenzo:
There appears to be a minor lint issue in the code, after that this looks good to me!

src/Phinx/Db/Adapter/PostgresAdapter.php
MasterOdin: Yeah, that makes a lot of sense. Fixed.

MasterOdin:
@lorenzo Lint issue fixed and merge conflicts resolved. Should be good to go. 👍

lorenzo:
Cool, good job on this one!

PR #1575, opened 2019-07-17
Cache names of created tables for exists check

### LINKED ISSUE #1569: Dry run wrong output with using "insert"
During our last deployment we learned that "--dry-run" option can output wrong sql script.

For this migration:

```sql
$this->table('my_table', ['id' => false, 'primary_key' => ['id']])
            ->addColumn('id', 'uuid')
            ->addColumn('name', 'string')
            ->save();

        $this->table('my_table')
            ->insert([
                'id' => '72daede1-1c23-4094-92cc-fa22e4419fb6',
                'name' => 'somename'
            ])->save();
```

the output is:

 ```
== 20190709152215 TestMigration: migrating
START TRANSACTION
CREATE TABLE `1mm_p4x_my_table` (`id` CHAR(36) NOT NULL, `name` VARCHAR(255) NOT NULL, PRIMARY KEY (`id`)) ENGINE = InnoDB CHARACTER SET utf8 COLLATE utf8_general_ci;
CREATE TABLE `1mm_p4x_my_table` (`id` INT(11) NOT NULL AUTO_INCREMENT, PRIMARY KEY (`id`)) ENGINE = InnoDB CHARACTER SET utf8 COLLATE utf8_general_ci;
INSERT INTO `1mm_p4x_my_table` (`id`, `name`) VALUES ('72daede1-1c23-4094-92cc-fa22e4419fb6', 'somename');
COMMIT
INSERT INTO `phinxlog` (`version`, `migration_name`, `start_time`, `end_time`, `breakpoint`) VALUES ('20190709152215', 'TestMigration', '2019-07-09 15:25:53', '2019-07-09 15:25:53', 0);
 == 20190709152215 TestMigration: migrated 0.0148s
```
as you can see - there is second "create table" and script fails because of it.

It's related to that "insert" - when I removed it - it was fine.

### PR DESCRIPTION
**Pull request by MasterOdin**
Fixes #1569 where because running under dry-run meant that the table was never actually created subsequent checks for existence of the table would fail, and an extra invalid `CREATE TABLE` statement would end up getting inserted into the dry-run log.

I'm not 100% satisfied with how this looks, but unsure of how might improve things and reduce the amount of copy/paste of things. It'll also still fail in the case where someone would directly call `execute()` with a create/drop/rename query.

### DIFF
--- a/src/Phinx/Db/Adapter/AbstractAdapter.php
+++ b/src/Phinx/Db/Adapter/AbstractAdapter.php
@@ -267,7 +267,8 @@ abstract class AbstractAdapter implements AdapterInterface
     /**
      * Adds user-created tables (e.g. not phinxlog) to a cached list
      */
-    public function addCreatedTable($tableName) {
+    public function addCreatedTable($tableName)
+    {
         if (substr_compare($tableName, 'phinxlog', -strlen('phinxlog')) !== 0) {
             $this->createdTables[] = $tableName;
         }
@@ -276,7 +277,8 @@ abstract class AbstractAdapter implements AdapterInterface
     /**
      * Updates the name of the cached table
      */
-    public function updateCreatedTableName($tableName, $newTableName) {
+    public function updateCreatedTableName($tableName, $newTableName)
+    {
         if (($key = array_search($tableName, $this->createdTables)) !== false) {
             $this->createdTables[$key] = $newTableName;
         }
@@ -285,7 +287,8 @@ abstract class AbstractAdapter implements AdapterInterface
     /**
      * Removes table from the cached created list
      */
-    public function removeCreatedTable($tableName) {
+    public function removeCreatedTable($tableName)
+    {
         if (($key = array_search($tableName, $this->createdTables)) !== false) {
             unset($this->createdTables[$key]);
         }
@@ -296,7 +299,8 @@ abstract class AbstractAdapter implements AdapterInterface
      * 
      * @return bool
      */
-    public function hasCreatedTable($tableName) {
+    public function hasCreatedTable($tableName)
+    {
         return in_array($tableName, $this->createdTables);
     }
 }
--- a/src/Phinx/Db/Adapter/AbstractAdapter.php
+++ b/src/Phinx/Db/Adapter/AbstractAdapter.php
@@ -266,6 +266,9 @@ abstract class AbstractAdapter implements AdapterInterface
 
     /**
      * Adds user-created tables (e.g. not phinxlog) to a cached list
+     *
+     * @param string $tableName The name of the table
+     * @return void
      */
     public function addCreatedTable($tableName)
     {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/Phinx/Db/Adapter/AbstractAdapter.php [hunk 1]
```diff
@@ -258,4 +263,44 @@ public function isDryRunEnabled()
...
+            $this->createdTables[] = $tableName;
+        }
+    }
+
+    /**
+     * Updates the name of the cached table
+     */
+    public function updateCreatedTableName($tableName, $newTableName)
+    {
+        if (($key = array_search($tableName, $this->createdTables)) !== false) {
```
dereuromark: It is usually cleaner to assign first, then go into the condition.

src/Phinx/Db/Adapter/AbstractAdapter.php
```diff
@@ -258,4 +263,44 @@ public function isDryRunEnabled()
...
+            unset($this->createdTables[$key]);
+        }
+    }
+
+    /**
+     * Check if the table is in the cached list of created tables
+     *
+     * @return bool
+     */
+    public function hasCreatedTable($tableName)
```
dereuromark: param missing in doc block

src/Phinx/Db/Adapter/AbstractAdapter.php
```diff
@@ -258,4 +263,56 @@ public function isDryRunEnabled()
...
+
+    /**
+     * Removes table from the cached created list
+     *
+     * @param string $tableName The name of the table
+     * @return void
+     */
+    public function removeCreatedTable($tableName)
+    {
+        if (($key = array_search($tableName, $this->createdTables)) !== false) {
```
dereuromark: I was also referring to all other cases of this :)

dereuromark:
I think we can't fix all cases, but this sure looks already like an improvement for most standard cases :+1:

src/Phinx/Db/Adapter/AbstractAdapter.php
MasterOdin: Agreed, fixed.

src/Phinx/Db/Adapter/AbstractAdapter.php
MasterOdin: Added missing param bits and return bits for all added functions.

src/Phinx/Db/Adapter/AbstractAdapter.php
MasterOdin: Agreed, sorry about missing this when I quickly glanced through the other changes.

MasterOdin:
Yeah, unless you put in a full SQL parser, the real complex cases around direct execute will never be fully captured. This will also not capture case of doing say an insert and save, then select over the table, and then use that to create additional queries. That case (assuming using internal functions) could be captured by just caching the various inserts/updates, but that's probably a bridge to burn if people actually report it. Can also probably just put a note in the docs about dry-run not being able to fully generate all SQL for complex cases (e.g. using execute, or the above example) due to it not being hooked up to a real DB.

src/Phinx/Db/Adapter/AbstractAdapter.php
```diff
@@ -258,4 +263,57 @@ public function isDryRunEnabled()
...
         return ($input && $input->hasOption('dry-run')) ? (bool)$input->getOption('dry-run') : false;
     }
+
+    /**
+     * Adds user-created tables (e.g. not phinxlog) to a cached list
+     *
+     * @param string $tableName The name of the table
+     * @return void
+     */
+    public function addCreatedTable($tableName)
```
lorenzo: I think all these new methods should be protected

MasterOdin: Done

lorenzo:
Thanks, this makes sense to me

raph1mm:
> Can also probably just put a note in the docs about dry-run not being able to fully generate all SQL for complex cases (e.g. using execute, or the above example) due to it not being hooked up to a real DB.

Good idea.

PR #1586, opened 2019-08-26
fix bad commit

Xakki:
This commit broke compatibility
https://github.com/symfony/console/commit/4510f04e70344d70952566e4262a0b11df39cb10#diff-25a75e093c4a0d5564a36cebbb1384b3R461

### DIFF
--- a/src/Phinx/Console/Command/Init.php
+++ b/src/Phinx/Console/Command/Init.php
@@ -38,13 +38,14 @@ use Symfony\Component\Console\Output\OutputInterface;
 class Init extends Command
 {
     const FILE_NAME = 'phinx';
-
+    private $name = 'init';
+    
     /**
      * {@inheritdoc}
      */
     protected function configure()
     {
-        $this->setName($this->getName() ?: 'init')
+        $this->setName($this->getName())
             ->setDescription('Initialize the application for Phinx')
             ->addOption('--format', '-f', InputArgument::OPTIONAL, 'What format should we use to initialize?', 'yml')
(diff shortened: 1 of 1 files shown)


src/Phinx/Console/Command/Init.php
```diff
@@ -38,13 +38,14 @@
...
-
+    private $name = 'init';
+    
     /**
      * {@inheritdoc}
      */
     protected function configure()
     {
-        $this->setName($this->getName() ?: 'init')
+        $this->setName($this->getName())
```
dereuromark: why not just `$this->setName('init')`?
After all, this current getter setter thing is super weird.

Xakki: For compatibility

dereuromark: What compatibility?

dereuromark:
I asked them to revert their BC breaking and bad change: https://github.com/symfony/console/commit/4510f04e70344d70952566e4262a0b11df39cb10#commitcomment-34845639

Xakki:
> I asked them to revert their BC breaking and bad change: [symfony/console@4510f04#commitcomment-34845639](https://github.com/symfony/console/commit/4510f04e70344d70952566e4262a0b11df39cb10#commitcomment-34845639)
> With private usage this easily becomes BC breaking on a major scale (as you cant reach the attr).

This option of course better.

dereuromark:
In the meantime we have a PR that solves it also future proof:
https://github.com/cakephp/phinx/pull/1588

Closing this then in favor of the other PR.

PR #1588, opened 2019-08-27
Fixed Command names

gnugat:
_Fix for #1585_

Symfony's Command require their name to return a string and don't allow a null
value, however in Phinx, in order to allow users to override the command names,
Commands in PhinxApplication are constructed without a name, and then in the
Command's configure method, a default name is set if getName is null.

Up until now, it worked as Symfony didn't enforce return types (types were
simply declared in PHPdoc), however for its version 5 Symfony plans to enforce
them, hence breaking Phinx commands.

To fix it, a COMMAND_NAME class constant has been introduced, and
PhinxApplication uses it when instantiating Commands.

### DIFF
--- a/src/Phinx/Console/Command/Breakpoint.php
+++ b/src/Phinx/Console/Command/Breakpoint.php
@@ -35,6 +35,8 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class Breakpoint extends AbstractCommand
 {
+    const COMMAND_NAME = 'breakpoint';
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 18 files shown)


dereuromark:
Travis should be green again once you merge master into this PR.

dereuromark:
git fetch
git merge origin/master

gnugat:
I've done the following instead:

```
git remote add upstream git@github.com:cakephp/phinx.git
git fetch upstream
git rebase upstream/master
git push -f
```

gnugat:
I've fixed the CS

dereuromark:
@gnugat What CS fail? https://travis-ci.org/cakephp/phinx/builds/577286086?utm_source=github_status&utm_medium=notification was green.

@rquadling Are you able to make a patch release for this asap?
( including the other https://github.com/cakephp/phinx/compare/v0.10.8...master diff changes )

I will merge this right now to resolve the broken master.

gnugat:
stickler-ci was (still is?) complaining about `use` statements in `PhinxApplication` not being in alphabetical order

dereuromark:
I would ignore stickler as long as travis CS is green.

Xakki:
@dereuromark Can add some tag, like 0.10.8 or 0.10.9 (may be need all support tags?)
This commit [Console fix return type declarations](https://github.com/symfony/console/commit/4510f04e70344d70952566e4262a0b11df39cb10#diff-25a75e093c4a0d5564a36cebbb1384b3R461) mark as v4.3.4 v3.4.31 tags

stof:
Another solution could be to use the `$defaultName` static property, which serves a similar need directly in `symfony/console`

saeideng:
@stof do you like to open a PR with using `defaultName` ?
it is good idea

Jeckerson:
Fast fix only with composer

```
"symfony/console": "~4.3.3, !=4.3.4",
"robmorgan/phinx": "^0.10.8"
```

lchhieu:
same error. How to fix it? Thanks

Issue #1590, opened 2019-08-27
Refactor travis matrix


dereuromark:
- [ ] Remove trusty build requirement.
- [ ] Instead of always creating mysql+postgres, we should run proper "per type" matrix runs
- [ ] and also cut down the matrix of php X db runs.

See https://github.com/dereuromark/cakephp-geo/blob/master/.travis.yml for a sane defaulting here. Maybe we can have a bit more, but running 5.6 until 7.3+ for all DBs is too much anyways.

dereuromark:
In the meantime I enabled daily cron run to find the failing master and issues faster in the future.

PR #1591, opened 2019-08-27
Update phpunit/phpunit requirement from >=5.7,<7.0 to >=5.7,<8.0

MasterOdin:
PR to overrule #1543. PHPUnit 8+ cannot be used unless this library goes PHP 7.1+ only as typehints were added across the codebase, most notably the `setUp` and `tearDown` functions got the `void` typehint. 

On the other PR, you will probably need to tell dependabot `@dependabot ignore this dependency` until such a time as phinx is 7.1+.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "symfony/yaml": "^2.8|^3.0|^4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": ">=5.7,<7.0",
+        "phpunit/phpunit": ">=5.7,<8.0",
         "sebastian/comparator": ">=1.2.3",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -55,8 +55,7 @@
     </filter>
 
     <logging>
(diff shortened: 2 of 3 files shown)


dereuromark:
For this to be merged I think we need a working --prefer-lowest check in matrix
for
https://github.com/cakephp/phinx/blob/master/.travis.yml

Refs https://github.com/cakephp/phinx/issues/1590 which links a travis file that contains such a check

dereuromark:
Both works. But without it we wouldnt know if previous phpunits still work.
You could make a PR against master for the matrix, and we could then just merge it into this one afterwards. To keep them separate.

dereuromark:
Please also see warnings in travis result:
```
PHPUnit 7.5.15 by Sebastian Bergmann and contributors.
  Warning - The configuration file did not pass validation!
  The following problems have been detected:
  Line 60:
  - Element 'log', attribute 'charset': The attribute 'charset' is not allowed.
  - Element 'log', attribute 'yui': The attribute 'yui' is not allowed.
  - Element 'log', attribute 'highlight': The attribute 'highlight' is not allowed.
```

dereuromark:
We could also merge this now and add prefer lowest afterwards?

PR #1592, opened 2019-08-27
add --prefer-lowest flag to composer install for travis-ci

### LINKED ISSUE #1587: Initial Unit Tests Do not run
Followed steps for How to contribute to Phinx. Used the Getting Started with docker instructions and get attached errors.

Capture

MasterOdin:
Adds a new job that runs `--prefer-lowest` for the lowest tested version of PHP (`5.6`) to ensure correctness of the library. Added per this comment https://github.com/cakephp/phinx/pull/1591#issuecomment-525316234 and trying to follow advice set in https://www.dereuromark.de/2019/01/04/test-composer-dependencies-with-prefer-lowest/ on adding it.

Fixes #1587

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "php": ">=5.6",
         "cakephp/collection": "^3.6",
         "cakephp/database": "^3.6",
+        "cakephp/datasource": "^3.6",
         "symfony/console": "^2.8|^3.0|^4.0",
         "symfony/config": "^2.8|^3.0|^4.0",
         "symfony/yaml": "^2.8|^3.0|^4.0"
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
     "require": {
         "php": ">=5.6",
         "cakephp/collection": "^3.6",
+        "cakephp/core": "^3.6",
         "cakephp/database": "^3.6",
         "cakephp/datasource": "^3.6",
         "symfony/console": "^2.8|^3.0|^4.0",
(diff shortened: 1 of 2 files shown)


dereuromark:
Looks like https://travis-ci.org/cakephp/phinx/jobs/577409759 reveals some too low constraints here.

dereuromark:
Yeah, some version bump will fix it. You can try to reproduce locally for this.

dereuromark:
One part is missing

    "cakephp/datasource": "^3.6",

in require-dev (as here the dependency wasnt autopulled yet)

composer.json
```diff
@@ -28,6 +28,7 @@
         "php": ">=5.6",
         "cakephp/collection": "^3.6",
         "cakephp/database": "^3.6",
+        "cakephp/datasource": "^3.6",
```
dereuromark: Ideally, this is in require-dev, as latest versions show it being auto-pulled as dependency.
But for now is also fine I think :)

dereuromark: Sounds like we should at some point fix the core here as well.

dereuromark: Also Installing cakephp/core (3.0.0): Downloading (100%) is not 3.6 which is needed :)

dereuromark:
Looks good. Let's merge this after https://github.com/cakephp/phinx/pull/1592 - and after we released a bugfix patch.

dereuromark:
Small conflict.

PR #1596, opened 2019-08-27
Use existing protected static defaultName for console commands

MasterOdin:
From the comments in #1588, namely:

> Another solution could be to use the `$defaultName` static property, which serves a similar need directly in `symfony/console`

_Originally posted by @stof in https://github.com/cakephp/phinx/pull/1588#issuecomment-525334319_

> @stof do you like to open a PR with using `defaultName` ?
> it is good idea

_Originally posted by @saeideng in https://github.com/cakephp/phinx/pull/1588#issuecomment-525458265_

This uses the `$defaultName` static variable available in Symfony\Console (https://github.com/symfony/console/blob/6b101586a5938f1a5958ef0d6ca598755b6d6ea8/Command/Command.php#L32-L35) to set a name for the command to be used if a different name is not passed in the constructor.

Note, this does requiring bumping the dependencies for symfony/console to 3.4 as that is where `$defaultName` was added. I also removed the RollbackMockStub in the one test as I'm not sure why it was there and there was no difference in the test output.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,9 +28,9 @@
         "php": ">=5.6",
         "cakephp/collection": "^3.6",
         "cakephp/database": "^3.6",
-        "symfony/console": "^2.8|^3.0|^4.0",
-        "symfony/config": "^2.8|^3.0|^4.0",
-        "symfony/yaml": "^2.8|^3.0|^4.0"
+        "symfony/console": "^3.4|^4.0",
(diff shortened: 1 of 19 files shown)


dereuromark:
My proposal: Let's first release a patch that fixes this without the version bump.
Then we have time to address this solution and how to fix it for the next minor. 
Does that sound good?

PR #1598, opened 2019-08-28
Update version in composer.json

jkavalik:
Last release missed that and now does not work through composer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,6 @@
     "keywords": ["phinx", "migrations", "database", "db", "database migrations"],
     "homepage": "https://phinx.org",
     "license": "MIT",
-    "version": "0.11.1",
     "authors": [{
         "name": "Rob Morgan",
         "email": "robbym@gmail.com",
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "keywords": ["phinx", "migrations", "database", "db", "database migrations"],
     "homepage": "https://phinx.org",
     "license": "MIT",
-    "version": "0.10.8",
+    "version": "0.11.1",
     "authors": [{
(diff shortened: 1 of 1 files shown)


dereuromark:
ping @markstory 
Maybe we have to automate this. Seems like this can easily go wrong.

dereuromark:
Or we just remove this? We usually dont do this anywhere anyway, as this is redundant data - unDRY and as mentioned highly likely to be wrong/misleading if not completely automated in the release process.

jkavalik:
Yes, imho better to not have it in there at all.

garas:
It could read version from `vendor/composer/installed.json` instead.

jkavalik:
I see.. What are the alternatives?

1. keep it in there and update manually
2. remove version info
3. use some other source (possibly `vendor/composer/installed.json`)

garas:
Version could stay if https://github.com/cakephp/phinx/wiki/Releasing-a-new-version-of-Phinx#steps-to-release-a-new-version would be automated. It is now too error prone.

dereuromark:
@markstory Can we formalized the process here and use the suggested automation for the beginning? To take out human error as factor on most part?
All releasers should then be fully aware of this requirement.

markstory:
For other CakePHP projects that have multi-step release processes, I've had good success with using Make to automate them. We could try something similar here.

dereuromark:
OK
Closing then for now as we will have to come up with automation instead.

glensc:
I use myself https://github.com/Ocramius/PackageVersions:
- - Use `ocramius/package-versions` for version info (@glensc, [#549](https://github.com/eventum/eventum/pull/549))

PR #1600, opened 2019-08-28
Do not print out status table when specifying --format json

### LINKED ISSUE #962: Status command with JSON format option prints text AND json
The `phinx status` works well with text format `-f text` but prints both formats when selecting JSON format `-f json`.

Here the output (notice the last line):

```
#> php vendor/bin/phinx -f json

Phinx by Rob Morgan - https://phinx.org. version 0.6.4

using config file ./phinx.yml
using config parser yaml
using migration path /foo/bar/data/migrations
using seed path /foo/bar/data/seeds
using environment development
using format json

 Status  Migration ID    Started              Finished             Migration Name
     up  20160829130325  2016-09-01 19:04:54  2016-09-01 19:04:55  xxx
     up  20160830123204  2016-09-01 19:06:02  2016-09-01 19:06:02  xxx
     up  20160830164920  2016-09-01 19:06:02  2016-09-01 19:06:02  xxx
     up  20160831142539  2016-09-01 19:06:02  2016-09-01 19:06:02  xxx
     up  20160901165958  2016-09-01 19:06:58  2016-09-01 19:06:58  xxx
     up  20160907155023  2016-09-07 17:52:35  2016-09-07 17:52:35  xxx
     up  20160907162327  2016-09-07 18:29:17  2016-09-07 18:29:17  xxx
     up  20160908151045  2016-09-13 17:40:52  2016-09-13 17:40:52  xxx
     up  20160908152839  2016-09-08 17:30:53  2016-09-08 17:30:53  xxx
     up  20160908154600  2016-09-08 17:49:26  2016-09-08 17:49:26  xxx
     up  20160908160020  2016-09-08 18:21:07  2016-09-08 18:21:07  xxx
     up  20160913085001  2016-09-27 17:29:23  2016-09-27 17:29:23  xxx
     up  20160913153834  2016-09-13 17:41:12  2016-09-13 17:41:12  xxx
     up  20160916142746  2016-09-27 17:29:23  2016-09-27 17:29:23  xxx
     up  20160916163002  2016-09-27 17:29:23  2016-09-27 17:29:23  xxx
     up  20160919175002  2016-09-27 17:29:23  2016-09-27 17:29:23  xxx
     up  20160922165747  2016-09-28 18:13:05  2016-09-28 18:13:05  xxx
     up  20160926134227  2016-09-27 17:29:23  2016-09-27 17:29:23  xxx
     up  20160926151932  2016-09-28 18:13:05  2016-09-28 18:13:05  xxx
   down  20160928170557                                            xxx
   down  20160928182625                                            xxx

{"pending_count":21,"migrations":[{"migration_status":"up","migration_id":"20160829130325","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160830123204","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160830164920","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160831142539","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160901165958","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160907155023","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160907162327","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160908151045","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160908152839","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160908154600","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160908160020","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160913085001","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160913153834","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160916142746","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160916163002","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160919175002","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160922165747","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160926134227","migration_name":"xxx"},{"migration_status":"up","migration_id":"20160926151932","migration_name":"xxx"},{"migration_status":"down","migration_id":"20160928170557","migration_name":"xxx"},{"migration_status":"down","migration_id":"20160928182625","migration_name":"xxx"}]}
```

I don't know if it works as expected, but it seems weird and very hard to parse.  
I advice to add the condition `(null === $format || 'text' == $format)` here https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Migration/Manager.php#L108

MasterOdin:
Sort of fixes #962

If using `--format json` for `phinx status`, it will print out just the JSON object and not the full status table as well.

Before:
```
Phinx by CakePHP - https://phinx.org. 0.11.1

using config file ./phinx.yml
using config parser yaml
using migration paths
 - /Users/mpeveler/Work/Github/phinx/db/migrations
using seed paths
warning no environment specified, defaulting to: production
using format json
ordering by creation time

 Status  [Migration ID]  Started              Finished             Migration Name
     up  20190828175258  2019-08-28 17:55:41  2019-08-28 17:55:41  TestMigration

{"pending_count":0,"missing_count":0,"total_count":1,"migrations":{"20190828175259":{"migration_status":"up","migration_id":"20190828175258","migration_name":"TestMigration"}}}
```

After:
```
Phinx by CakePHP - https://phinx.org. 0.11.1

using config file ./phinx.yml
using config parser yaml
using migration paths
 - /Users/mpeveler/Work/Github/phinx/db/migrations
using seed paths
warning no environment specified, defaulting to: production
using format json
ordering by creation time

{"pending_count":0,"missing_count":0,"total_count":1,"migrations":{"20190828175259":{"migration_status":"up","migration_id":"20190828175258","migration_name":"TestMigration"}}}
```

Working on a separate PR/proposal for how to deal with the header text in a nicer fashion making it easier to use this (and other) command in a pipeline without having to use tail or grep.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -100,8 +100,8 @@ class Manager
     /**
      * Prints the specified environment's migration status.
      *
-     * @param string $environment
-     * @param null $format
+     * @param string $environment environment to print status of
+     * @param string|null $format format to print status in (either text, json, or null)
      * @return int 0 if all migrations are up, or an error code
      */
     public function printStatus($environment, $format = null)
@@ -114,6 +114,10 @@ class Manager
         $missingCount = 0;
         $pendingMigrationCount = 0;
         $finalMigrations = [];
+        $verbosity = $output->getVerbosity();
+        if ($format === 'json') {
+            $output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
+        }
         if (count($migrations)) {
             // TODO - rewrite using Symfony Table Helper as we already have this library
             // included and it will fix formatting issues (e.g drawing the lines)
@@ -234,9 +238,11 @@ class Manager
 
         // write an empty line
         $output->writeln('');
+
         if ($format !== null) {
             switch ($format) {
                 case 'json':
+                    $output->setVerbosity($verbosity);
                     $output->writeln(json_encode(
                         [
                             'pending_count' => $pendingMigrationCount,
--- a/tests/Phinx/Migration/ManagerTest.php
+++ b/tests/Phinx/Migration/ManagerTest.php
@@ -198,7 +198,7 @@ class ManagerTest extends TestCase
         $this->assertSame(0, $return);
         rewind($this->manager->getOutput()->getStream());
         $outputStr = trim(stream_get_contents($this->manager->getOutput()->getStream()));
-        $this->assertStringEndsWith('{"pending_count":0,"missing_count":0,"total_count":2,"migrations":[{"migration_status":"up","migration_id":"20120111235330","migration_name":"TestMigration"},{"migration_status":"up","migration_id":"20120116183504","migration_name":"TestMigration2"}]}', $outputStr);
+        $this->assertEquals('{"pending_count":0,"missing_count":0,"total_count":2,"migrations":[{"migration_status":"up","migration_id":"20120111235330","migration_name":"TestMigration"},{"migration_status":"up","migration_id":"20120116183504","migration_name":"TestMigration2"}]}', $outputStr);
     }
 
     public function testPrintStatusMethodWithNamespace()
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ environment:
       - dependencies: highest
         db: 2012
         php_ddl: 56_nts
-        php_zip: php-5.6.28-nts-Win32-VC11-x86.zip
+        php_zip: php-5.6.40-nts-Win32-VC11-x86.zip
         db_dsn: 'sqlserver://sa:Password12!@.\SQL2012SP1/phinxtesting?MultipleActiveResultSets=false'
         sqlsrv: https://download.microsoft.com/download/C/D/B/CDB0A3BB-600E-42ED-8D5E-E4630C905371/SQLSRV32.EXE
 


tests/Phinx/Migration/ManagerTest.php
```diff
@@ -164,6 +164,45 @@ public function testPrintStatusMethod()
...
+                                'migration_name' => '',
+                                'breakpoint' => '0',
+                            ],
+                    ]
+                ));
+
+        $this->manager->setEnvironments(['mockenv' => $envStub]);
+        $this->manager->getOutput()->setDecorated(false);
+        $return = $this->manager->printStatus('mockenv', 'json');
+        $this->assertEquals(0, $return);
```
dereuromark: ideally Same() to check also Type.

MasterOdin:
So it looks like there's a bug where if the keys of the `$migrations` array are integers it gets printed one way versus if they're a string they get printed another way. The former is what's happening on Travis and the latter is on Windows.

PHP should (normally) automatically cast any string key that looks like an integer into an integer, e.g.:
```
$a = ["1" => 'a'];
var_dump(array_keys($a));
```
should give: 
```
array(1) {
  [0] =>
  int(1)
}
```

but on windows it's giving:
```
array(1) {
  [0] =>
  string(1)
}
```

MasterOdin:
Realized from #1602 what's going on Windows and why there's a string. It's because the build there is 32-bit and the date strings are past the value of MAX_PHP_INT for 32bit and so have to be used as a strings whereas Travis uses 64-bit and is happy to store things as integers with its much wider range.

I guess the only real solution here is to just fix whatever issue exists in printStatus so that it behaves the same regardless if the keys of the array are integer or string.

dereuromark:
Changes make sense to me. For format we do not want to mess it up, otherwise keep it as is. :+1:

dereuromark:
Would be good to have @lorenzo s feedback before merging, though.

src/Phinx/Migration/Manager.php
```diff
@@ -158,7 +160,9 @@ public function printStatus($environment, $format = null)
                 // this means we have no up migrations, so we write all the missing versions already so they show up
                 // before any possible down migration
                 foreach ($missingVersions as $missingVersionCreationTime => $missingVersion) {
-                    $this->printMissingVersion($missingVersion, $maxNameLength);
+                    if ($format !== 'json') {
+                        $this->printMissingVersion($missingVersion, $maxNameLength);
+                    }
```
lorenzo: Given how much this code is repeated, can we add the check for the format in the function itself?

dereuromark:
ping @MasterOdin

MasterOdin:
Sorry, I have ended up being busy the last few weeks with moving to a new country and starting a new job.

The main thing currently blocking me is:

> I guess the only real solution here is to just fix whatever issue exists in printStatus so that it behaves the same regardless if the keys of the array are integer or string.

So that behavior is consistent across 32 and 64 bit Phinx for printing out the status array.

dereuromark:
@lorenzo Are there any hard blockers? Or can we also do follow up PRs here terms of further improving this?

MasterOdin:
@dereuromark @lorenzo looking at this further, and running a couple of local tests, I think actually the 64-bit format is a _regression_ on 32-bit and that this behavior should be fixed. I've created #1612 to document the current behavior on the two platforms as I was able to get it, and a case that clearly (to me) demonstrates a bug in the 64-bit format.

I would argue that fixing #1612 is a blocker to this PR being merged so that I can write a single test for behavior for this change, and that resolving the merge conflicts between the two will not be that significant. Assuming you folks agree, I'm going to focus on trying to fix #1612 first before addressing the current merge conflict and @lorenzo's comments.

src/Phinx/Migration/Manager.php
MasterOdin: Yeah, these if statements were a clumsy way to go about it. I've implemented a vastly simpler solution.

MasterOdin:
Alright, with #1612 done and not having to fight with the output of appveyor vs travis, I've addressed the review comment here, vastly simplifying my original approach to something way more maintainable.

@dereuromark, @lorenzo this should be ready for review!

PR #1605, opened 2019-08-29
Fix PHP7.4

dereuromark:
Resolves https://github.com/cakephp/phinx/pull/1603

PR #1606, opened 2019-08-29
Fix ordering of arguments to implode

MasterOdin:
Fixes a batch of the errors in https://travis-ci.org/cakephp/phinx/jobs/578364719.

While implode allows either ordering currently for historical reasons, PHP 7.4 has now deprecated the old syntax.

### DIFF
--- a/src/Phinx/Util/Util.php
+++ b/src/Phinx/Util/Util.php
@@ -108,7 +108,7 @@ class Util
     {
         $arr = preg_split('/(?=[A-Z])/', $className);
         unset($arr[0]); // remove the first element ('')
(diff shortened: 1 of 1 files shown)


dereuromark:
If you want you can merge https://github.com/cakephp/phinx/pull/1605 into this one.

MasterOdin:
I think I'd like to keep each change necessary for PHP 7.4 separate just to follow my preferred git flow, and that I've got no idea how complicated the sqlite changes are going to be.

dereuromark:
That is fair enough.
But you can just make multiple atomic commits here inside a PR. That suffices. 
With the merged in travis run we can confirm that we are all done here then.

Merging now.

PR #1614, opened 2019-09-23
Make same output for migrations value of status JSON for 32 and 64 bit PHP

### LINKED ISSUE #1612: Unify json status print for 32 and 64 bit PHP
From #1600, depending on the bit of your PHP (32 or 64), the output for `phinx status --format=json` will give a different result. Given the age of the project and the format of the table, it seems to me that the 32-bit output is the right and that 64-bit has introduced a regression, especially as the keys are (I think) utterly meaningless for usage:

For reference for the two types:
#### 64bit:
```
{
	"pending_count": 0,
	"missing_count": 0,
	"total_count": 2,
	"migrations": {
		"20120116183505": {
			"migration_status": "up",
			"migration_id": "20120111235330",
			"migration_name": "TestMigration"
		},
		"20120116183506": {
			"migration_status": "up",
			"migration_id": "20120116183504",
			"migration_name": "TestMigration2"
		}
	}
}
```

#### 32bit:
```
{
	"pending_count": 0,
	"missing_count": 0,
	"total_count": 2,
	"migrations": [{
		"migration_status": "up",
		"migration_id": "20120111235330",
		"migration_name": "TestMigration"
	}, {
		"migration_status": "up",
		"migration_id": "20120116183504",
		"migration_name": "TestMigration2"
	}]
}
```

This is more pronounced when doing a print where not all migrations have been run. Consider the following example:

```
ordering by creation time

 Status  [Migration ID]  Started              Finished             Migration Name
   down  20190916153736                                            Test
```

Printing out the status on a 64-bit machine produces:
```
{
	"pending_count": 1,
	"missing_count": 0,
	"total_count": 1,
	"migrations": {
		"20190916153736": {},
		"20190916153737": {
			"migration_status": "down",
			"migration_id": "20190916153736",
			"migration_name": "Test"
		}
	}
}
```
which has an extra key/value pair that is empty and does not even make sense to exist.

MasterOdin:
Closes #1612, unblocks #1600.

The output for the `status` command is now the same on both 32-bit and 64-bit PHP, as a list of migrations instead of a map with random keys.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -242,7 +242,7 @@ class Manager
                             'pending_count' => $pendingMigrationCount,
                             'missing_count' => $missingCount,
                             'total_count' => $migrationCount + $missingCount,
-                            'migrations' => $migrations_final
+                            'migrations' => $finalMigrations
                         ]
                     ));
                     break;
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -113,6 +113,7 @@ class Manager
         $migrationCount = 0;
         $missingCount = 0;
         $pendingMigrationCount = 0;
+        $migrations_final = [];
         if (count($migrations)) {
             // TODO - rewrite using Symfony Table Helper as we already have this library
             // included and it will fix formatting issues (e.g drawing the lines)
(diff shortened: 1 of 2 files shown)


src/Phinx/Migration/Manager.php
```diff
@@ -113,6 +113,7 @@ public function printStatus($environment, $format = null)
         $migrationCount = 0;
         $missingCount = 0;
         $pendingMigrationCount = 0;
+        $migrations_final = [];
```
dereuromark: $finalMigrations maybe?

MasterOdin: that's better, and in the right case. Clearly did not give myself enough time to context switch between projects.

PR #1616, opened 2019-09-26
Fix: CamelCase Converted to PascalCase

ridvanaltun:
I think using PascalCase is more accurate.


dereuromark:
We are usually using the terms
- camelBacked
- CamelCase

here everywhere for consistency, see also CakePHP core.
Any reason this has to change?

ridvanaltun:
You are right, this PR not needed.

PR #1620, opened 2019-10-08
Remove duplication of get[Migration|Seed]Files functions

MasterOdin:
`Manager::getMigrationFiles` and `Manager::getSeedFiles` did more or less the same with just a deviation of the paths used. Given that the functions are marked as just `protected` instead of `private`, I elected to not remove them in-case they're being modified by downstream consumers of phinx.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -793,23 +793,7 @@ class Manager
      */
     protected function getMigrationFiles()
     {
-        $config = $this->getConfig();
-        $paths = $config->getMigrationPaths();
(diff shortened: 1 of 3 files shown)


dereuromark:
We do you need Util::getFiles()?
If this is the same class, there could just be a protected method right?

MasterOdin:
Yeah, my first draft of this was to move stuff into a private function within the Manager, but then figured Util was the better place so that the function itself could be directly unit tested and it fits with the other functions in the file already.

dereuromark:
not private, protected :)
but fair enough.

PR #1621, opened 2019-10-09
Update migrations.rst

juffalow:
The `drop()` method is not returning anything, therefore calling `save()` fails.

```php
// Table.php

...
    /**
     * Drops the database table.
     *
     * @return void
     */
    public function drop()
    {
        $this->getAdapter()->dropTable($this->getName());
    }
...
```

Also `dropTable` method is not returning anything.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -601,10 +601,6 @@ Dropping a Table
 Tables can be dropped quite easily using the ``drop()`` method. It is a
 good idea to recreate the table again in the ``down()`` method.
 
-Note that like other methods in the ``Table`` class, ``drop`` also needs ``save()``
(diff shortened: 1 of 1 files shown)


dereuromark:
Are you sure?
https://github.com/cakephp/phinx/commit/5c555079a0c6e9c0d79209e815d646d283728a6a shows that this has been explicitly added after recent changes, and as such you might be on an outdated version (major) instead?

juffalow:
You are right! I am sorry. But in this case, I would update the *major* version, because it would not work (or should not work) with the old way. Or?

juffalow:
@dereuromark can we close this PR? It is not relevant anymore.

PR #1622, opened 2019-10-11
Print error when using invalid environment

MasterOdin:
Right now, if you pass an invalid environment to the commands, it is inconsistent what will happen:

Throws `\InvalidArgumentException` from within the Manager.php:
* breakpoint
* rollback
* status (exception is printed after the first line of the table is printed)

Prints `<error>Could not determine database name! Please specify a database name in your config file.</error>`:
* migrate
* seed:run

The first is probably not quite the right thing as the bad environment could come from the command line or from the phinx yml file itself, and the second case (database name) is just misleading.

This unifies things to always print out `<error>The environment "%s" does not exist</error>` message (same as the `\InvalidArgumentException`, but without the exception) in all cases, as well as before actually attempting to execute the given command.

### DIFF
--- a/src/Phinx/Console/Command/Breakpoint.php
+++ b/src/Phinx/Console/Command/Breakpoint.php
@@ -69,7 +69,7 @@ EOT
      *
      * @param \Symfony\Component\Console\Input\InputInterface $input
      * @param \Symfony\Component\Console\Output\OutputInterface $output
-     * @return void
+     * @return int integer 0 on success, or an error code.
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
@@ -115,5 +115,7 @@ EOT
             // Toggle the breakpoint.
             $this->getManager()->toggleBreakpoint($environment, $version);
         }
+
+        return 0;
     }
 }
--- a/src/Phinx/Console/Command/Breakpoint.php
+++ b/src/Phinx/Console/Command/Breakpoint.php
@@ -90,6 +90,7 @@ EOT
 
         if (!$this->getConfig()->hasEnvironment($environment)) {
             $output->writeln(sprintf('<error>The environment "%s" does not exist</error>', $environment));
+
             return 1;
         }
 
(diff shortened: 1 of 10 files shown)


src/Phinx/Console/Command/Breakpoint.php
```diff
@@ -115,5 +115,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             // Toggle the breakpoint.
             $this->getManager()->toggleBreakpoint($environment, $version);
         }
+
+        return 0;
```
dereuromark: You should use the class constants instead of magic numbers.

MasterOdin: Err, what class constant? The only ones I could find are
https://github.com/cakephp/phinx/blob/6b1c0edb507bb5b3ffb5056015a75b32052b6489/src/Phinx/Migration/Manager.php#L76-L84

I agree with using constants, can make a follow-up PR that makes those changes across all commands if desired (after decision on where those constants should live).

garas: `Command::CODE_SUCCESS`

MasterOdin: Yeah, there are no constants within Phinx or its current dependencies that can be used for this unfortunately, and I think doing it here for all commands would just muddle up this PR changeset.

MasterOdin: Within this issue would also be making more consistent for when an exception is raised or when to print an error and return exit code.

dereuromark: https://github.com/cakephp/phinx/blob/master/src/Phinx/Console/Command/AbstractCommand.php#L47
Couldnt this just get 

    const CODE_SUCCESS = 0; 
    const CODE_ERROR = 1;

added?

MasterOdin: However, that then differs from how the codes are defined in Manager for status (they use a `EXIT_STATUS_` prefix). I think it would be best to unify everything in a separate PR after ironing out the details of #1623 then expanding the scope of this PR to do that, and for now just follow what `migrate` was doing which was already just returning magic numbers.

PR #1626, opened 2019-10-16
Suggestions of improvement in CONTRIBUTING file

smmd:


CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ Great, so you want to contribute. Let's get started:
 
 1. Clone your repository to a local directory on your development box.
 
-1. If you do not have Composer set up already, install it:
+1. If you do not have [Composer](getcomposer.org) set up already, install it:
```
garas: Link needs `https://`

CONTRIBUTING.md
```diff
@@ -63,13 +69,27 @@ Great, so you want to contribute. Let's get started:
...
+    ```
+    docker-compose run --rm phinx
+    ```
 
-To get started, just run `docker-compose run --rm phinx`. It will download all
-images, build & start development container and switch you to it. So just install dependencies with `composer update --prefer-lowest` and run unittests with `vendor/bin/phpunit`.
+1. Install dependencies:
+
+    ```
+    composer update --prefer-lowest
```
dereuromark: Why the prefer-lowest? That is weird. Usually this is only done in CI to ensure BC to those versions promised, but should not be part of normal dev and local default testing.

smmd: That's true, I just keep the commands of the old version of this file, but it is a good opportunity to change it, I will remove it.

Issue #1628, opened 2019-10-23
run Phinx via custom cli tool?


ltdev22:
Would appreciate some feedback if possible. I'm having a php app in which I'm using a custom cli tool like Artisan for generating files etc which is based on Symfony's Console. So I'm having a folder of commands like this

```php
<?php

namespace App\Console\Commands;
use App\Console\Command;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputOption;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

class HelloWorldCommand extends Command
{

    protected $command = 'say:hello';

    protected $description = 'Say the classic "hello world!"';

    public function handle(InputInterface $input, OutputInterface$output)
    {
            $this->info('Hello World !');
    }

    protected function arguments()
    {
        return [
            ['name', InputArgument::REQUIRED, 'Enter your name.'],
        ];
    }

    protected function options()
    {
        return [
            ['somehthing', 'r', InputOption::VALUE_OPTIONAL, 'Do something.', 1],
        ];
    }
}
```

I'm also using Phinx for creating and running migrations and seeds. So whenever I want to run a phinx command I do something like `vendor/bin/phinx create/migrate/rollback etc`, which works fine. However I'm looking for a way to run the phinx commands via the cli tool.

Is there a better way of doing this instead of doing something like this `exec('vendor/bin/phinx migrate/rollback etc');` in my handle method? I think this looks a bit hacky

sabl0r:
I implemented this some time ago: https://github.com/cakephp/phinx/issues/1365#issuecomment-396245197

dereuromark:
Maybe someone can put this into docs as showcase? Then it might be found more easily.

Issue #1365, opened 2018-06-11
Run migrations progmatically


iamdann:
I'm trying to integrate Phinx into my own custom CLI with [Symfony Console Component](https://symfony.com/doc/current/components/console.html). 

My goal is to be able to reset the database completely and re-run all the migrations. The first part I accomplished, but I have no clue how I could run the migrations progmatically. 

I know I can do something like `exec('vendor/bin/phinx migrate');` and `exec('vendor/bin/phinx seed:run');` but that's a pretty dirty solution IMO 🤷‍♂️ 

Anyone has some pointers?

lorenzo:
Maybe you can look at how the CakePHP framework wraps phinx to run it inside its own CLI environment? 

Here's an example of using the Migrate command to add some extra features: https://github.com/cakephp/migrations/blob/master/src/Command/Migrate.php#L59

sabl0r:
I use this to simply run the phinx migrate command from my own CLI app.

```php
<?php
protected function execute(InputInterface $input, OutputInterface $output) {

	$arguments = [
		'command' => 'migrate',
		'-c' => PATH.'res/database/phinx.php',
		'-e' => 'default',
		'-vvv' => true
	];

	return (new PhinxApplication())->find('migrate')->run(new ArrayInput($arguments), $output);

}

iamdann:
Thanks @sabl0r

PR #1629, opened 2019-10-23
Add block on executing Phinx inside another symfony console app

MasterOdin:
Adds documentation per @dereuromark's request in #1628. Uses a combination of https://github.com/cakephp/phinx/issues/1365#issuecomment-396245197 and https://symfony.com/doc/current/console/calling_commands.html to explain how to do it so that it's better highlighted.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -306,6 +306,51 @@ and to rollback use `<http://localhost:8000/rollback>`__.
         or other another dynamic option, set ``$_SERVER['PHINX_DBNAME']`` before
         running commands. Available options are documented in the Configuration page.
 
+Wrapping Phinx in another Symfony Console Application
+-----------------------------------------------------
+
+Phinx can be wrapped and run as part of a separate Symfony console application. This
+may be desirable to present a unified interface to the user for all aspects of your
+application, or because you wish to run multiple Phinx commands. While you could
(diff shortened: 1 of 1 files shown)


docs/commands.rst
```diff
@@ -306,6 +306,51 @@ and to rollback use `<http://localhost:8000/rollback>`__.
...
+-----------------------------------------------------
+
+Phinx can be wrapped and run as part of a separate Symfony console application. This
+may be desirable to present a unified interface to the user for all aspects of your
+application, or because you wish to run multiple Phinx commands. While you could
+run the commands through ``exec`` or use the above ``Phinx\Wrapper\TextWrapper``,
+though this makes it hard to deal with the return code and output in a similar fashion
+as your application.
+
+Luckily, Symfony makes doing this sort of "meta" command straight-forward:
```
dereuromark: Does the final : needs to be :: ?
I am not too familiar with this rst stuff, but I thought this is needed to quote this char.

MasterOdin: You use the `::` to make something a [Literal Block](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#literal-blocks), which is handled by the `::` on line 321 (`.. code-block:: php`). Here I'm just using the `:` as a `:` to signify the connected content is coming on the next bit.

PR #1631, opened 2019-10-23
feat: add validation in PostgresAdapter for timestamp column limit

smmd:
Maybe this is not needed it since PostgreSQL reduces precision to the maximum allowed if you try a greater number, but since this was a topic in https://github.com/cakephp/phinx/pull/1289 I decided to open this PR


dereuromark:
Can anyone give feedback on this?

othercorey:
6 is the correct maximum for postgres timestamp types.

However, I would prefer the default postgresql behavior here. If it throws an error, that's good. If not, we should either throw an error in phinx or ignore it.

PR #1642, opened 2019-11-22
Introduce new command to list migration template creation aliases.

rquadling:
For those using migration template generators, the use of aliases provides a nice way to keep the names short, rather than providing full class names.

But the list of aliases and what they are aliasing is only available within the config file. Which is fine. 

This command simply lists the aliases and the aliased class.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -237,7 +237,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      */
     public function getAliases()
     {
-        return !empty($this->values['aliases']) ? $this->values['aliases'] : null;
+        return !empty($this->values['aliases']) ? $this->values['aliases'] : [];
     }
 
     /**
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -235,6 +235,14 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * {@inheritdoc}
      */
+    public function getAliases()
+    {
+        return !empty($this->values['aliases']) ? $this->values['aliases'] : null;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getConfigFilePath()
     {
         return $this->configFilePath;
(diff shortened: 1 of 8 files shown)


src/Phinx/Config/ConfigInterface.php
```diff
@@ -81,6 +81,13 @@ public function getDefaultEnvironment();
      */
     public function getAlias($alias);
 
+    /**
+     * Get all the aliased values.
+     *
+     * @return array|null
```
dereuromark: Why not just always array? would be cleaner IMO.

rquadling: Fair enough.

src/Phinx/Config/Config.php
```diff
@@ -232,6 +232,14 @@ public function getAlias($alias)
         return !empty($this->values['aliases'][$alias]) ? $this->values['aliases'][$alias] : null;
     }
 
+    /**
+     * {@inheritdoc}
```
dereuromark: Can we use the proper `@inheritDoc` tag here instead? This one means partial overwrite.

src/Phinx/Config/ConfigInterface.php
```diff
@@ -81,6 +81,13 @@ public function getDefaultEnvironment();
      */
     public function getAlias($alias);
 
+    /**
+     * Get all the aliased values.
+     *
+     * @return array
```
dereuromark: Technically, those are `string[]` here.

dereuromark:
Looks good :)

src/Phinx/Config/Config.php
rquadling: Ah. My bad.

src/Phinx/Config/ConfigInterface.php
rquadling: Yep!

src/Phinx/Config/Config.php
rquadling: Hmmm. I'm just following the existing pattern. If this is wrong, then a separate issue should be raised to fix this.

PR #1644, opened 2019-11-23
Upgrade to Symfony 5

tams:
With Symfony 5.0.0 out, added 5.x dependencies.

Greetings from #SymfonyHackday

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -147,7 +147,7 @@ class Create extends AbstractCommand
      * @param \Symfony\Component\Console\Output\OutputInterface $output
      * @throws \RuntimeException
      * @throws \InvalidArgumentException
-     * @return void
+     * @return int 0 on success
      */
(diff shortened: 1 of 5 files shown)


src/Phinx/Console/Command/Create.php
```diff
@@ -311,5 +311,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         $output->writeln('<info>created</info> ' . str_replace(getcwd() . DIRECTORY_SEPARATOR, '', $filePath));
+
+        return 0;
```
dereuromark: We have class constants for those afaik

tams: When looking for a convention in this namespace, I found integers for existing `execute()` default return codes.

In the scope of the version bump, it OK for me to keep following that pattern?

dereuromark:
https://github.com/cakephp/phinx/releases/tag/0.11.2

PR #1653, opened 2019-12-12
remove version

dereuromark:
We need to remove this version thingie
Without tooling/automation releasing otherwise is always by design error prone


dereuromark:
Should we remove the 
Test\Phinx\Console\Command\ListTest::testVersionInfo
part? Or read the current version somehow differently?

garas:
It could read and find itself in `vendor/composer/installed.json`.

But overall version doesn't add much value. I'd vote to remove any version related code.

`composer info robmorgan/phinx` is enough to find version for troubleshooting problems.

dereuromark:
@garas Tests are green now, good to go?

PR #1658, opened 2019-12-16
Update travis to test against 7.4 stable

MasterOdin:
PHP 7.4 got released on Nov 28, and so it's no longer necessary to test against the snapshot on Travis.

The next version (8.0) is currently not possible to use with PHPUnit so cannot add `nightly` or `8.0snapshot` for future testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - 7.4snapshot
+  - 7.4
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ matrix:
       env: PHPSTAN=1 DEFAULT=0
 
 before_install:
-  - if [[ ${TRAVIS_PHP_VERSION} != "7.4snapshot" ]]; then phpenv config-rm xdebug.ini; fi
+  - phpenv config-rm xdebug.ini
```
dereuromark: 7.4 doesnt have xdebug enabled, does it?

MasterOdin: Snapshot and nightly builds on Travis do not have any third-party packages, like xdebug. The stable build does include a number of them, like xdebug.

PR #1659, opened 2019-12-16
Add save() call after removeIndex in docs

### LINKED ISSUE #1391: removeIndex needs save(), but...
removeIndex() needs save(), but it is written in the document that "There is no need to call the save() method when using removeIndex(). The index will be removed immediately."
And it is called save() for tests of removeIndex().

MasterOdin:
Closes #1391

Adds `save()` call after calling `removeIndex` and `removeIndexByName` as it's necessary to do that now.

It should be noted that on https://book.cakephp.org/4/en/phinx/migrations.html#working-with-indexes, they already had the `save()` call, which makes me wonder where it's pulling its sources from for its docs as I could not find any branch in the repo that had this.

### DIFF
--- a/docs/migrations.rst
+++ b/docs/migrations.rst
@@ -1305,10 +1305,12 @@ call this method for each index.
             public function up()
             {
                 $table = $this->table('users');
-                $table->removeIndex(['email']);
+                $table->removeIndex(['email'])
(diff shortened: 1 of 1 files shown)


dereuromark:
Looks good!

PR #1660, opened 2019-12-16
Update version on phinx.org

MasterOdin:
phinx.org currently shows the latest version as being 0.10.8. This updates that version per the step listed in the wiki page https://github.com/cakephp/phinx/wiki/Releasing-a-new-version-of-Phinx.

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,7 +67,7 @@
       <div class="col-xs-12 col-sm-8 col-md-7">
       <h1>Phinx</h1>
       <h2>PHP Database Migrations For Everyone</h2>
-      <p><a type="button" class="btn btn-primary" href="#installModal" role="button" data-toggle="modal" onclick="_gaq.push(['_trackEvent', 'Hero Unit actions', 'Install', 'Install v0.10.8']);">INSTALL NOW v0.10.8</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -67,7 +67,7 @@
       <div class="col-xs-12 col-sm-8 col-md-7">
       <h1>Phinx</h1>
       <h2>PHP Database Migrations For Everyone</h2>
-      <p><a type="button" class="btn btn-primary" href="#installModal" role="button" data-toggle="modal" onclick="_gaq.push(['_trackEvent', 'Hero Unit actions', 'Install', 'Install v0.10.8']);">INSTALL NOW v0.10.8</a>
+      <p><a type="button" class="btn btn-primary" href="#installModal" role="button" data-toggle="modal" onclick="_gaq.push(['_trackEvent', 'Hero Unit actions', 'Install', 'Install v0.11.3']);">INSTALL NOW v0.11.3</a>
```
dereuromark: Should we only mention the current major? Avoids manual fix of this file all the time.

MasterOdin: Personally, I'd say create a bash script that takes as parameter the version to set that uses regex to update these two spots. This then could be run on tagged releases on Travis and use Travis to push the change to GH on build.

dereuromark: For now we dont have any such thing, until then I would prefer we keep human error to a minimum.

MasterOdin: Fair enough, remove bugfix version.

PR #1669, opened 2019-12-18
Print version order when running migrate

MasterOdin:
When running `rollback`, it prints out `ordering by [creation|execution] time` while migrate did not. This adds that print line to migrate to keep them consistent.

Added tests for when no order is provided (and defaults to creation) and when an order is provided.

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -105,7 +105,7 @@ EOT
         }
 
         $versionOrder = $this->getConfig()->getVersionOrder();
-        $output->writeln('<info>ordering by </info>' . $versionOrder . " time");
+        $output->writeln('<info>ordering by</info>'  . $versionOrder . ' time');
 
         if ($fake) {
             $output->writeln('<comment>warning</comment> performing fake migrations');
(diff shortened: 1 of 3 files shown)


src/Phinx/Console/Command/Migrate.php
```diff
@@ -104,6 +104,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $output->writeln('<info>using table suffix</info> ' . $envOptions['table_suffix']);
         }
 
+        $versionOrder = $this->getConfig()->getVersionOrder();
+        $output->writeln('<info>ordering by </info>' . $versionOrder . " time");
```
dereuromark: why /info closing before the rest of the line?

MasterOdin: This is a copy of what is going on in the rollback command which has had that since bba44b381fef9756c1d2f5a6c65ddc6757e7f6bb. I can update both to have a different format if so desired.

dereuromark:
Nah but there is now small conflict

src/Phinx/Console/Command/Migrate.php
dereuromark: I think I would have expected more like 

    <info>ordering by</info> ...

The whitespace after the closing tag.
But no big deal.

MasterOdin: Fixed in both commands.

MasterOdin:
Addressed merge conflict and spacing for the end info tag in both migrate and rollback commands.

PR #1670, opened 2019-12-18
Standardize exit codes to class constants

### LINKED ISSUE #1623: Standarize exit codes to constants across console commands
From discussion in #1622.

Right now, all console commands (with the slight exception of error codes for status) use magic integers for returns of exit codes. Phinx should define constants under a class to have a central place for where these constants are defined. My vote would be in the [`AbstractCommand`](https://github.com/cakephp/phinx/blob/master/src/Phinx/Console/Command/AbstractCommand.php) base class, and would include moving the two existing commands in the [`Manager`](https://github.com/cakephp/phinx/blob/master/src/Phinx/Migration/Manager.php) to there as well:
https://github.com/cakephp/phinx/blob/6b1c0edb507bb5b3ffb5056015a75b32052b6489/src/Phinx/Migration/Manager.php#L76-L84

Finally, there should be a consensus on when a command should throw an exception and when to print `<error>` and return an exit code and it seems somewhat inconsistent at this point with some potentially only throwing exceptions (`create`, `seed:create`), some doing a mix (`rollback`) and some only doing error codes(`migrate`).

MasterOdin:
Closes #1623

Currently, the exit codes were magic returned numbers within each command. This creates a class constant in AbstractCommand to use instead over all commands. I also moved the two status exit codes from the Manager to AbstractCommand as well to make the manager a bit more agnostic to how the console is working to make it potentially easier for someone to extend if they wanted to.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -53,6 +53,26 @@ abstract class AbstractCommand extends Command
     protected $manager;
 
     /**
+     * @var int
+     */
+    const EXIT_SUCCESS = 0;
+
+    /**
+     * @var int
+     */
+    const EXIT_ERROR = 1;
+
+    /**
+     * @var int
+     */
+    const EXIT_STATUS_MISSING = 2;
+
+    /**
+     * @var int
+     */
+    const EXIT_STATUS_DOWN = 3;
+
+    /**
      * {@inheritDoc}
      *
      * @return void
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -53,24 +53,29 @@ abstract class AbstractCommand extends Command
     protected $manager;
 
     /**
+     * Exit code for when command executes successfully
      * @var int
      */
-    const EXIT_SUCCESS = 0;
+    const CODE_SUCCESS = 0;
 
     /**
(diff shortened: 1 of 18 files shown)


src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -52,6 +52,26 @@ abstract class AbstractCommand extends Command
      */
     protected $manager;
 
+    /**
+     * @var int
+     */
+    const EXIT_SUCCESS = 0;
```
dereuromark: Should we use EXIT_ or CODE_? Refs https://github.com/cakephp/migrations/blob/master/src/Command/BaseCommand.php

Technically, it is a (status) code to be returned into CLI when existing, right?

CODE_SUCCESS
CODE_ERROR
CODE_MISSING
CODE_DOWN

?

dereuromark: And better to explain them a bit maybe in the docblock

MasterOdin: Ah, I did not realize there was an existing standard in Cake, I'll update it to:
* CODE_SUCCESS
* CODE_ERROR
* CODE_STATUS_MISSING
* CODE_STATUS_DOWN

dereuromark:
Are those fails related?

MasterOdin:
@dereuromark yup, fixed.

PR #1672, opened 2019-12-19
Switch docs builds to use CakePHP project tooling

markstory:
This moves phinx over to the same documentation tooling as we use for other projects like chronos and the various CakePHP plugins. This change depends on additional configuration and deployment changes.


docs/conf.py
```diff
@@ -3,240 +3,44 @@
...
-# Add any paths that contain custom themes here, relative to this directory.
-#html_theme_path = []
-
-# The name for this set of Sphinx documents.  If None, it defaults to
-# "<project> v<release> documentation".
-#html_title = None
-
-# A shorter title for the navigation bar.  Default is the same as html_title.
-#html_short_title = None
+release = '0.11.x'
```
othercorey: Aren't we officially 0.next now?

PR #1673, opened 2019-12-22
Add docs translations from book.cakephp.org

markstory:
Import the docs translations as this repository will soon be the only
copy of the documentation as we've resolved the hosting challenges that
lead to a fork in the first place.


docs/en/conf.py
```diff
@@ -1,46 +1,9 @@
...
-#
-# Phinx documentation build configuration file, created by
-# sphinx-quickstart on Thu Jun 14 17:39:42 2012.
-#
+import sys, os
 
-# Import the base theme configuration
-from cakephpsphinx.config.all import *
+# Append the top level directory of the docs, so we can import from the config dir.
+sys.path.insert(0, os.path.abspath('..'))
```
othercorey: Can you update this to what I did in cakephp/docs? The new docutils package has issues when the folder name is the same name as a language module that it can't find.  We shouldn't need this file and just pass `-D language=en` at the command line, but this works for now.

https://github.com/cakephp/docs/blob/4.x/en/conf.py

markstory: Sure.

Issue #1674, opened 2019-12-22
PostgresAdapter::getColumns does not handle postgis columns well


MasterOdin:
The `getColumns` function for the PostgresAdapter does not handle geometric or geographic types very well, missing in capturing the defined shape, as well as set coordinate system. This is due to the fact that this information is not set in the `information_schema.columns` table, but rather Postgis defines two secondary tables `geometry_columns` and `geography_columns` tables. While cycling through each column, if the type is `geography` or `geometry`, query the above tables as appropriate (e.g. `SELECT * FROM public.geography_columns WHERE f_table_schema = %s AND f_table_name = %s`).

dereuromark:
I this still open?

MasterOdin:
Yes, though I will probably make a new issue that combines this one and #1506, as the geometry handling in phinx needs quite a bit of work that should probably be done all at the same time. Leaving this open until such an issue is made though.

PR #1675, opened 2019-12-22
Allow setting SRID for geometry types

MasterOdin:
Supersedes #1558 

Fixes the merge conflict and adds the missing type. Manually tested as working, but tests for this will come as part of #1674.


src/Phinx/Db/Table/Column.php
```diff
@@ -101,6 +101,11 @@ class Column
      */
     protected $encoding;
 
+    /**
+     * @var int|null
+     */
+    protected $srid = null;
```
dereuromark: The = null is not needed here

MasterOdin: Fixed

PR #1681, opened 2019-12-23
Run init method for migration/seeds after setAdapter

### LINKED ISSUE #1095: adapter not available in init()
i would like to do some common init, like escaping fields and table names, so `up` and `down` could reuse them. `protected function init() {}` seemed right place to do that.

however, `$this->getAdapter()` returns `null` at the time `init()` is invoked:

```php
class SuperCoolMigration extends AbstractMigration
{
    public function init()
    {
        $this->email_table = $this->getAdapter()->quoteColumnName('support_email');
        $this->body_table = $this->getAdapter()->quoteColumnName('support_email_body');
    }
```

MasterOdin:
Closes #1095 

Init was being run for migrations/seeds at the end of the __construct method. This limited how it could be used, especially if looking to do things with the adapter for a migration/seed.

The function calls for a migration though does change from:
* __construct
* init (as part of __construct)
* preFlightCheck
* setAdapter
* up/down/change

to:
* __construct
* preFlightCheck
* setAdapter
* init
* up/down/change

For seeds, the execution order goes from:
* __construct
* init
* setAdapter
* run

to:
* __construct
* setAdapter
* init
* run


docs/en/migrations.rst
```diff
@@ -155,6 +155,13 @@ The down method is automatically run by Phinx when you are migrating down and
 it detects the given migration has been executed in the past. You should use
 the down method to reverse/undo the transformations described in the up method.
 
+The Init Method
+~~~~~~~~~~~~~~~
+
+The init method is run by Phinx before the migration methods if it exists. This
```
dereuromark: init() and code fencing please.

dereuromark:
I wonder if we should put this behavior change into 0.next (0.12 release) coming soon
This way we can have some migration guide for this breaking change.

docs/en/migrations.rst
MasterOdin: Done

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -82,17 +82,6 @@ abstract class AbstractMigration implements MigrationInterface
...
-
-        $this->init();
-    }
-
-    /**
-     * Initialize method.
-     *
-     * @return void
-     */
-    protected function init()
```
dereuromark: Why are you removing the method from the abstract class and then needing method_exists() check?
Couldnt we keep this method here?

MasterOdin: I can either remove it here or add a `public function int()` to the AbstractInterface.php and change this from `protected` to `public` here (for both seed and migration). I'm open to either, but was following the convention of up/down/change not existing (and since `run()` in AbstractSeed was protected by a `function_exists()` call).

dereuromark: I see. Do others have any opinion here?

tests/Phinx/Migration/Manager/EnvironmentTest.php
```diff
@@ -411,4 +411,51 @@ public function testGettingInputObject()
         $inputObject = $this->environment->getInput();
         $this->assertInstanceOf('\Symfony\Component\Console\Input\InputInterface', $inputObject);
     }
+
+    public function testExecuteMigrationCallsInit()
+    {
+        // stub adapter
+        $adapterStub = $this->getMockBuilder('\Phinx\Db\Adapter\PdoAdapter')
```
saeideng: try to not use the mock in future 
this deprecated in new phpunit versions

MasterOdin: Err, I'm not sure what you mean and that `getMockBuilder` is used heavily in the PHPUnit 8.5 documentation? What would you prefer I replace this with?

dereuromark:
Are we ok with the change of order? Then we can merge.

tests/Phinx/Migration/Manager/EnvironmentTest.php
saeideng: mock will deprecated in future
no replacement for it 
but goods for now

MasterOdin: Could you point to a link for what is being deprecated? I do know that the `setMethods` function is being soft deprecated in 8.x branch, but this is the first I'm hearing that `getMockBuilder` (and mocks in general?) is being deprecated.

dereuromark: For now this is fine.

saeideng: yes looks only `setMethods` and `getMock` deprecated now

PR #1682, opened 2019-12-23
Merge branch 'master' into 0.next 

MasterOdin:
Necessary to merge #1681 against 0.next without having it include all of master's commits in the PR history and changeset.


.travis.yml
```diff
@@ -11,6 +11,7 @@ services:
 
 php:
   - 7.2
+  - 7.3
```
dereuromark: the min/max should suffice here

docs/config/all.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Phinx documentation build configuration file, created by
+# sphinx-quickstart on Thu Jun 14 17:39:42 2012.
+#
+
+# Import the base theme configuration
+from cakephpsphinx.config.all import *
+
+# The full version, including alpha/beta/rc tags.
+release = '0.11.x'
```
dereuromark: these might have to be fixed to 0.12 then

docs/en/install.rst
```diff
@@ -10,7 +10,7 @@ website for more information.
 
 .. note::
 
-    Phinx requires at least PHP 5.4 (or later).
+    Phinx requires at least PHP 5.6 (or later).
```
dereuromark: 7.2 now

.travis.yml
MasterOdin: I'm not sure I follow this? Travis uses this as the list of versions to test, not a range, and so removing 7.3 would just cause Travis to not test 7.3 (see https://travis-ci.org/MasterOdin/phinx/builds/628789340 as example).

docs/en/install.rst
MasterOdin: Done

docs/config/all.py
MasterOdin: Do you want me to change this or is this something that the cakephp folks will handle once 0.next is going to be merged into master?

.travis.yml
dereuromark: Yes, but there is no use case where min/max alone wouldnt suffice.

docs/config/all.py
dereuromark: If you can would be nice, so we dont forget when merging

composer.json
```diff
@@ -27,9 +27,7 @@
     "require": {
         "php": ">=7.2",
         "cakephp/collection": "^4.0",
-        "cakephp/core": "^4.0",
         "cakephp/database": "^4.0",
-        "cakephp/datasource": "^4.0",
```
saeideng: why these removed ?

MasterOdin: Neither are directly depended upon and are just dependencies of dependencies.

saeideng: fair enough 
yes , they don't use directly in phinx , so good to remove

docs/config/all.py
MasterOdin: Updated to 0.12.x

.travis.yml
MasterOdin: Removed 7.3

docs/config/all.py
```diff
@@ -0,0 +1,43 @@
...
+
+# The marketing display name for the book.
+version_name = ''
+
+# Project name shown in the black header bar
+project = 'Phinx'
+
+# Other versions that display in the version picker menu.
+version_list = [
+    {'name': '0.11', 'number': '/phinx/11', 'title': '0.11', 'current': True},
```
dereuromark: maybe we can make this version number more DRY in the future.

docs/config/all.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Languages available.
+languages = ['en', 'es', 'fr', 'ja']
+
+# The GitHub branch name for this version of the docs
+# for edit links to point at.
+branch = 'master'
+
+# Current version being built
+version = '0.11'
```
dereuromark: here is another one

PR #1684, opened 2019-12-24
Migrate to phpunit 8+ for testing

MasterOdin:
With the dropping of everything below PHP 7.2, there is now no reason not to migrate to using only PHPUnit 8 for testing beyond the deprecation of usage of certain methods. It also has the added benefit (for me) of allowing alternative code coverage utilities, like pcov to be used out of the box.

As part of this PR, I've resolved all deprecation notices for PHPUnit with the exception of two in ConfigTest which are testing private class properties. The options as I see them here would be to:
1. Add a getter for `values` in Config object.
2. Do not test that `values` property of Config object is empty.

### DIFF
--- a/tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
+++ b/tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
@@ -40,11 +40,12 @@ class ConfigDefaultEnvironmentTest extends AbstractConfigTest
 
     public function testGetDefaultEnvironmentOverridenByEnvButNotSet()
     {
-        $this->expectException(RuntimeException::class);
-        $this->expectExceptionMessage('The environment configuration (read from $PHINX_ENVIRONMENT) for \'conf-test\' is missing');
         // set dummy
         $dummyEnv = 'conf-test';
         putenv('PHINX_ENVIRONMENT=' . $dummyEnv);
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('The environment configuration (read from $PHINX_ENVIRONMENT) for \'conf-test\' is missing');
 
         try {
             $config = new Config([]);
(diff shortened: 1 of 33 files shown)


tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
```diff
@@ -37,12 +38,10 @@ public function testConfigReplacesTokensWithEnvVariables()
         $this->assertEquals('1234', $env['port']);
     }
 
-    /**
-     * @expectedException \RuntimeException
-     * @expectedExceptionMessage The environment configuration (read from $PHINX_ENVIRONMENT) for 'conf-test' is missing
-     */
     public function testGetDefaultEnvironmentOverridenByEnvButNotSet()
     {
+        $this->expectException(RuntimeException::class);
```
dereuromark: As a good practice/code-sniff we should move the expect* methods directly before the part that is supposed to throw it.
If we dont do that, other code can have side effects with similar exceptions cloaking it.
So if we move it inline, we better use the best of this.

Also, please add an extra newline before and after the expect* method blocks for readability.

dereuromark: I am actually writing a sniff for our company here (whitespace for readability), we could also add this one to the core sniffer repo if we want to.

tests/Phinx/Config/ConfigDefaultEnvironmentTest.php
```diff
@@ -37,16 +38,15 @@ public function testConfigReplacesTokensWithEnvVariables()
...
     public function testGetDefaultEnvironmentOverridenByEnvButNotSet()
     {
         // set dummy
         $dummyEnv = 'conf-test';
         putenv('PHINX_ENVIRONMENT=' . $dummyEnv);
 
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('The environment configuration (read from $PHINX_ENVIRONMENT) for \'conf-test\' is missing');
+
         try {
```
garas: This `try-catch` is better as `try-finally`, because in case exception is not thrown, environment wouldn't reset, there is also no need to manually rethrow exception.

dereuromark:
The move showed already a (positive) side effect of placement of those expectations.
A test fails here it seems.

PR #1687, opened 2019-12-30
SQLite - To create NOT NULL column without DEFAULT value

charlienux:
There is an error in **SQLiteAdapter** that creates the columns as NULL despite having indicated the field as NOT NULL. _Currently, the only way to create a NOT NULL field is to specify a DEFAULT value for that column._

With this change, it is allowed to create a NOT NULL column without indicating its DEFAULT value.

In order to allow for backward compatibility, this change will cause SQLiteAdapter continue to create a column as NULL by default when it is not specified whether it is null or not.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -1458,7 +1458,7 @@ PCRE_PATTERN;
 
         $default = $column->getDefault();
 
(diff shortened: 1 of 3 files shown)


charlienux:
The SQL statements generated when creating the columns are correct when **phinx migrate** is executed.
The test is not successful because the MockAdapter returns NULL when requesting getAdapter()->getAdapterType.
The condition in Table.php does not run because getAdapterType is not equal to **sqlite.**
What to do to successfully pass the tests?

dereuromark:
tests failing indicate that there is some bc issue here maybe.

PR #1689, opened 2020-01-02
Rename default_database setting to default_environment

### LINKED ISSUE #984: Get environment accessible from a migration
Conceptually, the key _default_database_ in phinx.yml file should be named _default_environment_. Also the environment value should be accessible from a migration, something like:

`$this->get_environment()`

I know that right now I can get environment (parameter **e**) if it is used when phinx is executed but if no environment parameter is passed, there is no way to know the **default_environment**.

MasterOdin:
Closes #984 (second part of it)

This PR renames the config setting `default_database` to `default_environment` which is more consistent with how it's used in the code and referenced throughout Phinx (e.g. `getDefaultEnvironment` method). For compatibility, `default_database` is still allowed, with an idea that it would be removed in the following major version (similar to how the deprecation/removal of the legacy pre-plan functions were handled from 0.10 to 0.11).


dereuromark:
Can we note all breaking changes in some Migration note?

dereuromark:
src/data/Phinx/phinx.php.dist

=> Can we move those to its own ROOT dir?
Having them inside src is a bit dirty as those are not psr4 autoload classes.

PR #1695, opened 2020-01-18
Improved Wrapper to support status command in json format

jackie-d:
With this simple PR we grant the TextWrapper status command the support for json output format
for ease of use.


jackie-d:
Any news about this?

dereuromark:
Usually, a test case along with the change would be nice.
This seems simple enough for now, though.

Issue #1702, opened 2020-02-05
Make dependency on symfony/yaml optional.


slepic:
Phinx has this dependency in composer.json:
`"symfony/yaml": "^3.4|^4.0|^5.0"`

This can be moved to require-dev (and suggest section) if `Phinx\Config\Config::fromYaml` check existance of the class `Symfony\Component\Yaml\Yaml`

Furhter in `Phinx\Console\Command\AbstractCommand::loadConfig` can autodetect parser from file extension and if unknown extension is provided, it defaults to yaml. It may eventually default to ie. php if symfony yaml component is not installed, but of course it can remain default but it would throw error if used as default without yaml component installed.
Note: If the default changes, it could be good to also look for `.yaml` not just `.yml`

It seem quite straight forward, except it has to be decided what should the behaviour be when using default but yaml component is not installed. And I leave that decision to you.
Other than that I am willing to make PR once we settle how it should be handled.

---------------------------------------

Oh and now I found an old issue with the same request
https://github.com/cakephp/phinx/issues/363
But it was never implemented and I am not sure if I understand the reasoning
> I'm closing this, Phinx still relies on the Yaml by default.

Does it mean you dont intend to drop yaml as required dependency, just not to break others' codes that rely on the default?
Just bump the major version number if you see it as problem, this package should have been on 1.0.0 long ago anyway. Unless you don't care for semver on which composer relies (in a way), in which case you need not care for BC at all.

--------------------------------------
PS: btw the same goes for ext-json which is not in the composer.json at all. This is not that much of an issue because json extension is bundled with PHP by default. But for the sake of correctness, it might be handled as optional dependency with checks for the extension being installed and disallow using json format otherwise. But this should have its own issue, I am only mentioning it because it came to my mind in association with the yaml.

dereuromark:
Do you want to make a PR here?

PR #1704, opened 2020-02-05
Mark ext-json as optional and check it exists for usage

MasterOdin:
A bottom point from #1702 and my comment from https://github.com/cakephp/phinx/pull/1491#issuecomment-580784114.

Adds `ext-json` as a suggested package to be installed for phinx to function if using a JSON configuration file. I also add a check that the function exists before using it so that the error message can be a bit nicer (instead of throwing a fatal error with a stacktrace).

If this is merged #1491 should definitely be closed, given that its proposal was making ext-json mandatory (as it was when that PR was opened, but is no longer).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "symfony/yaml": "^3.4|^4.0|^5.0"
     },
     "require-dev": {
+        "ext-json": "*",
         "phpunit/phpunit": ">=5.7,<8.0",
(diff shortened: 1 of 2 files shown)


dereuromark:
We could even add it into

    "require-dev"

section then (as it is needed for our tests).

MasterOdin:
@dereuromark Fair enough, done.

PR #1717, opened 2020-03-30
make symfony/yaml optional and check if it is installed for usage

### LINKED ISSUE #1702: Make dependency on symfony/yaml optional.

slepic:
resolves #1702

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,14 +29,14 @@
         "cakephp/collection": "^3.7",
         "cakephp/database": "^3.7",
         "symfony/console": "^3.4|^4.0|^5.0",
-        "symfony/config": "^3.4|^4.0|^5.0",
-        "symfony/yaml": "^3.4|^4.0|^5.0"
+        "symfony/config": "^3.4|^4.0|^5.0"
     },
     "require-dev": {
         "ext-json": "*",
         "phpunit/phpunit": ">=5.7,<8.0",
         "sebastian/comparator": ">=1.2.3",
-        "cakephp/cakephp-codesniffer": "^3.0"
+        "cakephp/cakephp-codesniffer": "^3.0",
+        "symfony/yaml": "^3.4|^4.0|^5.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -49,7 +49,8 @@
         }
     },
     "suggest": {
-        "ext-json": "Install if using JSON configuration format"
+        "ext-json": "Install if using JSON configuration format",
+        "symfony/yaml": "Install if using Yaml configuration format"
```
dereuromark: YAML

dereuromark:
Next time we should do this in a major (0.12 upcoming) release :)
https://github.com/cakephp/phinx/issues/1726

PR #1718, opened 2020-04-05
Use not null by default for sqlite column definitions

### LINKED ISSUE #1446: SQLite defaulting all columns to nullable
Version 0.10.6

When deploying to SQLlite, all columns (except identity) are defaulting to allowing null.  This behavior differs from all other adapters, so it should be corrected.  This issue seems to only occur when the default value of a column isn't set.

SQLite
`$def .= (!$column->isIdentity() && ($column->isNull() || is_null($default))) ? ' NULL' : ' NOT NULL';`

MySQL
`$def .= ($column->isNull() == false) ? ' NOT NULL' : ' NULL';`

MasterOdin:
Fixes #1446

This brings the SQLite adapter inline with all other adapters, making it so that all creates columns as `NOT NULL` by default. SQLite does have an annoying "bug" where a column can be `NOT NULL` without specifying a default on table creation, but cannot if appending it to an existing table. I've added a note to the docs about this

Either this PR or #1687 should be merged and the other should be closed.

The change to the MySQLAdapter is just cosmetic to make it match the other adapters.

### DIFF
--- a/docs/en/migrations.rst
+++ b/docs/en/migrations.rst
@@ -809,7 +809,7 @@ Option  Description
 limit   set maximum length for strings, also hints column types in adapters (see note below)
 length  alias for ``limit``
 default set default value or action
-null    allow ``NULL`` values (should not be used with primary keys!)
+null    allow ``NULL`` values, defaults to false (should not be used with primary keys!) (see note below)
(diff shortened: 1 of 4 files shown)


sitedyno:
#1872 :ping_pong:

PR #1719, opened 2020-04-05
Fix annotation to use right Helper class

MasterOdin:


dereuromark:
@MasterOdin Thanks four your outstanding help, man!

PR #1722, opened 2020-04-06
Use stable dependencies.

dereuromark:
soon then phpstan level 5


composer.json
```diff
@@ -59,9 +59,9 @@
         ],
         "cs-check": "phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP --ignore=/tests/Phinx/Config/_rootDirectories/,/tests/log/ src/ tests/ app/",
         "cs-fix": "phpcbf --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP --ignore=/tests/Phinx/Config/_rootDirectories/,/tests/log/ src/ tests/ app/",
+        "stan": "phpstan analyse src/",
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:^0.12 && mv composer.backup composer.json",
```
MasterOdin, line 63: Why not just add phpstan to the dev dependencies, much like codesniffer?

dereuromark: We could but then it might change actual dependendies for testing.

PR #1727, opened 2020-04-08
Do not throw exception for missing phinxlog for dry run

MasterOdin:
If doing `phinx migrate --dry-run` without having previously run phinx, it would throw an exception as the `phinxlog` table is to be created as part of the dry run, but since it is not, this query then throws with the table not existing.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -370,7 +370,17 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
                 throw new RuntimeException('Invalid version_order configuration option');
         }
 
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -370,7 +370,17 @@ public function getVersionLog()
                 throw new RuntimeException('Invalid version_order configuration option');
         }
 
-        $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY %s', $this->quoteTableName($this->getSchemaTableName()), $orderBy));
+        // This will throw an exception if doing a --dry-run without any migrations as phinxlog
+        // does not exist, so in that case, we can just expect to trivially return empty set
+        try {
+            $rows = $this->fetchAll(sprintf('SELECT * FROM %s ORDER BY %s', $this->quoteTableName($this->getSchemaTableName()), $orderBy));
```
saeideng, line 376: you could use `cakephp/database` query here

PR #1729, opened 2020-04-08
Improving handling of add column instructions in sqlite

### LINKED ISSUE #1611: SQLite error with addTimestamps()
With current implementation of `addTimestamps()` method, it is impossible to add timestamp columns for already created table in SQLite environment. As there must be default value for column...

Migration 1:
```php
$this->table('users')
    ->addColumn('name', 'string', ['limit' => 255])
    ->addColumn('email', 'string', ['limit' => 255])
    ->addColumn('password', 'char', ['limit' => 60])
    ->save();
```

Migration 2:
```php
$this->table('users')->addTimestamps()->update();
```

Which generates:
```
PDOException: SQLSTATE[HY000]: General error: 1 Cannot add a column with non-constant default in /srv/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php:168
```

MasterOdin:
SQLite has many limitations with the ALTER TABLE commands around dropping columns, altering them, etc. which have been covered in previous PRs. However, while SQLite does
support using ALTER TABLE to add columns, it comes with quite a bit of caveats which include not being able to define a NOT NULL without a default, cannot define a default that is a function (e.g. CURRENT_TIMESTAMP), etc. As such, to overcome these things which people probably would not expect coming from other DBs, this commit makes it so that add column commands, like the alter column and drop column commands, we create a temp table with our desired changes, then copy in all old data into.

Closes #1611 and removes the cop-out that was my original solution for #1718 (with something that works as expected).

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -814,7 +814,7 @@ PCRE_PATTERN;
         $selectColumns = array_map([$this, 'quoteColumnName'], $selectColumns);
         $writeColumns = array_map([$this, 'quoteColumnName'], $writeColumns);
 
-        if (!$found && $columnName !== false) {
+        if ($columnName && !$found) {
             throw new InvalidArgumentException(sprintf(
                 'The specified column doesn\'t exist: ' . $columnName
             ));
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -673,14 +673,30 @@ PCRE_PATTERN;
      */
     protected function getAddColumnInstructions(Table $table, Column $column)
     {
-        $alter = sprintf(
-            'ALTER TABLE %s ADD COLUMN %s %s',
-            $this->quoteTableName($table->getName()),
-            $this->quoteColumnName($column->getName()),
-            $this->getColumnSqlDefinition($column)
-        );
+        $tableName = $table->getName();
 
-        return new AlterInstructions([], [$alter]);
+        $instructions = $this->beginAlterByCopyTable($tableName);
+
+        $instructions->addPostStep(function ($state) use ($column) {
+            $sql = $state['createSQL'];
+            $sql = preg_replace(
+                "/\)$/",
+                sprintf(', %s %s$1)', $this->quoteColumnName($column->getName()), $this->getColumnSqlDefinition($column)),
+                $state['createSQL'],
+                1
+            );
+            $this->execute($sql);
+
+            return $state;
+        });
+
+        $instructions->addPostStep(function ($state) use ($tableName) {
+            $newState = $this->calculateNewTableColumns($tableName, false, false);
+
+            return $newState + $state;
+        });
+
+        return $this->copyAndDropTmpTable($instructions, $tableName);
     }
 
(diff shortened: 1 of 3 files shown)


tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1117,9 +1117,28 @@ public function testAlterTableColumnAdd()
...
+            ['name' => 'created_at', 'type' => 'timestamp', 'default' => 'CURRENT_TIMESTAMP', 'null' => false],
+            ['name' => 'updated_at', 'type' => 'timestamp', 'default' => null, 'null' => true],
+        ];
+
+        $this->assertEquals(count($expected), count($columns));
+        $columnCount = count($columns);
+        for ($i = 0; $i < $columnCount; $i++) {
+            $this->assertEquals($expected[$i]['name'], $columns[$i]->getName(), "Wrong name for {$expected[$i]['name']}");
+            $this->assertEquals($expected[$i]['type'], $columns[$i]->getType(), "Wrong type for {$expected[$i]['name']}");
+            $this->assertEquals($expected[$i]['default'], $columns[$i]->getDefault(), "Wrong default for {$expected[$i]['name']}");
```
dereuromark, line 1139: Can we use assertSame() please. The false positive comparison of null vs false etc can be quite a problem.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -798,7 +814,7 @@ protected function calculateNewTableColumns($tableName, $columnName, $newColumnN
         $selectColumns = array_map([$this, 'quoteColumnName'], $selectColumns);
         $writeColumns = array_map([$this, 'quoteColumnName'], $writeColumns);
 
-        if (!$found) {
+        if (!$found && !empty($columnName)) {
```
dereuromark, line 817: !empty() is not needed here.

MasterOdin: Changed to `!==` as I think that's what you mean by "non-cloaking"?

dereuromark: No, i was talking about the equivalent of !empty() but without cloaking. Fixed it for you.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -798,7 +814,7 @@ protected function calculateNewTableColumns($tableName, $columnName, $newColumnN
         $selectColumns = array_map([$this, 'quoteColumnName'], $selectColumns);
         $writeColumns = array_map([$this, 'quoteColumnName'], $writeColumns);
 
-        if (!$found) {
+        if (!$found && $columnName !== false) {
```
dereuromark, line 817: Please don't force push. It is not needed, and just reverted my fix.

dereuromark: if (!$found && $columnName) {

MasterOdin: Done.

I've been force-pushing, so as to make the history a bit cleaner as I've had to look back a handful of times on commits and getting pieces of a larger changeset has been tedious for me.

I will stop combining my PR commits though going forward though if so desired.

PR #1730, opened 2020-04-09
Support setting SRID for geometry columns for MySQL 8+

### LINKED ISSUE #1677: Support setting SRID for MySQL 8
Per the [docs](https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html), you can now do the following:
```
CREATE TABLE geom (
    p POINT SRID 0,
    g GEOMETRY NOT NULL SRID 4326
);
```
Where now those columns have a set SRID. This was not available prior to 8.0.

MasterOdin:
Closes #1677

MySQL 8 added support for setting the SRID attribute of a column. This makes it so that you could only insert geometries of that SRID into that column.

I've set-up a travis build with MySQL 8 to show the tests passing (though it shows a lot of failing other MySQL tests on 8): https://travis-ci.org/github/MasterOdin/phinx/builds/672758069

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1163,7 +1163,8 @@ class MysqlAdapter extends PdoAdapter
         $def .= (!$column->isSigned() && isset($this->signedColumnTypes[$column->getType()])) ? ' unsigned' : '';
         $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
 
-        if (version_compare($this->getAttribute(\PDO::ATTR_SERVER_VERSION), '8') > -1
+        if (
+            version_compare($this->getAttribute(\PDO::ATTR_SERVER_VERSION), '8') > -1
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1162,6 +1162,15 @@ protected function getColumnSqlDefinition(Column $column)
         $def .= $column->getCollation() ? ' COLLATE ' . $column->getCollation() : '';
         $def .= (!$column->isSigned() && isset($this->signedColumnTypes[$column->getType()])) ? ' unsigned' : '';
         $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
+
+        if (
+            version_compare($this->getAttribute(\PDO::ATTR_SERVER_VERSION), '8') > -1
```
MasterOdin, line 1167: Stickler-CI putting this on a newline is probably correct (as it follows PSR-12), but I had purposely not so that phpcs would pass. Not sure how to proceed to get a green build.

dereuromark:
Do you want to try to get travis green here?

Issue #1732, opened 2020-04-10
Support binaryuuid out of the box


dereuromark:
As per https://github.com/cakephp/migrations/issues/331#issuecomment-610698332

Currently, this is needed:

	$table->addColumn('id', 'binary', [
		'limit' => 16,
		'default' => null,
		'null' => false,
	]);

But this should be possible/detected for primary and foreign keys:

        $this->table('uuid16_test_rows', ['id' => false, 'primary_key' => ['id']])
            ->addColumn('id', 'binaryuuid', [
                'default' => null,
                'limit' => null,
                'null' => false,
            ])
            ->addColumn('foreign_key_without_index', 'binaryuuid', [
                'default' => null,
                'limit' => null,
                'null' => true,
            ])
            ->addColumn('foreign_key_with_index', 'binaryuuid', [
                'default' => null,
                'limit' => null,
                'null' => true,
            ])
            ->create();

in sync with uuid.

PR #1734, opened 2020-04-10
Start adding binaryuuid support.

dereuromark:
First stab at https://github.com/cakephp/phinx/issues/1732

I might need some help, as I am not too familiar with the lower levels here.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -32,6 +32,7 @@ class PostgresAdapter extends PdoAdapter
         self::PHINX_TYPE_INET,
         self::PHINX_TYPE_MACADDR,
         self::PHINX_TYPE_INTERVAL,
+        self::PHINX_TYPE_BINARYUUID,
     ];
 
     /**
@@ -979,6 +980,8 @@ class PostgresAdapter extends PdoAdapter
                 return ['name' => 'real'];
             case static::PHINX_TYPE_DATETIME:
                 return ['name' => 'timestamp'];
+            case static::PHINX_TYPE_BINARYUUID:
+                return ['name' => 'uuid'];
             case static::PHINX_TYPE_BLOB:
             case static::PHINX_TYPE_BINARY:
                 return ['name' => 'bytea'];
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -56,6 +56,7 @@ class SQLiteAdapter extends PdoAdapter
         self::PHINX_TYPE_TEXT => 'text',
         self::PHINX_TYPE_TIME => 'time_text',
         self::PHINX_TYPE_UUID => 'uuid_text',
+        self::PHINX_TYPE_BINARYUUID => 'binary_blob',
         self::PHINX_TYPE_TIMESTAMP => 'timestamp_text',
         self::PHINX_TYPE_VARBINARY => 'varbinary_blob',
     ];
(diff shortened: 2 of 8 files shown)


dereuromark:
This now works for MySQL already:
```
CREATE TABLE `uuid16_test_rows` (
  `id` binary(16) NOT NULL,
  `name` varchar(100) COLLATE utf8_unicode_ci NOT NULL DEFAULT '',
  `foreign_key_without_index` binary(16) DEFAULT NULL,
  `foreign_key_with_index` binary(16) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;

ALTER TABLE `uuid16_test_rows`
  ADD KEY `foreign_key_with_index` (`foreign_key_with_index`);
COMMIT;
```

dereuromark:
Tests are green so far except SQLite

SQlite has some issues still
"General error: 1 near "`user_id`": syntax error"

uuid

> CREATE TABLE `tmp_ztable` (`id` UUID_TEXT NOT NULL, PRIMARY KEY (`id`), `user_id` INTEGER NOT NULL)

binaryuuid

> CREATE TABLE `tmp_ztable` (`id` BINARY_BLOB NOT NULL, PRIMARY KEY (`id`), `user_id` INTEGER NOT NULL)

dereuromark:
The way SQLite creates those primary keys seems to be the issue
```
CREATE TABLE tmp_ztable
( id BINARY_BLOB NOT NULL PRIMARY KEY,
  user_id INTEGER NOT NULL
);
```
would be better

but that is a different topic and not directly related to the changes here.
So all good it seems.

I tested it also with CakePHP and migrations plugin, which now works correctly for all 3 db types: https://github.com/cakephp/migrations/issues/331

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -443,6 +443,10 @@ public function getColumns($tableName)
                 $column->setPrecision($columnInfo['numeric_precision']);
             }
 
+            if ($column->getType() === static::PHINX_TYPE_BINARY && $column->getLimit() === 16) {
+                $column->setType(static::PHINX_TYPE_BINARYUUID);
+            }
```
MasterOdin, line 448: Is it possible to enter this if statement?

dereuromark: Based on your findings it might not be necessary.

MasterOdin:
For SQLServer, `binaryuuid` would probably map to the same type as `static::PHINX_TYPE_UUID`, namely `uniqueidentifier`, though like the postgresql adapter, you would not be able to do a comparison against `static::PHINX_TYPE_BINARYUUID` on saved columns in a table.

src/Phinx/Db/Adapter/PostgresAdapter.php
markstory: Binary UUIDs in postgres would use the standard `UUID` type. Not a binary column.

dereuromark: That means we could safely remove this?

othercorey: I believe so. You should always use the uuid type with postgres.

dereuromark: Done

dereuromark:
@MasterOdin For SqlServer it seems like UNIQUEIDENTIFIER could be the one to be used.
Do you want to look into this?
I think for the main DBs this support is sufficient for now.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -56,6 +56,7 @@ class SQLiteAdapter extends PdoAdapter
         self::PHINX_TYPE_TEXT => 'text',
         self::PHINX_TYPE_TIME => 'time_text',
         self::PHINX_TYPE_UUID => 'uuid_text',
+        self::PHINX_TYPE_BINARYUUID => 'binary_blob',
```
dereuromark, line 59: That seems to be the problem/wrong. See https://github.com/cakephp/migrations/issues/660

dereuromark: uuid_blob is the correct one it seems

PR #1738, opened 2020-04-11
Properly ignore folders at root of project

MasterOdin:


.gitignore
```diff
@@ -22,8 +22,8 @@ composer.lock
 tests/log
 
 # ignore core development artifacts
-./db
-./migrations
+/db
+/migrations
```
dereuromark, line 26: I prefer a trailing slash to clarify folders vs files :)
But this is also fine for now

MasterOdin: Fixed, and added them to a handful of others, though I think I missed a handful of other folders.

dereuromark:
I think I broke a test with my last merge

```
There was 1 failure:

1) Test\Phinx\Db\Adapter\MysqlAdapterTest::testsHasForeignKeyWithSchemaDotTableName
Failed asserting that false is true.
```

MasterOdin:
Looks like Travis is passing, not sure what might be causing that test failure.

dereuromark:
Maybe my local DB or mysql version.

Issue #267, opened 2014-07-07
Command [Create] doesn't work on single-letter words in camel case


hmp5bs:
eg)  t_foo

```
$ phinx create TFoo

  [InvalidArgumentException]
  The migration class name "TFoo" is invalid. Please use CamelCase format.
```

jasonwatt:
Ran into the same issue. isValidPhinxClassName would need to updated

```
public static function isValidPhinxClassName($className)
    {
        return (bool) preg_match('/^([A-Z][A-Za-z0-9]+)+$/', $className);
    }
```

73sl4:
@jasonwatt Regex would be
```
^([A-Z][a-z0-9]*)+$
```
isValidPhinxClassName would change to
```
public static function isValidPhinxClassName($className)
{
    return (bool) preg_match('/^([A-Z][a-z0-9]*)+$/', $className);
}
```

tuannh99:
The regex should be:
```php
^([A-Z][a-z0-9]*)+$
```

Then isValidPhinxClassName would be:
```php
{
    return (bool) preg_match('/^([A-Z][a-z0-9]*)+$/', $className);
}
```

73sl4:
@tuannh99 This regex will match capital letter word also.

We can use this regex but I'm not sure.
```
^([A-Z]{1,2}[a-z0-9]+)+$
```

tuannh99:
@73sl4 Re-look my regex, I updated it a bit.
If nobody has objection on that, I will make write several unit test cases as well as pull-request for that.

73sl4:
@tuannh99 Same issue with this regex. I even wrote the same regex before but then realised that it won't work. You can use `http://regexr.com/` to test the regex.

tuannh99:
@73sl4 : The above regex works with:
- TFoo
- CreateTableArticles
...

What is the case you said it failed?

rquadling:
Some thoughts/suggestions/ideas.


One issue is capitalised words (SQL, CSV, HTML, etc.) become single letter snake case.

So many different patterns would be useful. We have recently started using Jira for our project tracking. Creating a migration that included the ticket number.

We would want to be able to have a migration named something like ``DT396_MigrateSomething``. Ideally, leaving the class as that, rather than changing it all would be ideal.

Now considering class names are validated by a static method, MAYBE (and this is just a random idea), who about allowing the method to be injected/overwritten/defined elsewhere/somehow.

[...]

PR #1744, opened 2020-04-11
Relax migration & seed class name restrictions

Roy-Orbison:
Allow single letter words and initialisms as per issue #267.

### DIFF
--- a/src/Phinx/Util/Util.php
+++ b/src/Phinx/Util/Util.php
@@ -21,12 +21,17 @@ class Util
     /**
      * @var string
      */
-    protected const MIGRATION_FILE_NAME_PATTERN = '/^\d+_([\w_]+).php$/i';
+    protected const MIGRATION_FILE_NAME_PATTERN = '/^\d+_([a-z][a-z\d]*(?:_[a-z\d]+)*)\.php$/i';
 
     /**
      * @var string
      */
-    protected const SEED_FILE_NAME_PATTERN = '/^([A-Z][a-z0-9]+).php$/i';
+    protected const SEED_FILE_NAME_PATTERN = '/^([a-z][a-z\d]*)\.php$/i';
+
+    /**
+     * @var string
+     */
+    protected const CLASS_NAME_PATTERN = '/^(?:[A-Z][a-z\d]*)+$/';
 
     /**
      * Gets the current timestamp string, in UTC.
@@ -56,11 +61,12 @@ class Util
         }
 
         // filter the files to only get the ones that match our naming scheme
-        $phpFiles = glob($path . DIRECTORY_SEPARATOR . '*.php');
+        $phpFiles = static::getFiles($path);
 
         foreach ($phpFiles as $filePath) {
-            if (preg_match('/([0-9]+)_([_a-z0-9]*).php/', basename($filePath))) {
-                $classNames[] = static::mapFileNameToClassName(basename($filePath));
+            $fileName = basename($filePath);
+            if (preg_match(static::MIGRATION_FILE_NAME_PATTERN, $fileName)) {
+                $classNames[] = static::mapFileNameToClassName($fileName);
             }
         }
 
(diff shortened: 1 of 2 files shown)


dereuromark:
If you allow string to go into getFiles(), please document it as such:

     * @param string|string[] $paths Path or array of paths to get PHP files.
     *
     * @return string[]
     */
    public static function getFiles($paths)

Roy-Orbison:
The force pushes are because I'm abusing the CI tools for QA.

I can't see the log for why [only PHP 7.4 fails](https://travis-ci.org/github/cakephp/phinx/jobs/673819766), does it take time to be accessible?

dereuromark:
Travis has a maintenance issue and backlogs, so we just have to be a bit patient today.

dereuromark:
```
 ------ ----------------------------------------------------------------------- 
  Line   Phinx/Util/Util.php                                                    
 ------ ----------------------------------------------------------------------- 
  86     Method Phinx\Util\Util::getVersionFromFileName() should return string  
         but return statement is missing.                                       
 ------ ----------------------------------------------------------------------- 

 [ERROR] Found 1 error      
```

We might want to throw an exception if the version cannot be retrieved.

Roy-Orbison:
I've set it to explicitly return `null` as this is the current behaviour but without the possibility of triggering a Notice error when no match occurs (I don't think it does in practice). I think an Exception is outside the scope of this change because it would be an API break. Up to you.

src/Phinx/Util/Util.php
```diff
@@ -82,10 +82,13 @@ public static function getExistingMigrationClassNames($path)
      */
     public static function getVersionFromFileName($fileName)
     {
+        $version = null;
```
dereuromark, line 85: This violates the return Type. The exception is cleaner.

dereuromark: If you look into the code the result is used directly downstream. So it would break anyway. A meaningful error message is more helpful here anyway. 
This is not a bc break then.

Roy-Orbison: The return type is already violated because it currently returns null when no match occurs. `$match` gets set to an empty array, thus the returned `$match[0]` does not exist, is coerced to null, and triggers a notice level error. This should not block this PR as compatibility is maintained. It should be marked nullable, i.e.:
```php
public static function getVersionFromFileName($fileName): ?string 
```
but that breaks compatibility for PHP <7.1.

I still believe changing the behaviour of this method is beyond the scope of this change.

Roy-Orbison: It's a public method, you cannot guarantee no bc break because application code may expect `null` without wrapping in `try` + `catch` blocks.

Roy-Orbison: As far as I can see. both internal uses of the method operate on file lists where the names have already been validated, hence it not being a problem in real-world use. I understand that this makes the code better, but I see no practical benefit untless the migration filename pattern changes.

dereuromark: The code expects string. Lets keep it simple.

dereuromark: > changing the behaviour of this method is beyond the scope of this change

You are already changing the behavior of phinx as a whole, making the class name more relaxed. At the same time we want to keep the version strictness (it currently had implicitly downstream). The exception at this point is warranted and currently needed due to your refactorings and phpstan noticing now.
If any library would build on this low level class then they can easily adjust their code. but it is very likely they do not need to as this use case is not normal procedure and the invalid file name would have already blown up somewhere else before anyway.

Roy-Orbison:
>
> You are already changing the behavior of phinx as a whole
>
I disagree; no process has really changed. The only effects are:

1. Dropping an arguably arbitrary restriction which reduces chances but
does not prevent a certain exception from one command, `phinx create …`.
2. Making some new migration filenames slightly more readable.

All the logic of reading, checking, and applying migrations is exactly as
it was.

> we want to keep the version strictness
>
This was not relaxed.

> The exception at this point is warranted and currently needed due to your
> refactorings and phpstan noticing now.
>
Okay, so add a `throw` where and how you think it should be. This is a
limitation of phpstan not a failure to resolve the current issue. This is
creep that would be better resolved here by setting `$version = '';` so
return type isn't "broken", then making your desired change elsewhere. I
was reluctant to do that because that's a break to me.

> If any library would build on this low level class then they can easily
> adjust their code.
>
Yes and will they be notified that they have to?

> the invalid file name would have already blown up somewhere else before
> anyway.
>
This was my point earlier, no change in functionality.

Roy-Orbison:
I have no problem with your proposed change, only that it should be a fix unto itself, not bundled in with this enhancement. I have removed the offending hunk from this PR.

PR #1753, opened 2020-04-20
Remove cakephp/collection dependency

MasterOdin:
This rewrites all usage of `cakephp/collection` to use built-in PHP structures, removing it as a dependency.


dereuromark:
What is the motivation, the reason for this? Other than pulling in a small related dependency.
Since we have Cake orm/core here anyway, having this other package shouldn't be a big deal.
Some might be concerned with the readability of standalone iterations now.

MasterOdin:
Sorry, I had more typed up in a draft of the PR message that I guess I forgot to include.

My motivation was that I wanted to better understand what it was that the collections was doing with the `unfold()` method as I found some of it a bit opaque as I've never done anything with that level of nested iterators and wanted to relate it to how it would look in just regular PHP. After doing a handful, I found that it was more readable and shorter in a number of places doing it in straight PHP rather that `cakephp/collections` with the added benefit of not involving iterator construction overhead (a rather micro optimization I'm sure) that I decided to see what the codebase looked like sans collection. I will not feel hurt if you decide to close this PR, as I'm happy with the better understanding I have of what all the iterators were doing, and learning about the `__invoke` magic function, though I would probably prefer to factor that away as it's not exactly a common method in PHP code.

> Some might be concerned with the readability of standalone iterations now.
I would argue almost all of the straight PHP transformations are more readable in that it does not require anyone to become familiar with a third-party library doing iterator operations (like I was doing this). Take for example this one:

```
        $indexes = collection($actions)
            ->map(function ($alter) use ($table, $columns, $dropIndexActions) {
                if ($alter->getTable()->getName() !== $table->getName()) {
                    return $alter;
                }
                $newAlter = new AlterTable($table);
                collection($alter->getActions())
                    ->map(function ($action) use ($columns) {
                        if (!$action instanceof DropIndex) {
                            return [$action, null];
                        }
                        if ($action->getIndex()->getColumns() === $columns) {
                            return [null, $action];
                        }

                        return [$action, null];
                    })
                    ->each(function ($tuple) use ($newAlter, $dropIndexActions) {
                        list($action, $dropIndex) = $tuple;
                        if ($action) {
                            $newAlter->addAction($action);
                        }
                        if ($dropIndex) {
                            $dropIndexActions->append($dropIndex);
                        }
                    });

                return $newAlter;
            })
            ->toArray();
```
vs
```
        $indexes = array_map(function ($alter) use ($table, $columns, $dropIndexActions) {
            if ($alter->getTable()->getName() !== $table->getName()) {
                return $alter;
            }

            $newAlter = new AlterTable($table);
            foreach ($alter->getActions() as $action) {
                if ($action instanceof DropIndex && $action->getIndex()->getColumns() === $columns) {
                    $dropIndexActions->append($action);
                } else {
                    $newAlter->addAction($action);
                }
            }
            return $newAlter;
        }, $actions);
```

The rewritten version is half the length, and does not require tracking multiple iterators, especially the inner collection which I view as being much less readable than the simple foreach loop (this similar pattern is repeated elsewhere again).

Given that this PR _reduces_ the codebase by 45 lines of code using builtins, without doing anything super fancy or clever, I'm not sure I buy the argument it makes things less readable. It additionally does not bring in the overhead of iterators for little gain as almost all operations can be done inline with the array (sans two) so it's not like having the iterators is saving on users memory or anything.

However, having done this PR, I do see a number of ways the current usage of collections could be improved to be shorter / more efficient.

othercorey:
I think we can merge several of these changes for simplification and cleanup. We don't need to worry about the goal of removing a dependency. Forcing an operation to use Collection is no better than forcing code to not use Collection if it's difficult to read/maintain.

I'll review and mark which changes I think are unnecessary just to remove the dependency.

src/Phinx/Db/Plan/Solver/ActionSplitter.php
```diff
@@ -68,48 +68,45 @@ public function __construct($conflictClass, $conflictClassDual, callable $confli
      * @return \Phinx\Db\Plan\AlterTable[] A list of AlterTable that can be executed without
      * this type of conflict
      */
-    public function __invoke(AlterTable $alter)
+    public function split(AlterTable $alter)
```
othercorey, line 71: Let's leave this as `__invoke`

src/Phinx/Db/Plan/Plan.php
```diff
@@ -209,30 +206,42 @@ protected function resolveConflicts()
...
-            // creating it again with the new stuff. Unfortunately, we have already bundled
-            // everything together in as few AlterTable statements as we could, so we need to
-            // resolve this conflict manually
-            ->unfold(new ActionSplitter(DropForeignKey::class, AddForeignKey::class, function (DropForeignKey $a, AddForeignKey $b) {
+            }
+        );
+        $tableUpdates = [];
+        foreach ($this->tableUpdates as $tableUpdate) {
+            array_push($tableUpdates, ...$splitter->split($tableUpdate));
+        }
```
othercorey, line 222:
```suggestion
        foreach ($this->tableUpdates as $update) {
            array_push($tableUpdates, $splitter($update));
        }
```

src/Phinx/Db/Plan/Plan.php
```diff
@@ -209,30 +206,42 @@ protected function resolveConflicts()
...
+            function (DropForeignKey $a, AddForeignKey $b) {
                 return $a->getForeignKey()->getColumns() === $b->getForeignKey()->getColumns();
-            }))
-            ->toList();
+            }
+        );
+        $constraints = [];
+        foreach ($this->constraints as $constraint) {
+            $constraint = $this->remapContraintAndIndexConflicts($constraint);
+            array_push($constraints, ...$splitter->split($constraint));
```
othercorey, line 242:
```suggestion
            $constraints = array_merge($constraints, $splitter($constraint));
```

src/Phinx/Db/Plan/Plan.php
```diff
@@ -442,61 +405,62 @@ protected function gatherCreates($actions)
...
-            ->reject(function ($action) {
-                return isset($this->tableCreates[$action->getTable()->getName()]);
-            })
-            ->each(function ($action) {
-                $table = $action->getTable();
-                $name = $table->getName();
+                    || $action instanceof RenameColumn
+                )
+                || isset($this->tableCreates[$action->getTable()->getName()])
+            ) {
```
othercorey, line 417: Let's just break this up into two if statements so it's cleaner.

src/Phinx/Db/Plan/Plan.php
```diff
@@ -442,61 +405,62 @@ protected function gatherCreates($actions)
...
-                    || $action instanceof RenameTable
-                    || $action instanceof ChangePrimaryKey
-                    || $action instanceof ChangeComment;
-            })
-            ->each(function ($action) {
-                $table = $action->getTable();
-                $name = $table->getName();
+        foreach ($actions as $action) {
+            if (
+                !(
```
othercorey, line 447: This does not read well.

src/Phinx/Db/Plan/Plan.php
```diff
@@ -508,26 +472,26 @@ protected function gatherTableMoves($actions)
...
-                // Indexes for new tables are created inline
-                // so we don't wan't them here too
-                return isset($this->tableCreates[$action->getTable()->getName()]);
-            })
-            ->each(function ($action) {
-                $table = $action->getTable();
-                $name = $table->getName();
+        foreach ($actions as $action) {
+            if (
+                !(
```
othercorey, line 477: Same here.

src/Phinx/Db/Plan/Solver/ActionSplitter.php
```diff
@@ -68,48 +68,45 @@ public function __construct($conflictClass, $conflictClassDual, callable $confli
...
-                }
-                if ($new) {
-                    $newAlter->addAction($new);
-                }
-            });
+            return [$action, null];
+        }, $alter->getActions());
+
+        foreach ($actions as $pair) {
+            [$original, $new] = $pair;
```
othercorey, line 102: This section should be re-written into the above array_map like the others or this leave this as a collection which reads easier.

src/Phinx/Db/Table.php
```diff
@@ -702,15 +702,9 @@ protected function executeActions($exists)
...
-        $renamed = collection($this->actions->getActions())
-            ->filter(function ($action) {
-                return $action instanceof RenameTable;
-            })
-            ->first();
-
-        if ($renamed) {
-            $exists = true;
-        }
+        $exists = $exists || count(array_filter($this->actions->getActions(), function ($action) {
```
othercorey, line 705: Revert this.

src/Phinx/Db/Plan/Solver/ActionSplitter.php
MasterOdin: done

src/Phinx/Db/Plan/Plan.php
MasterOdin: done

src/Phinx/Db/Plan/Plan.php
MasterOdin: done

src/Phinx/Db/Plan/Solver/ActionSplitter.php
MasterOdin: Worked this into the above array_map, which was rewritten to just be a foreach loop now as well.

src/Phinx/Db/Plan/Plan.php
MasterOdin: I'm not clear what you would suggest here. I would agree initially with just modifying it so that the `!(` are on the opening line:
```php
            if (!(
                $action instanceof DropTable
                || $action instanceof RenameTable
                || $action instanceof ChangePrimaryKey
                || $action instanceof ChangeComment
            )) {
                continue;
            }
```

An alternative would be to move the negation down a level:

```php
            if (
                !($action instanceof DropTable)
                && !($action instanceof RenameTable)
                && !($action instanceof ChangePrimaryKey)
                && !($action instanceof ChangeComment)
            ) {
                continue;
            }
```

The only other idea I've got would be to utilize an outer array_filter on `$actions`, except I'd want to avoid that in-favor of the inline option to save on array cycles.

othercorey: I would prefer the 2nd implementation. it seems like more text, but it's easier to read.

MasterOdin: alright, done.

src/Phinx/Db/Plan/Plan.php
MasterOdin: done

src/Phinx/Db/Plan/Plan.php
MasterOdin: done

src/Phinx/Db/Table.php
MasterOdin: semi-reverted so that it retains the same set-up, just using `current(array_filter(...))` instead of collection.

src/Phinx/Db/Table.php
```diff
@@ -703,11 +703,9 @@ protected function executeActions($exists)
...
         // down. We will just assume the table already exists if the user commands a
         // table rename.
-        $renamed = collection($this->actions->getActions())
-            ->filter(function ($action) {
-                return $action instanceof RenameTable;
-            })
-            ->first();
+        $renamed = current(array_filter($this->actions->getActions(), function ($action) {
+            return $action instanceof RenameTable;
+        }));
```
othercorey, line 708: foreach ($this->actions->getActions() as $action) {
    if ($action instanceof RenameTable) {
        $exists = true;
        break;
    }
}

MasterOdin: Yup, that's a way more efficient way of doing this. Missing the forest for the trees. 🤦 

I've also wrapped this with a `if (!$exists)` check as well.

PR #1755, opened 2020-04-28
Add support for tinyint. Fix bool issue.

dereuromark:
Part of https://github.com/cakephp/migrations/pull/457

Needed for https://github.com/cakephp/migrations/issues/355

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -946,7 +946,7 @@ class MysqlAdapter extends PdoAdapter
             case static::PHINX_TYPE_SMALL_INTEGER:
                 return ['name' => 'smallint', 'limit' => $limit ?: 6];
             case static::PHINX_TYPE_TINY_INTEGER:
-                return ['name' => 'tinyint', 'limit' => $limit ?: 3];
+                return ['name' => 'tinyint', 'limit' => $limit ?: 4];
             case static::PHINX_TYPE_INTEGER:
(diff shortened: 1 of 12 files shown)


dereuromark:
I pushed some CS fixes in a separate branch to easier verify and review these here
https://github.com/cakephp/phinx/compare/tinyint-cs

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -956,6 +957,8 @@ public function getSqlType($type, $limit = null)
                 return ['name' => 'bit', 'limit' => $limit ?: 64];
             case static::PHINX_TYPE_SMALL_INTEGER:
                 return ['name' => 'smallint', 'limit' => $limit ?: 6];
+            case static::PHINX_TYPE_TINY_INTEGER:
+                return ['name' => 'tinyint', 'limit' => $limit ?: 3];
```
ADmad, line 961: Isn't the default limit/length for tinyint in MySQL "4"?

dereuromark: Is it? I thought 0...255 would be 3, but since it will fill one byte anyway you might be right.

dereuromark: I did some research and the length here is the max display width, so given the negative char `-128` it seems to be 4 then.

dereuromark:
Whats wrong with stickler these days?

dereuromark:
Ah, now it is green. Maybe some PR change is necessary to update the push result.

PR #1756, opened 2020-04-30
Support user-configured DI container instantiation of seeds

joelperry93:
For our use-case of Phinx, we require some complex services for generating the correct seed data. This creates difficulties in having to manually instantiate all the services within the init methods of the Seed. If users could provide a PSR-11 container that implements the psr/container package interface then it would allow for much more flexability in this regard, including auto-wiring.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -318,6 +318,10 @@ abstract class AbstractCommand extends Command
     {
         if ($this->getManager() === null) {
             $manager = new Manager($this->getConfig(), $input, $output);
+            $container = $this->config->getContainer();
+            if ($container !== null) {
+                $manager->setContainer($container);
+            }
             $this->setManager($manager);
         } else {
             $manager = $this->getManager();
--- a/src/Phinx/Seed/AbstractSeed.php
+++ b/src/Phinx/Seed/AbstractSeed.php
@@ -40,17 +40,10 @@ abstract class AbstractSeed implements SeedInterface
     protected $output;
 
     /**
-     * @param \Symfony\Component\Console\Input\InputInterface|null $input Input
-     * @param \Symfony\Component\Console\Output\OutputInterface|null $output Output
+     * Override to specify dependencies for dependency injection from the configured PSR-11 container
      */
-    final public function __construct(InputInterface $input = null, OutputInterface $output = null)
+    public function __construct()
     {
-        if ($input !== null) {
-            $this->setInput($input);
-        }
-        if ($output !== null) {
-            $this->setOutput($output);
-        }
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "php": ">=7.2",
         "cakephp/collection": "^4.0",
         "cakephp/database": "^4.0",
+        "psr/container": "^1.0",
         "symfony/console": "^3.4|^4.0|^5.0",
         "symfony/config": "^3.4|^4.0|^5.0"
     },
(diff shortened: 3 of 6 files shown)


src/Phinx/Seed/AbstractSeed.php
```diff
@@ -40,17 +40,10 @@ abstract class AbstractSeed implements SeedInterface
     protected $output;
 
     /**
-     * @param \Symfony\Component\Console\Input\InputInterface|null $input Input
-     * @param \Symfony\Component\Console\Output\OutputInterface|null $output Output
+     * Override to specify dependencies for dependency injection from the configured PSR-11 container
      */
-    final public function __construct(InputInterface $input = null, OutputInterface $output = null)
+    public function __construct()
```
dereuromark, line 45: Wouldnt this change break all current seeds?

joelperry93: I don't see why it would. No one would have been able to override this constructor as it was final. Unless you can see something I can't.

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -318,6 +318,10 @@ protected function loadManager(InputInterface $input, OutputInterface $output)
     {
         if ($this->getManager() === null) {
             $manager = new Manager($this->getConfig(), $input, $output);
+            $container = $this->config->getContainer();
```
dereuromark, line 321: This might have caused a regression in master

> Error: [Error] Call to a member function getContainer() on null in vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php on line 321

dereuromark: For some reason this seems to fix it: https://github.com/cakephp/phinx/commit/18202cdb700cac236436dc1e0a934faf2dcf0b04

joelperry93: Oh dear. Sorry about that! I'm a bit confused as to how $this->config is null when above $this->getConfig() is passing $this->config into new Manager which expects an object which is not null. Could you give me steps to reproduce?

joelperry93: That is very weird. Can't see why that would make a difference.

dereuromark: Yeah, I also dont understand it. The changed code worked though for now.

MasterOdin: Is this in phinx proper or in a wrapper of it? How does `$this->config` be null, yet `getConfig()`, which simply returns `$this->config` return something that is not null?

dereuromark: It is in the Migrations wrapper ( https://github.com/cakephp/migrations )

joelperry93: Perhaps ->getConfig() is being overridden in a child class

mnsami:
I'm trying to set my own container in the config with something like this

```
"container" => \Face\Core\Container\Container::class,
```

But it doesn't work, I don't see an example of this in the documentation and I'm not sure how to make it work.

I'm using phpslim v3 framework.

@joelperry93 can you provide more insight? thanks

joelperry93:
You'll need to pass the container instance itself in, rather than the class string. That way you can wire up all your services within that container instance for the Seed class dependencies to be resolved against.

Hope that helps

mnsami:
First, thanks for getting back so fast, I really appreciate it, I doubted that, but if you are in a config file how is it possible to pass an instance? any. insights or pointers ?

joelperry93:
Well the config file is just a regular PHP file, so you have access to any general functionality you need to construct your container. It's project-specific on what that might look like. Perhaps you need to require some bootstrap file to bring it in to scope?

PR #1757, opened 2020-05-03
Simplify check for config in command bootstrap

MasterOdin:
The `$config` variable is not used beyond this check in the function, with everything after this point using `$this->getConfig()`.


dereuromark:
The reason was the phpstan check here, using inline annotation. I remember now.

MasterOdin:
Yeah, I guess the more proper thing would be to properly document that `$this->getConfig()` could return null, but that would then require type-checking everywhere else. Messy either way you slice it.

PR #1762, opened 2020-05-09
add symfony/yaml back as dependency for 0.11.x

MasterOdin:
As discussed with @dereuromark, this re-adds `symfony/yaml` as a dependency to the 0.11.x branch, as this was a breaking change and should have been held off on for 0.12.x release.

Follow-up here after merging would be:

* cut new 0.11.x release
* Update release notes for 0.12.x to mention this breaking change (symfony/yaml no longer installed by default, if using yaml conf file, you must install this dependency yourself).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,14 +29,14 @@
         "cakephp/collection": "^3.7",
         "cakephp/database": "^3.7",
         "symfony/console": "^3.4|^4.0|^5.0",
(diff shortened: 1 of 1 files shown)


MasterOdin:
Closing this as the build pipeline will never pass as it's using master's travis-ci configuration with 0.11 dependencies which does not work for the prefer-lowest flag.

dereuromark:
We need to fork off from last tag release here.

PR #1764, opened 2020-05-09
Split parsing DSN out to utility helper

othercorey:
This is preparation for replacing all the unit test defines with DB_DSN env variables similar to cakephp core.

The regex was replaced with the regex used by `StaticConfigTrait::parseDsn()` with two behavior changes:

- `name` does not include the leading `/`
- There is no `fragment` string parsed, only `query`

The only originally valid pattern that is now rejected is the random character string. The adapter names can't be completely arbitrary symbols now.

### DIFF
--- a/tests/Phinx/Config/ConfigTest.php
+++ b/tests/Phinx/Config/ConfigTest.php
@@ -14,32 +14,6 @@ use UnexpectedValueException;
 class ConfigTest extends AbstractConfigTest
 {
     /**
-     * @covers \Phinx\Config\Config::__construct
-     * @covers \Phinx\Config\Config::getConfigFilePath
-     */
-    public function testConstructEmptyArguments()
-    {
-        $config = new Config([]);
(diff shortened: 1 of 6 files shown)


tests/Phinx/Config/ConfigTest.php
```diff
@@ -13,32 +13,6 @@
...
-     */
-    public function testConstructEmptyArguments()
-    {
-        $config = new Config([]);
-        // this option is set to its default value when not being passed in the constructor, so we can ignore it
-        unset($config['version_order']);
-        $this->assertAttributeEmpty('values', $config);
-        $this->assertAttributeEmpty('configFilePath', $config);
-        $this->assertNull($config->getConfigFilePath());
-    }
```
othercorey, line 28: These were removed because asserting private and protected properties is deprecated in phpunit. These tests don't really validate anything other than an empty constructor.

PR #1769, opened 2020-05-09
Do not set precision for integer columns coming from postgres DB

MasterOdin:
The code would use the value of numeric_precision to set a limit on integer columns. For integer values, this is set to the max range of the type (e.g. 2^32 for integer), but is not actually something the user can meaningfully use, nor has a counterpart for info coming into postgres in a migration, so better to just report null for limit instead.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -440,10 +440,15 @@ class PostgresAdapter extends PdoAdapter
 
             if (in_array($columnType, [static::PHINX_TYPE_TIME, static::PHINX_TYPE_DATETIME], true)) {
                 $column->setPrecision($columnInfo['datetime_precision']);
-            } else {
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -440,10 +440,15 @@ public function getColumns($tableName)
 
             if (in_array($columnType, [static::PHINX_TYPE_TIME, static::PHINX_TYPE_DATETIME], true)) {
                 $column->setPrecision($columnInfo['datetime_precision']);
-            } else {
+            } elseif (
+                !in_array($columnType, [
+                    self::PHINX_TYPE_SMALL_INTEGER,
+                    self::PHINX_TYPE_INTEGER,
+                    self::PHINX_TYPE_BIG_INTEGER,
```
dereuromark, line 447: what about TINY_INTEGER?

PR #1771, opened 2020-05-09
Allow name to be optional for create command

### LINKED ISSUE #566: Possibility of UnNamed Migrations?

MasterOdin:
Closes #566 (if it wasn't already closed for being "inactive"...)

Currently, the create command requires a name to be provided, and this is used for a suffix on the filename and then as the class name. This PR makes the name optional, such that if it is omitted, the system will automatically generate a name that is valid. The filename is "<datetime>_migration.php" and the class name is "V<Datetime>".

The filename is set-up such as the way it is since the classname must start with a string, and the `V` is for a future suggestion of potentially making phinx migration files be PSR-4 discoverable.

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -41,7 +41,7 @@ class Create extends AbstractCommand
         parent::configure();
 
         $this->setDescription('Create a new migration')
-            ->addArgument('name', InputArgument::REQUIRED, 'What is the name of the migration (in CamelCase)?')
+            ->addArgument('name', InputArgument::OPTIONAL, 'Class name of the migration (in CamelCase)')
             ->setHelp(sprintf(
                 '%sCreates a new database migration%s',
                 PHP_EOL,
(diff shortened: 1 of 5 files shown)


src/Phinx/Console/Command/Create.php
```diff
@@ -165,12 +165,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $path = realpath($path);
         $className = $input->getArgument('name');
+        if ($className === null) {
+            $className = "V" . Util::getCurrentTimestamp();
+            $fileName = Util::getCurrentTimestamp() . '_migration.php';
```
dereuromark, line 170: You shouldnt call this twice, as this can lead to leap second issues.
Best to store in a variable and reuse that one.

MasterOdin: oh, yeah, duh

MasterOdin: done

MasterOdin:
@dereuromark revised the PR some, added better tests in Utils and such. Not sure what the best way to add tests to Manager for this that would not be full-on integration tests.

PR #1773, opened 2020-05-10
Consolidate environment and environmentOptions printing for commands

### LINKED ISSUE #1116: Creating SQLite in memory fails
Attempt to create SQLIte database in memory as documented in docs, i.e.

```
environments:
    testing:
        adapter: sqlite
        memory: true     # Setting memory to *any* value overrides name
```

fails:

```
$ vendor/bin/phinx migrate -e testing
Phinx by Rob Morgan - https://phinx.org. 0.8.1

using config file ./phinx.yml
using config parser yaml
using migration paths 
 - /<redacted>/db/migrations
using seed paths 
 - /<redacted>/db/seeds
using environment testing
using adapter sqlite
Could not determine database name! Please specify a database name in your config file.
```

The workaround for this is to provide `name` regardless it is used or not:

```
environments:
    testing:
        adapter: sqlite
        name: /tmp/db
        memory: true
```

MasterOdin:
Closes #1116

Several commands were using the same sequence of checks and outputs for getting the environment string, as well as then printing out details about the environment. This commit consolidates all of that repeated code into two functions within AbstractCommand. As part of this, also fixed the bug of not being able to use sqlite memory database without also setting a db name.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -401,4 +401,73 @@ abstract class AbstractCommand extends Command
     {
         return __DIR__ . self::DEFAULT_SEED_TEMPLATE;
     }
+
+    /**
+     * Gets environment to use for command.
+     *
+     * @return string|null name of environment, null if could not be found
+     */
+    protected function getEnvironment(InputInterface $input, OutputInterface $output): ?string
+    {
+        $environment = $input->getOption('environment');
+
+        if ($environment === null) {
+            $environment = $this->getConfig()->getDefaultEnvironment();
+            $output->writeln('<comment>warning</comment> no environment specified, defaulting to: ' . $environment);
+        } else {
+            $output->writeln('<info>using environment</info> ' . $environment);
+        }
+
+        if (!$this->getConfig()->hasEnvironment($environment)) {
+            $output->writeln(sprintf('<error>The environment "%s" does not exist</error>', $environment));
+
+            return null;
+        }
+
+        return $environment;
+    }
+
+    /**
+     * Prints out general runtime information about loaded environment.
+     *
+     * @param array|null $envOptions environment options
+     * @param \Symfony\Component\Console\Output\OutputInterface $output Output
+     * @return bool
+     */
+    protected function checkEnvironmentOptions(?array $envOptions, OutputInterface $output): bool
+    {
+        if (isset($envOptions['adapter'])) {
+            $output->writeln('<info>using adapter</info> ' . $envOptions['adapter']);
+        }
+
+        if (isset($envOptions['wrapper'])) {
+            $output->writeln('<info>using wrapper</info> ' . $envOptions['wrapper']);
+        }
+
+        if (
+            isset($envOptions['adapter'])
+            && $envOptions['adapter'] === 'sqlite'
+            && isset($envOptions['memory'])
+            && $envOptions['memory'] === true
+        ) {
+            $output->writeln('<info>using :memory: database</info>');
+        } elseif (isset($envOptions['name'])) {
+            $output->writeln('<info>using database</info> ' . $envOptions['name']);
+        } else {
+            $output->writeln(
+                '<error>Could not determine database name! Please specify a database name in your config file.</error>'
+            );
+
+            return false;
+        }
+
+        if (isset($envOptions['table_prefix'])) {
+            $output->writeln('<info>using table prefix</info> ' . $envOptions['table_prefix']);
+        }
+        if (isset($envOptions['table_suffix'])) {
+            $output->writeln('<info>using table suffix</info> ' . $envOptions['table_suffix']);
+        }
+
+        return true;
+    }
 }
--- a/src/Phinx/Console/Command/Breakpoint.php
+++ b/src/Phinx/Console/Command/Breakpoint.php
@@ -62,22 +62,13 @@ EOT
     {
         $this->bootstrap($input, $output);
 
-        $environment = $input->getOption('environment');
         $version = (int)$input->getOption('target') ?: null;
         $removeAll = $input->getOption('remove-all');
         $set = $input->getOption('set');
         $unset = $input->getOption('unset');
 
+        $environment = $this->getEnvironment($input, $output);
         if ($environment === null) {
-            $environment = $this->getConfig()->getDefaultEnvironment();
-            $output->writeln('<comment>warning</comment> no environment specified, defaulting to: ' . $environment);
-        } else {
-            $output->writeln('<info>using environment</info> ' . $environment);
-        }
-
-        if (!$this->getConfig()->hasEnvironment($environment)) {
-            $output->writeln(sprintf('<error>The environment "%s" does not exist</error>', $environment));
-
             return self::CODE_ERROR;
         }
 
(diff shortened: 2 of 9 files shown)


MasterOdin:
If so desired, I can split out the sqlite memory db changes into its own changeset that builds off the overall consolidation, as that was the whole motivation for doing this in the first place, or can just leave it as one commit.

othercorey:
What part of this fixes in-memory sqlite db?

MasterOdin:
@othercorey https://github.com/cakephp/phinx/blob/6eb8895bd504b64ffdce02cc9d92758bcb98a1f4/src/Phinx/Console/Command/AbstractCommand.php#L447-L452

src/Phinx/Console/Command/SeedRun.php
```diff
@@ -57,45 +57,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $envOptions = $this->getConfig()->getEnvironment($environment);
-        if (isset($envOptions['adapter'])) {
-            $output->writeln('<info>using adapter</info> ' . $envOptions['adapter']);
-        }
-
-        if (isset($envOptions['wrapper'])) {
-            $output->writeln('<info>using wrapper</info> ' . $envOptions['wrapper']);
-        }
-
-        if (isset($envOptions['name'])) {
```
othercorey, line 84: The printing out of environment information doesn't seem to fix it. I assume removing this check is the actual fix.

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -401,4 +401,73 @@ protected function getSeedTemplateFilename()
...
+            $output->writeln(sprintf('<error>The environment "%s" does not exist</error>', $environment));
+
+            return null;
+        }
+
+        return $environment;
+    }
+
+    /**
+     * Prints out general runtime information about loaded environment.
```
othercorey, line 431: This seems to be a validation routine. The bool return is ambiguous. There should be a validation routine that's separate from logging.

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -401,4 +401,73 @@ protected function getSeedTemplateFilename()
...
+        if (isset($envOptions['adapter'])) {
+            $output->writeln('<info>using adapter</info> ' . $envOptions['adapter']);
+        }
+
+        if (isset($envOptions['wrapper'])) {
+            $output->writeln('<info>using wrapper</info> ' . $envOptions['wrapper']);
+        }
+
+        if (
+            isset($envOptions['adapter'])
```
othercorey, line 448: Isn't the adapter required? Validation should early out on missing required information.

src/Phinx/Console/Command/Breakpoint.php
```diff
@@ -62,22 +62,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->bootstrap($input, $output);
 
-        $environment = $input->getOption('environment');
         $version = (int)$input->getOption('target') ?: null;
         $removeAll = $input->getOption('remove-all');
         $set = $input->getOption('set');
         $unset = $input->getOption('unset');
 
+        $environment = $this->getEnvironment($input, $output);
```
othercorey, line 70: It doesn't look like `getEnvironment()` caches the result. Instead, it will trigger fallback logic and logging each time it is called.

src/Phinx/Console/Command/AbstractCommand.php
MasterOdin: Technically, no. It'll crash later on in this process, but right now there's no error checking when loading the config file / environment options.

othercorey: How is that not a required setting then?

src/Phinx/Console/Command/Breakpoint.php
MasterOdin: This is true of the `bootstrap` function as well. I guess the name here is a bit misleading on what it does, but I could not come up with a better one as I'm not sure I see the value in doing caching on this method.

othercorey: The value is a `get()` that doesn't logic fallback and failure logic each time. This code is trying to do too much at once.

src/Phinx/Console/Command/AbstractCommand.php
MasterOdin: Sorry, I misspoke slightly. So you can either specify:

* `adapter` (string that's mysql, sqlite, etc.)
* `dsn` (that gets parsed into components, and will set adapter when loading config)
* `connection` (PDO object, will set adapter once Environment class is initialized which is done by the Manager when it's doing its thing after all of these prints are done).

https://github.com/cakephp/phinx/blob/a14cc9e0a9258b30f20e5c1a711a3a8c266a3c91/src/Phinx/Migration/Manager/Environment.php#L337-L344

Incidentally, the Environment class is where the lack of an adapter is finally thrown:
https://github.com/cakephp/phinx/blob/a14cc9e0a9258b30f20e5c1a711a3a8c266a3c91/src/Phinx/Migration/Manager/Environment.php#L345-L347

So, it could be that those 10 lines get moved into the Config class where the environment is loaded (and after `parseAgnosticDsn` is run) and then `adapter` would be guaranteed to be set when that block of code we're talking about here runs.

src/Phinx/Console/Command/SeedRun.php
MasterOdin: The printing out of information is where it was crashing, because name was not set. I could have simply added the check to all three files, but that's not coding. Why I condensed it while making my sqlite fix (with the exact lines posted above).

src/Phinx/Console/Command/AbstractCommand.php
othercorey: In that case, I think we should validate them when constructing the environment. Validating and logging them in the command feels out of order.

The command can check if it has a valid environment and log then fail.

src/Phinx/Console/Command/SeedRun.php
othercorey: Just linking me that code didn't explain anything. You're defending a decision that I don't even know happened.

MasterOdin: The problem is that having an environment of:

```
        adapter: sqlite
        memory: true
```

gives you:

```
$envOptions = [
    'adapter' => 'sqlite',
    'memory' => true
];
```

which when it hit:
```
        if (isset($envOptions['name'])) {
            $output->writeln('<info>using database</info> ' . $envOptions['name']);
        } else {
            $output->writeln('<error>Could not determine database name! Please specify a database name in your config file.</error>');

            return self::CODE_ERROR;
        }
```
would fail the if statement, enter the else and return an error because name was not set. When I consolidated these if checks from these three different commands, I added the special check of:

```php
        if (
            isset($envOptions['adapter'])
            && $envOptions['adapter'] === 'sqlite'
            && isset($envOptions['memory'])
            && $envOptions['memory'] === true
        ) {
```

on that if/else to catch this case, and so now the above configuration works fine and without error. This gets the [second sqlite configuration example in the docs](https://book.cakephp.org/phinx/0/en/configuration.html#id1) to work.

src/Phinx/Console/Command/AbstractCommand.php
MasterOdin: Construction of the environment happens after all of these log statements (in the manager method that each of these methods calls).

I suppose each of these methods could call `$this->getManager()->getEnvironment()` before printing out the environment details as they do now to get that error checking. Or do some other sort of larger refactor.

MasterOdin:
Closing this as a PR is the wrong venue for discussing fix large systemic problems of the command structure of phinx.

PR #1776, opened 2020-05-10
Fix docstring return type for child methods that always throw

MasterOdin:
phpstan 0.12.24 added the capacity to handle merging phpdocs, which was not the case for three functions which always threw, and so defined themselves as returning void. This removes the incompatible return, and adds a note to the functions that they always throw.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -453,7 +453,7 @@ class SQLiteAdapter extends PdoAdapter
     }
 
     /**
-     * {@inheritDoc}
+     * @inheritDoc
      *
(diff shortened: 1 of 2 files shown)


MasterOdin:
As I do not use phpdoc all that much, I'm not sure what the considered best practices here are where you combine inheritdoc with a note about a particular implementation, especially where inheritdoc defines the params, returns, etc. and then the implementation note comes after those fields, which is not normally the case?

I'm happy to change this, but either way, all future builds of phpstan in Travis are going to fail for phinx until the `@return void` are removed.

dereuromark:
`{}` should be removed here usually, and `@inheritDoc` used where possible.
That should fix it, when you add additional throws on top

Otherwise, try to keep the return type to what it actually returns.
void was already wrong here it seems.

othercorey:
phinx doesn't use the current cakephp codesniffer standard unfortunately.

PR #1778, opened 2020-05-10
fix using quoted and escaped values for mysql enums

### LINKED ISSUE #887: MySQL enum values with special characters

MasterOdin:
Fixes #887.

The previous implementation was relatively simplistic in that it would break if attempting to use a value with a `'` character in it and would give different result in getValues() after pulling the column out of the DB if it had a `'` or `,` in it. This patch makes the process more robust by using PDO::quote to handle values going in properly and then de-escapes things on the way out. Unfortunately, there's no builtin method we can use to get "unescaped" strings and have to do it by hand in custom fashion. Luckily, mysql's escape rules on values coming out of the DB are relatively simple with just really '' and \\ being the things we have to worry about.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1119,7 +1119,51 @@ class MysqlAdapter extends PdoAdapter
         ];
 
         if ($type === static::PHINX_TYPE_ENUM || $type === static::PHINX_TYPE_SET) {
-            $phinxType['values'] = explode("','", trim($matches[6], "()'"));
+            $values = trim($matches[6], "()");
+            $phinxType['values'] = [];
+            $opened = false;
+            $escaped = false;
+            $wasEscaped = false;
+            $value = '';
+            $valuesLength = strlen($values);
+            for ($i = 0; $i < $valuesLength; $i++) {
+                $char = $values[$i];
+                if ($char === "'" && !$opened) {
+                    $opened = true;
+                } elseif (
+                    !$escaped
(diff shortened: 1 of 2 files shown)


MasterOdin:
Fixing stickler-ci error (where it pointed at the wrong line to boot) and now codecov is failing as I had to touch unrelated lines of code to this patch. Ugh.

dereuromark:
Wow, quite the fix ;)

codecov can be wrong, don't wry about that.

PR #1782, opened 2020-05-11
Added support for configuring test db through DSN

othercorey:
This removes the large list of environment variables needed to configure a test and replaces with a single environment variable `TEST_DSN`. This dsn is parsed into TEST_DB_CONFIG which all the tests used to configure.

This also fixes `MysqlAdapter` and `PostgresAdapter` when user or pass is default.

Code coverage is updated to coveralls to match cakephp core and allow multiple builds to submit coverage.


tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -102,18 +89,12 @@ public function testConnectionWithInvalidCredentials()
...
-        $options = [
-            'name' => TESTS_PHINX_DB_ADAPTER_MYSQL_DATABASE,
-            'user' => TESTS_PHINX_DB_ADAPTER_MYSQL_USERNAME,
-            'pass' => TESTS_PHINX_DB_ADAPTER_MYSQL_PASSWORD,
-            'unix_socket' => TESTS_PHINX_DB_ADAPTER_MYSQL_UNIX_SOCKET,
-        ];
-
-        $adapter = new MysqlAdapter($options, new ArrayInput([]), new NullOutput());
+        $options = ['unix_socket' => getenv('MYSQL_UNIX_SOCKET')] + TEST_DB_CONFIG;
+        $adapter = new MysqlAdapter(TEST_DB_CONFIG, new ArrayInput([]), new NullOutput());
```
othercorey, line 97:
```suggestion
        $adapter = new MysqlAdapter($options, new ArrayInput([]), new NullOutput());
```

MasterOdin:
For my local development workflow, this is going to make things a lot more annoying as now instead of just doing `composer run test` to test all adapters at once, I've got to run it four times while juggling the `TEST_DSN` variable.

othercorey:
Most test harnessed don't test multiple databases at the same time. You would normally iterate on one and final test on all.

However, if that is a hard requirement for phinx, we can split it up.

dereuromark:
I wonder if we can provide a high level entry point or composer script to test all at once locally.
Could also be something you put into your phpunit.xml and it will auto-read from it.

@MasterOdin btw: `composer run test` can just be used as `composer test`

MasterOdin:
> final test on all

Yeah, that's what I meant by testing against all databases at once. My workflow is roughly:

1. Make changes
1. Run specific tests that I know are affected (be it specific test class, specific adapter, using `--filter` argument)
1. Run all tests against all databases
1. Run linting / static analysis

That third step now is a much greater nuisance to do now for the sake of CI cleanliness. It's also repeating several hundred tests per run that are not database specific just making the whole thing take longer. Ideally, I think that phpunit.xml should have DB specific variables (e.g. `MYSQL_TEST_DSN`, `POSTGRES_TEST_DSN`) that parse into DB specific configs (e.g. `MYSQL_TEST_DB_CONFIG`) that then activate the various adapter specific tests.

othercorey:
Since this test suite only takes a few seconds, running a few extra tests isn't a concern.

To ease the transition, the dsn is now split into separate environment variables for each db:

- `SQLITE_DSN`
- `MYSQL_DSN`
- `PGSQL_DSN`
- `SQLSRV_DSN`

PR #1785, opened 2020-05-12
Set default timezone if not set in php.ini for cli

### LINKED ISSUE #852: PHP warning system's timezone
Add ability to add timezone in phinx.yml and use it set environment. 

PHP Warning:  date(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in 
/robmorgan/phinx/src/Phinx/Migration/Manager/Environment.php on line 126

MasterOdin:
Closes #852 

If the php.ini setting is not set, then PHP will always fallback to UTC and issue a warning, which will get issued in the middle of a run. This just explicitly sets the timezone to UTC when using the CLI app if it's not set.

This maybe should issue a warning if we set it, or I guess just document it in the release notes that this happens, though given how little this was reported, it's also maybe probable this does not affect a ton of people.

### DIFF
--- a/app/phinx.php
+++ b/app/phinx.php
@@ -25,6 +25,10 @@
 
 $autoloader = require __DIR__ . '/../src/composer_autoloader.php';
 
+if (!ini_get('date.timezone')) {
+    date_default_timezone_set('UTC');
+}
+
 if (!$autoloader()) {
     die(
         'You need to set up the project dependencies using the following commands:' . PHP_EOL .


othercorey:
Wouldn't you want this to be configured to match the user's app environment? Should we just error and exit?

MasterOdin:
ugh, misread what was asked, closing this and doing it proper later.

PR #1788, opened 2020-05-21
Fix gh-actions mysql8 workflow

MasterOdin:
#1782 broke the mysql8 github actions workflow as it converted the environment variable to an invalid yaml key/value, using a `=` instead of a `:` between the key and value in a dictionary.

Example of broken run: https://github.com/cakephp/phinx/actions/runs/107320570


othercorey:
Thanks!

Seems like malformed github actions yml should show up as an error instead of 0 runs.

PR #1789, opened 2020-05-21
Ignore un-testable function/class existence in Config

MasterOdin:
Without adding something like [php-mock/php-mock](https://github.com/php-mock/php-mock), it's not possible to enter these if statements and so we should not collect code coverage statistics for it.


othercorey:
I think it's fair to say there is no code coverage for it.

PR #1794, opened 2020-05-25
use php as default configuration file format

### LINKED ISSUE #1726: Missing yaml parser, symfony/yaml package is not installed.
We're using Phinx in a plain PHP project with no other dependencies. After Phinx update to v 0.11.6 we get an error: "Missing yaml parser, symfony/yaml package is not installed.". Seems like Phinx misses YAML dependency.

MasterOdin:
Wasn't sure if this should be on 0.13 or 0.12 as on the one hand it's a BC break, on the other hand, the current default (yml) does not work out of the box on 0.12 due to the making optional of symfony/yaml.

In 0.12, the symfony/yaml extension was made optional, which means it may no longer be available in a user's environment. As such, phinx should not default to using it for `init` as it may generate a file the user cannot actually use. Instead, we use `php` as the default as it's guaranteed to always work and be an available parser (as `ext_json` is equally optional).

Closes #1726

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ tests/log
 /db/
 /migrations/
 /phinx.yml
+/phinx.yaml
+/phinx.php
+/phinx.json
 
 # python artifacts
 *.pyc
--- a/docs/en/commands.rst
+++ b/docs/en/commands.rst
@@ -77,27 +77,30 @@ The Init Command
 ----------------
 
 The Init command (short for initialize) is used to prepare your project for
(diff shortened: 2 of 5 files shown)


dereuromark:
If the defaults stay the same, this sounds BC enough for 0.12

MasterOdin:
The default for `phinx init` is changing from `yml` to `php`, so the defaults are changing, it's just that the defaults right now is broken out of the box, so this could be considered just a bugfix, even though it's changing behavior.

dereuromark:
As this is documented this should be fine as bugfix then.

Issue #868, opened 2016-06-17
MYSQL default value: CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP


ricksanchez:
It is not supported yet:

```
$table->addColumn('updated', 'datetime', array('default' => 'CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'))
```

Error:

```
  [PDOException]                                                                               
  SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'updated'
```

Ma-ve:
That's a feature I'd like too.

In the meantime, you can do this:

    $table = $this->table('my_table');
    $table
        ->addColumn('updated_at', 'datetime')
        ->save();
     $this->execute('ALTER TABLE `my_table` MODIFY COLUMN `updated_at` DATETIME DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP');

rquadling:
According to http://dev.mysql.com/doc/refman/5.5/en/timestamp-initialization.html

> With neither DEFAULT CURRENT_TIMESTAMP nor ON UPDATE CURRENT_TIMESTAMP, it is the same as specifying both DEFAULT CURRENT_TIMESTAMP and ON UPDATE CURRENT_TIMESTAMP.

So, technically, Phinx DOES support this. Just don't supply the default value for a timestamp type. But I've not tested that yet.

In addition to that though, Phinx also supports an `'update'` option for `timestamp` : http://docs.phinx.org/en/latest/migrations.html#working-with-columns

So
```php
$table->addColumn('updated', 'timestamp', array('default' => 'CURRENT_TIMESTAMP', 'update' => 'CURRENT_TIMESTAMP'));
```
may be more appropriate. Notice the column type is a `timestamp` and not a `datetime`.

mamchenkov:
We are seeing this issue now as well.  Is there any way we can help with resolving the issue?

lorenzo:
@mamchenkov send a pull request if you know how to fix it :)

mamchenkov:
@lorenzo after digging a bit more into it, it seems the newer versions have the fix already applied:

https://github.com/cakephp/phinx/blob/v0.8.0/src/Phinx/Db/Adapter/PdoAdapter.php#L531-L532

If we could only get v0.8.0 into cakephp/migrations now. :)

lorenzo:
[...]

PR #1798, opened 2020-05-28
addTimestamps: update updated_at timestamp on row update

kevingentile:
Link #868 . Updates don't reflect in the updated_at column for the default convenience methods.

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -551,6 +551,7 @@ class Table
              ->addColumn($updatedAt, 'timestamp', [
                  'null' => true,
                  'default' => null,
+                 'update' => 'CURRENT_TIMESTAMP',
                  'timezone' => $withTimezone,
              ]);
 
--- a/tests/Phinx/Db/TableTest.php
+++ b/tests/Phinx/Db/TableTest.php
@@ -131,7 +131,7 @@ class TableTest extends TestCase
         $this->assertEquals($expectedUpdatedAtColumnName, $columns[1]->getName());
         $this->assertEquals('timestamp', $columns[1]->getType());
         $this->assertEquals($withTimezone, $columns[1]->getTimezone());
-        $this->assertEquals('', $columns[1]->getUpdate());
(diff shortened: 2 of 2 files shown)


kevingentile:
Looks like the SQLiteAdapter doesn't like the update syntax:

```
1) Test\Phinx\Db\Adapter\SQLiteAdapterTest::testAlterTableColumnAdd
PDOException: SQLSTATE[HY000]: General error: 1 near "UPDATE": syntax error
```

MasterOdin:
I will take a look later on today or tomorrow. SQLite is annoying with how it supports altering a table with some pretty awful restrictions that require cumbersome workarounds.

kevingentile:
@MasterOdin cool! Thanks for checking. It looks like some level of abstraction around the `Column` would let you override `getValidOptions` so at least you can remove the "update" option and handle the exception.

MasterOdin:
Ah, it looks like we should delete this check from the sqlite adapter:
https://github.com/cakephp/phinx/blob/54ece1f0ca380218ab2c93216929f101326f82df/src/Phinx/Db/Adapter/SQLiteAdapter.php#L1483-L1485

SQLite does not support the `ON UPDATE` syntax (it's specific to mysql) and postgresql and sqlserver just ignore that value. It's possible to support a similar function within non-MySQL DBs, but it requires creating triggers which phinx does not have a good handle for dealing with automatically. I've created PR #1800 to clarify this as well as remove the offending lines from sqlite adapter.

-----

This all does raise the point of whether or not this should be accepted as it does make the default behavior of one adapter (MySQL) different than the others as it's the only one that supports this clause.

dereuromark:
Can we merge?

MasterOdin:
> Can we merge?

Assuming people are fine with MySQL having different behavior, yes? I can write up a PR adding a note about this to the documentation.

dereuromark:
@lorenzo @markstory @othercorey any opionions?

markstory:
I'm fine with MySQL being different. People using MySQL will want to be able to use `ON UPDATE` syntax.

dereuromark:
We might want to further document this then, though. Merging for now.

Issue #1799, opened 2020-05-28
MysqlAdapter getPrimaryKey method doesn't specify schema.


ajmahoney:
The MysqlAdapter.getPrimaryKey() method does not specify which database schema to get the key from this can lead to errors during Primary key operations when a primary key exists on another table, of the same name, on the same database server.
e.g. whilst running
```
        $table = $this->table($this->my_table);
        $table
            ->removeIndex([$this->column1, $this->column2])
            ->removeColumn("id")
            ->changePrimaryKey([$this->column1, $this->column2])
            ->save();
```
The same may be true for the SqlServerAdapter but the PostgresAdapter appears to do this.

PR #1803, opened 2020-05-29
MySQL use schema name in getPrimaryKey

ajmahoney:
MySQL now only checks current database schema for Primary Key, rather than the entire server for a table name. #1799


ajmahoney:
Is there something I need to do to trigger the checks (e.g. push an empty commit) for the failed Travis CI sqllite PHP 7.4 checks?
Thanks

PR #1807, opened 2020-06-16
Fix issues with Postgres schemas that contain hyphens and underscores.

rrigby:
Escape schema names so underscores and hyphens work. Also use 'CREATE SCHEMA IF NOT EXISTS'.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -79,7 +79,7 @@ class PostgresAdapter extends PdoAdapter
 
             try {
                 if (isset($options['schema'])) {
-                    $db->exec('SET search_path TO ' . $options['schema']);
+                    $db->exec('SET search_path TO ' . $this->quoteSchemaName($options['schema']));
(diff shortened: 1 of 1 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -1297,7 +1297,7 @@ public function createSchemaTable()
     public function createSchema($schemaName = 'public')
     {
         // from postgres 9.3 we can use "CREATE SCHEMA IF NOT EXISTS schema_name"
-        $sql = sprintf('CREATE SCHEMA %s', $this->quoteSchemaName($schemaName));
+        $sql = sprintf('CREATE SCHEMA IF NOT EXISTS %s', $this->quoteSchemaName($schemaName));
```
dereuromark, line 1300: What about the comment and <9.3?

MasterOdin: postgresql 9.3 was end of life back in 2018, and I do not think it should be a target for support of phinx.

Issue #1813, opened 2020-06-21
No way to configure schema name


Bilge:
It seems to me there is no way to configure which schema Phinx uses. It just unceremoniously dumps its table into the `public` schema assuming that's acceptable. I have a sneaking suspicion this may be because MySQL has poor or no support for schemas, and Phinx makes no secret about being a MySQL-first, other-engines-second project. Supporting schemas would go some way towards correcting this bias.

Bilge:
Although undocumented, this is supported with, for example, `default_migration_table: phinx.log`.

PR #1815, opened 2020-06-21
add subsection on migration table to configuration doc

MasterOdin:
As pointed out in #1813, there is no mention in the docs about setting the `default_migration_table` setting in the configuration. This adds such a subsection.

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -186,6 +186,22 @@ file:
 
     export PHINX_ENVIRONMENT=dev-`whoami`-`hostname`
 
+Migration Table
+---------------
+
+To keep track of the migration statuses for an environment, phinx creates
+a table to store this information. You can customize where this table
+is created by configuring ``default_migration_table``:
+
+.. code-block:: yaml
+
+    environment:
+        default_migration_table: phinxlog
+
+If this field is omitted, then it will default to ``phinxlog``. For adapters
+that support it (e.g. Postgres), you can prefix the value with a schema if
+you to have the table in a different schema as the rest of the environment,
+for example ``phinx.log`` to create a table ``log`` in the ``phinx`` schema.
 
 Table Prefix and Suffix
 -----------------------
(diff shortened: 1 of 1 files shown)


docs/en/configuration.rst
```diff
@@ -186,6 +186,22 @@ file:
...
+is created by configuring ``default_migration_table``:
+
+.. code-block:: yaml
+
+    environment:
+        default_migration_table: phinxlog
+
+If this field is omitted, then it will default to ``phinxlog``. For adapters
+that support it (e.g. Postgres), you can prefix the value with a schema if
+you to have the table in a different schema as the rest of the environment,
```
Bilge, line 203: >if you to have the table in a different schema as the rest of the environment

This isn't strictly when you would have to specify the schema. The schema needs to be specified if it's not the default `public` schema. However, I think this whole clause can just be removed because people wanting to specify a schema should already know why they want to do this. I might write this as the following,

_For databases that support it, e.g. Postgres, the schema name can be prefixed with a period separator (`.`). For example, `phinx.log` will create the table `log` in the `phinx` schema._

MasterOdin: Makes sense. Changed to your suggestion.

docs/en/configuration.rst
```diff
@@ -186,6 +186,23 @@ file:
...
+
+.. code-block:: yaml
+
+    environment:
+        default_migration_table: phinxlog
+
+If this field is omitted, then it will default to ``phinxlog``. For
+databases that support it, e.g. Postgres, the schema name can be prefixed
+with a period separator (``.``). For example, ``phinx.log`` will create
+the table log in the ``phinx`` schema instead of ``phinxlog`` in the
```
Bilge, line 204: `log` should have been ` ``log`` `.

dereuromark: Thank you, resolved in https://github.com/cakephp/phinx/commit/96c731595faff67dbcca2c303c067fcf219d226a

PR #1819, opened 2020-06-22
Changed migration template to adopt strict types

Bilge:
Since Phinx targets PHP >= 7.2, it is safe and probably a good idea to include the `strict_types` declaration. How good of an idea? Well, not *very* good, because Phinx's API appears to be entirely devoid of scalar type hints and also misses `array` type hints in [places where only arrays are valid](https://github.com/cakephp/phinx/blob/ac22eb22bd6bc36fab58610f385bc1f12d7f2150/src/Phinx/Db/Table.php#L295). However, once these areas are updated, templates from now on will be configured correctly to take advantage of type hints and the errors generated by mismatching types. Without these type hints, PHP just emits warnings such as `Invalid argument supplied for foreach() `, but continues regardless until something worse blows up later on.

You will also notice I added a few other bonuses which can be reverted if it's too many changes for this PR. First is a `final` type declaration to signal that migrations are not intended to be inherited (there should be no use case for this). The second is the `void` return type hint just to keep static analysers happy. Again, either of these can be reverted but it is not expected to be particularly contentious.

### DIFF
--- a/src/Phinx/Migration/Migration.template.php.dist
+++ b/src/Phinx/Migration/Migration.template.php.dist
@@ -1,8 +1,10 @@
 <?php
+declare(strict_types=1);
+
 $namespaceDefinition
 use $useClassName;
 
-class $className extends $baseClassName
+final class $className extends $baseClassName
 {
     /**
      * Change Method.
@@ -15,7 +17,7 @@ class $className extends $baseClassName
      * Remember to call "create()" or "update()" and NOT "save()" when working
      * with the Table class.
      */
-    public function change()
+    public function change(): void
     {
 
     }
--- a/src/Phinx/Migration/Migration.template.php.dist
+++ b/src/Phinx/Migration/Migration.template.php.dist
@@ -1,6 +1,5 @@
 <?php
 declare(strict_types=1);
-
 $namespaceDefinition
 use $useClassName;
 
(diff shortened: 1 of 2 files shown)


dereuromark:
I don't think adding the strict types now provides any real measurable effect at this point for the library or the project use of it.

More importantly, we probably want to get phpstan to level 6, and make sure we start to add more typehints in general in the next majors probably.

Bilge:
It doesn't hurt to add the strict types requirement for migration files now.

src/Phinx/Migration/Migration.template.php.dist
```diff
@@ -1,8 +1,9 @@
 <?php
+declare(strict_types=1);
```
dereuromark, line 2: Are we losing a newline here?

dereuromark:
Fair enough.
If everyone else is fine with this.

src/Phinx/Migration/Migration.template.php.dist
Bilge: Yes. But it looks correct because `$namespaceDefinition` appears to always be empty. If it is possible to populate that variable, the code should also prefix and suffix a blank line around the namespace definition. Of course, it is not possible to add such logic to this template and [I don't know how the code surrounding namespaces actually works](https://github.com/cakephp/phinx/issues/1820) (if it does at all).

MasterOdin: This works fine as if `$namespaceDefinition` is defined, it bookends the string with `PHP_EOL`s

Bilge:
Can we get a tag for this?

dereuromark:
https://github.com/cakephp/phinx/releases/tag/0.12.3

Issue #164, opened 2013-12-20
Error reporting for create


michaeljs1990:
on creating a new migration is wants you to use CamelCase but is really requiring you to use PascalCase

robmorgan:
Closing this - PascalCase is technically correct, however when I present Phinx more people understand camel case.

kafoso:
> Closing this - PascalCase is technically correct, however when I present Phinx more people understand camel case.

But it's wrong... Just because a great number of people fail to educate themselves that doesn't make things acceptable.

dereuromark:
@kafoso Maybe a name that isnt visually clear isn't the best way to describe it then?
CamelCase and camelBacked fit better because they make it crystal clear what is meant. Or would you also know what HenryCase, RobertCase and ElliCase could mean?

kafoso:
PHP class names are Pascal Case (https://techterms.com/definition/pascalcase). Or rather, as per PSR-1 (https://www.php-fig.org/psr/psr-1/), "StudlyCaps".

Camel case should/must always begin with a lowercase letter.

Developers then write the class name in camel case and then get shouted at for not using camel case...

```sh
$ vendor/bin/phinx create sorryWhat
```

```
In Create.php line 169:

  The migration class name "sorryWhat" is invalid. Please use CamelCase format.

```

PR #1825, opened 2020-06-25
Correct usage of CamelCase to PascalCase

MasterOdin:
The codebase uses the term `CamelCase` to reference the titling of migrations and seeds and such, though as pointed out in #164, this is incorrect as camel case is that middle words are capitalized with a lowercase start (`camelCase`) and that what phinx is actually using is `PascalCase`. I agree with @kafoso in https://github.com/cakephp/phinx/issues/164#issuecomment-649309841 that just because there's people who incorrectly mistake the two, phinx should not and should use the proper one.

### DIFF
--- a/docs/en/commands.rst
+++ b/docs/en/commands.rst
@@ -42,7 +42,7 @@ The Create Command
 
 The Create command is used to create a new migration file. It requires one
 argument: the name of the migration. The migration name should be specified in
-CamelCase format.
+PascalCase format.
 
 .. code-block:: bash
 
@@ -208,7 +208,7 @@ The Seed Create Command
 
 The Seed Create command can be used to create new database seed classes. It
 requires one argument, the name of the class. The class name should be specified
-in CamelCase format.
+in PascalCase format.
 
 .. code-block:: bash
 
--- a/docs/es/commands.rst
+++ b/docs/es/commands.rst
@@ -23,7 +23,7 @@ Abre este archivo en tu editor de texto para configurar tu projecto. Por favot m
 Comando de creación.
 --------------------
 
-El comando de creación se usa para crear un nuevo archivo de migración. Requiere un argumento: el nombre de la migración. El nombre de la migracion debera especificarse en el formato CamelCase:
+El comando de creación se usa para crear un nuevo archivo de migración. Requiere un argumento: el nombre de la migración. El nombre de la migracion debera especificarse en el formato PascalCase:
 
 .. code-block:: bash
 
@@ -160,7 +160,7 @@ Los puntos de interrupción son visibles cuando ejecuta el comando de estado.
 "Database Seeding"
 ------------------
 
-El comando Crear semilla se puede usar para crear nuevas clases de base de datos. Requiere un argumento, el nombre de la clase. El nombre de la clase debe especificarse en formato CamelCase.
+El comando Crear semilla se puede usar para crear nuevas clases de base de datos. Requiere un argumento, el nombre de la clase. El nombre de la clase debe especificarse en formato PascalCase.
 
 .. code-block:: bash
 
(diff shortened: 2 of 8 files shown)


dereuromark:
CamelCase vs camelBacked is the default words here.

MasterOdin:
The first sentence from wikipedia article on [camel case](https://en.wikipedia.org/wiki/Camel_case):
> Camel case (stylized as camelCase; also known as camel caps or more formally as medial capitals) is the practice of writing phrases such that each word or abbreviation in the middle of the phrase begins with a capital letter, with no intervening spaces or punctuation. Common examples include "iPhone" and "eBay". It is also sometimes used in online usernames such as "johnSmith", and to make multi-word domain names more legible, for example in advertisements.

> CamelCase vs camelBacked is the default words here.

Is this a CakePHP convention?

dereuromark:
I think it is a very basic and general convention also documented all over the internet
Most people don't know what Pascal case means or is supposed to be. CamelCase and camelBacked is an agnostic and visual way to describe it and thus a good default anyway.
It is also in sync with frameworks like CakePHP, and since this is mainly coupled with it (or CakePHP Migrations using Phinx and Phinx using CakePHP) it makes sense to not confuse people already used to those very good default style names.
https://en.wikipedia.org/wiki/Camel_case lists PascalCase as well, but as I said: It has less visual and thus less useful meaning.

kafoso:
Generally, when people encounter something they don't understand, they tend to Google things instead of just giving up. So the argument that people don't know the term "Pascal case" is a bit vague, in my oppinion.

Alternatively, use the definition "StudlyCaps", which is defined in PSR-1 (https://www.php-fig.org/psr/psr-1/).

MasterOdin:
PSR-12 (https://www.php-fig.org/psr/psr-12) contains this line:

> The term ‘StudlyCaps’ in PSR-1 MUST be interpreted as PascalCase where the first letter of each word is capitalized including the very first letter.

PSR-1 differentiates `StudlyCaps` from `camelCase`:

> Method names MUST be declared in `camelCase`.

rquadling:
FYI, we have also seen the term lowerCamelCase, UpperCamelCase, and TitleCamelCase.

kafoso:
> FYI, we have also seen the term lowerCamelCase, UpperCamelCase, and TitleCamelCase, and https://wiki.c2.com/?LowerCamelCase

Camel case gets its name from the capital letters representing the hunches on a camel's back. This hardly needs explaining. "UpperCamelCase" is just stupid in this regard. It would result in an abomination, with the camel having a large hunch on either its head or rear.

#quitstupidlingo

rquadling:
>:D > #quitstupidlingo

I'd add the term 'StudlyCaps' to that too. For some reason I get the whole Monty Python Lumberjack song going mixed with Desperate Dan from The Beano comic. It is just a super-daft term.

rquadling:
Seems all the humps and lumps are in the right place to me!!!
image

kafoso:
> Seems all the humps and lumps are in the right place to me!!!

That's a head.

And visually, it's the wrong way. "camelCasE"?

I planned on doing a FIFY Photoshop of the image, flipping it and putting a hunch on its head for maximum ridiculousness, but this has already gone too far.

The point is clear.

saeideng:
> It is also in sync with frameworks like CakePHP, and since this is mainly coupled with it 

:+1: 

my opinion is using current word ( `CamelCase` ) instead of `PascalCase`

dereuromark:
We are using the same terms as also defined in https://docs.zendframework.com/zend-filter/word/ for years.
And follow the main docs on https://book.cakephp.org/4/en/contributing/cakephp-coding-conventions.html#naming-convention
Lets not further debate this and instead focus on the functional changes and issues that matter.

PR #1826, opened 2020-06-25
Add support for BinaryUUID to SqlServer

### LINKED ISSUE #1746: Add binaryuuid type support to sql server
From https://github.com/cakephp/phinx/pull/1734#issuecomment-612300523, https://github.com/cakephp/phinx/pull/1734#issuecomment-612337231

> For SQLServer, binaryuuid would probably map to the same type as static::PHINX_TYPE_UUID, namely uniqueidentifier, though like the postgresql adapter, you would not be able to do a comparison against static::PHINX_TYPE_BINARYUUID on saved columns in a table.

The `static::PHINX_TYPE_BINARYUUID` was added to MySQL, PostgreSQL, and SQLite. Needs to be added to SQL Server, this issue acting as a reminder for that.

MasterOdin:
Continuation of #1734
Closes #1746 

Adds `PHINX_TYPE_BINARYUUID` to SqlServer in a similar fashion to Postgres where it aliases to the same thing as `PHINX_TYPE_UUID` which is already a binary uuid field (`uniqueidentifier` in the case of SqlServer).

The tests here mimic the ones added for the other adapters from the above PR.

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -34,6 +34,7 @@ class SqlServerAdapter extends PdoAdapter
      */
     protected static $specificColumnTypes = [
         self::PHINX_TYPE_FILESTREAM,
+        self::PHINX_TYPE_BINARYUUID,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


MasterOdin:
@dereuromark it looks like coveralls does not get a report of the test coverage for sql server from appveyor at all, so not sure how much I can do to fix that particular failure.

PR #1828, opened 2020-06-30
Consistent output of shortened file paths

Roy-Orbison:
as per commit 2b7036ec684337a69d33d794b6f9652656bfde84

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -113,7 +113,7 @@ abstract class AbstractCommand extends Command
         $this->loadManager($input, $output);
 
         if ($bootstrap = $this->getConfig()->getBootstrapFile()) {
-            $output->writeln('<info>using bootstrap</info> .' . str_replace(getcwd(), '', realpath($bootstrap)) . ' ');
+            $output->writeln('<info>using bootstrap</info> ' . Util::relativePath($bootstrap) . ' ');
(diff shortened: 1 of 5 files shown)


src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -113,7 +113,7 @@ public function bootstrap(InputInterface $input, OutputInterface $output)
         $this->loadManager($input, $output);
 
         if ($bootstrap = $this->getConfig()->getBootstrapFile()) {
-            $output->writeln('<info>using bootstrap</info> .' . str_replace(getcwd(), '', realpath($bootstrap)) . ' ');
```
dereuromark, line 116: I wonder if we should move the realpath() part also into that new method.

Roy-Orbison: I'm not against it. If it were used everywhere `relativePath` is, I'd have included it.

dereuromark: Sure, go ahead.

dereuromark:
ping @Roy-Orbison

Roy-Orbison:
I'll get to it.

dereuromark:
Not a fan of Yoda ;) But I guess this is fair enough for now
Thanks a lot for the improvement!

Roy-Orbison:
Not full Yoda, no `false === ...`.

More like spoken language to code: "Does the directory in question start with the current directory?"

PR #1830, opened 2020-07-02
Add support for native Mysql blob types and fallback for binary

liqueurdetoile:
Hi,

This PR came from discussion here : https://github.com/cakephp/migrations/issues/359

There's no breaking change.

The extended behavior is the ability with `MysqlAdapter` to :

- use `binary` or `varbinary` type with any length and get the best fitted column type
- directly use `tinyblob`, `mediumblob` or `longblob` with no limit to get this column type and right length settled

In case of conflict between required type and limit, the limit is used to select the best fitted column type.

Aside this, I've also fix some tests that were relying on hard coded db name and CR/LF issues for string comparisons

Hope it can be useful and that I've not mistaken any guideline :smile:


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -922,28 +925,57 @@ public function getSqlType($type, $limit = null)
 
                 return ['name' => 'text'];
             case static::PHINX_TYPE_BINARY:
-                return ['name' => 'binary', 'limit' => $limit ?: 255];
+                if (is_null($limit)) {
```
dereuromark, line 928: `=== null` better

liqueurdetoile: Miss that one

dereuromark:
There seems to be a conflict.

liqueurdetoile:
There's no conflict in my opinion. Previous commit simply replace hard coded db name with another one. There's no way to have a `phinx_testing` nor a `phinx` database settled in these tests apart config DSN config setup in `phpunit.xml`.

Im' pretty confident that my replacements should be kept as it uses DSN db name and will work in any setup.

othercorey:
It isn’t an opinion. git just requires merging master to your branch because of edits in the same file.

liqueurdetoile:
Not really, you've switched back to hard coded db name that will make tests fails in local dev setup if not choosing 'phinx' as db name.

dereuromark:
> Conflicting files
tests/Phinx/Db/Adapter/MysqlAdapterTest.php

Maybe you didnt make the changes on top of latest master?
You might want to rebase.

liqueurdetoile:
Conflict is from a PR merged after mine was submitted. I propose to leave hard-coded way :smile:

othercorey:
The merge looks good. It should use the new config array like you have it.

src/Phinx/Db/Adapter/MysqlAdapter.php
dereuromark:
```suggestion
                if ($limit === null) {
```

src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1103,9 +1136,19 @@ public function getPhinxType($sqlTypeDef)
                 $limit = static::TEXT_LONG;
                 break;
             case 'binary':
+                if (is_null($limit)) {
```
dereuromark, line 1139: Same here

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -890,6 +892,102 @@ public function testTinyTextColumn()
...
+    /** @dataProvider varbinaryToBlobAutomaticConversionData */
+    public function testVarbinaryToBlobAutomaticConversion(?int $limit = null, string $expectedType, int $expectedLimit)
+    {
+        $table = new \Phinx\Db\Table('t', [], $this->adapter);
+        $table->addColumn('column1', 'varbinary', ['limit' => $limit])
+              ->save();
+        $columns = $table->getColumns();
+        $sqlType = $this->adapter->getSqlType($columns[1]->getType(), $columns[1]->getLimit());
+        $this->assertEquals($expectedType, $sqlType['name']);
+        $this->assertEquals($expectedLimit, $columns[1]->getLimit());
```
dereuromark, line 946: Can we please use assertSame() everywhere to make sure the types are also as expected?

liqueurdetoile: Sure. Should I also switch all others assertEquals ? I've simply mirrored pattern found in test file.

dereuromark: If you want, sure.
We can also do that in a separate PR for other code, though.

dereuromark:
Some errors seem left:

```
There were 4 errors:

1) Test\Phinx\Db\Adapter\MysqlAdapterTest::testCreateTableWithForeignKeys
Undefined index: table_name

/home/runner/work/phinx/phinx/tests/Phinx/Db/Adapter/MysqlAdapterTest.php:207

2) Test\Phinx\Db\Adapter\MysqlAdapterTest::testAddPrimarykey
Undefined index: constraint_name

...
```

othercorey:
Those failures started with mysql 8.0.21 in other PRs.

dereuromark:
Weird, master is green ( https://travis-ci.com/github/cakephp/phinx/builds/176148908 )
Maybe this PR might need another merge of master?

liqueurdetoile:
Thx @othercorey 

I was wondering how the hell it may have been related to this PR.

othercorey:
This only started with the mysql 8.0.21 release. You can see my draft PR showing 8.0.20 builds.

dereuromark:
Merging then, and we figure out to solve travis in the separate PR.

PR #1833, opened 2020-07-10
Decimal support for sqlite adapter

cnizzardini:
This PR adds support for `decimal` to the SQLITE adapter. It looks like there was a clear intent to not support this by the author, however, I see no reason for this. I encountered the following when writing unit tests for a custom data seeder: 

> InvalidArgumentException: An invalid column type "decimal" was specified for column "rental_rate". in /var/www/personal/cakephp-sakila-plugin/vendor/robmorgan/phinx/src/Phinx/Db/Table.php:305

This PR adds the `decimal` type and alters the unit tests to allow decimal to pass. I tested this with my local project and was able to browse the SQLite database with DBeaver GUI client.

Reference:
https://www.sqlite.org/datatype3.html


dereuromark:
That was easy :)
:+1:

PR #1834, opened 2020-07-17
Updated travis config to only run code coverage

othercorey:
With github actions setup, this reduces the configs run in travis similar to cakephp.

Adds php nightly build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,9 @@ services:
   - mysql
 
 addons:
-  postgresql: "9.6"
+  postgresql: '9.6'
 
 php:
-  - 7.2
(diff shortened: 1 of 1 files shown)


othercorey:
@ADmad Can you take a look?

MasterOdin:
Why not do the code coverage reporting from the GH actions?

ADmad:
@MasterOdin Because the coverage services require passing a token, which is saved in GH secrets and GH secrets are not accessible for pull requests made by users who don't have commit access to the repo.

othercorey:
I added postgresql 9.6 config back since we should try to test lowest supported.

PR #1836, opened 2020-07-17
Run with mysql:8.0.20

othercorey:


dereuromark:
Shall we merge this then for now?

othercorey:
Well, something in the `getPrimaryKey` fails in the very latest mysql8. I don't know why, but I don't know if restricting the test version is good long-term.

It's possible that 8.0.21 is simply buggy and so should be ignored -- but then we'll miss 8.0.22.

dereuromark:
We can open a separate matrix build with the higher one and keep it allow_failure for now until we find more details on how to solve it
But having red master and red PRs is not helpful either for now.

othercorey:
No objections. Just explaining the little I found out.

PR #1759, opened 2020-05-07
fix failing tests for mysql 8

MasterOdin:
Closes #1751

MySQL 8.0.17 deprecated usage of the display width for integer data types, stating that it did not really do anything as it would allow integers to go beyond the specified width, and was only really used when you wanted to pad a number. The tests have been modified now to check on version of mysql to determine what the expected type for integer columns will be, but a more proper (and breaking) fix would be to probably not bother with display widths at all for MysqlAdapter.

PR #1837, opened 2020-07-19
Use capitalized references to INFORMATION_SCHEMA for mysql

MasterOdin:
This fixes the failing tests under MySQL 8.0.21+. As part of MySQL 8, the INFORMATION_SCHEMA tables were supposed to always return their columns as capitalized always (see . I guess pre 8.0.21, some tables would still return lowercase if the query was lowercase. This is related to changes done in #1759 along this same vein.

This should render #1836 unnecessary, and can leave phinx running against the latest version of mysql.


othercorey:
Looks good.

Issue #1840, opened 2020-07-24
SQLite Adapter does not create foreign key with custom name


charlienux:
**Migration Instructions:**
```
$table = $this->table('tbl_child');
$table->addColumn('column1', 'integer')
          ->addColumn('column2', 'integer')
          ->addColumn('column3', 'string', ['limit' => 40])
          ->addColumn('master_id', 'integer')
          ->addIndex(['column2'])
          ->addIndex(['column1', 'column2'], ['unique' => true, 'name' => 'uq_tbl_child_column1_column2_ndx'])
          ->addForeignKey(
              'master_id', 'tbl_master', 'id',
              ['delete'=> 'NO_ACTION', 'update'=> 'NO_ACTION', 'constraint' => 'fk_master_id' ]
          )->create();
```

**Result:**
```
-- createTable('tbl_child') 
CREATE TABLE tbl_child (
    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
   column1 INTEGER NOT NULL,
   column2 INTEGER NOT NULL,
   column3 VARCHAR(40) NOT NULL,
   master_id INTEGER NOT NULL); 

CREATE INDEX tbl_child_column2_index ON tbl_child (column2 ASC); 

CREATE UNIQUE INDEX uq_tbl_child_column1_column2_ndx ON tbl_child (column1 ASC, column2 ASC);
-> 0,0008s -- Altering table tbl_child

pragma foreign_keys = ON; 
CREATE TABLE tmp_tbl_child (
    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
    column1 INTEGER NOT NULL,
    column2 INTEGER NOT NULL,
    column3 VARCHAR(40) NOT NULL,
    master_id INTEGER NOT NULL,
    CONSTRAINT fk_master_id);

INSERT INTO tmp_tbl_child(id, column1, column2, column3, master_id)

SELECT id, column1, column2, column3, master_id FROM tbl_child;

DROP TABLE tbl_child;

ALTER TABLE tmp_tbl_child RENAME TO tbl_child;
```

### The SQL statement is incomplete
CONSTRAINT `fk_master_id`

PR #1842, opened 2020-07-25
Fixes problems in SQLite Adapter in foreign key creation

charlienux:
This PR fixes two problems #1840 and #1841.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,9 @@ services:
   - mysql
 
 addons:
-  postgresql: "9.6"
+  postgresql: '9.6'
 
 php:
-  - 7.2
   - 7.4
 
 env:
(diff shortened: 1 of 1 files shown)


dereuromark:
It looks like there is some error left here in the tests.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -47,7 +47,6 @@ class SQLiteAdapter extends PdoAdapter
         self::PHINX_TYPE_CHAR => 'char',
         self::PHINX_TYPE_DATE => 'date_text',
         self::PHINX_TYPE_DATETIME => 'datetime_text',
-        self::PHINX_TYPE_DECIMAL => 'decimal',
```
dereuromark, line 50: Are you sure we need to remove this type here? It can have other side effects and regressions.

charlienux: Indeed that line of code should not be removed.

I reversed the change.

charlienux:
> It looks like there is some error left here in the tests.

Errors in the tests are due to the line removed by mistake.

PR #1849, opened 2020-08-03
allow host parameter for postgres to be optional

### LINKED ISSUE #1832: PostgreSQL & UNIX sockets
It appears that you are unable to use the default UNIX socket connection with PostgreSQL because https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/PostgresAdapter.php#L64 forces the inclusion of the `host` parameter

MasterOdin:
Closes #1832

Allows the `host` parameter for a postgres connection to be optional. If not defined, it will fallback to attempting to use a socket connection for PDO.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
   global:
     - CODECOVERAGE=1
   matrix:
-    - PGSQL_DSN='pgsql://postgres@127.0.0.1/phinx'
+    - PGSQL_DSN='pgsql://postgres@127.0.0.1/phinx' POSTGRES_SOCKET=true
     - MYSQL_DSN='mysql://root@127.0.0.1/phinx' MYSQL_UNIX_SOCKET='/var/run/mysqld/mysqld.sock'
     - SQLITE_DSN='sqlite:///phinx'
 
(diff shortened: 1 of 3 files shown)


tests/Phinx/Db/Adapter/PostgresAdapterTest.php
```diff
@@ -110,6 +110,20 @@ public function testConnectionWithInvalidCredentials()
...
 
+    public function testConnectionWithSocketConnection()
+    {
+        if (!getenv('POSTGRES_SOCKET')) {
+            $this->markTestSkipped('Postgres socket connection skipped.');
+        }
+
+        $options = PGSQL_DB_CONFIG;
+        unset($options['host']);
+        $adapter = new PostgresAdapter(PGSQL_DB_CONFIG, new ArrayInput([]), new NullOutput());
```
othercorey, line 121: This doesn't use the socket config.

.travis.yml
```diff
@@ -16,7 +16,7 @@ env:
   global:
     - CODECOVERAGE=1
   matrix:
-    - PGSQL_DSN='pgsql://postgres@127.0.0.1/phinx'
+    - PGSQL_DSN='pgsql://postgres@127.0.0.1/phinx' POSTGRES_SOCKET=true
```
othercorey, line 19: Can we change this to `POSTGRES_TEST_SOCKETS`?

PR #1854, opened 2020-08-07
Use Interface constants for function names

MasterOdin:
This replaces the usage of the named function calls to a migration with the constants from MigrationInterface and calls to a seed with the constants from SeedInterface. This helps make the bind between them a bit more stable, in the rare case that the values of these constants were to change.

### DIFF
--- a/src/Phinx/Migration/Manager/Environment.php
+++ b/src/Phinx/Migration/Manager/Environment.php
@@ -80,8 +80,7 @@ class Environment
         $startTime = time();
         $migration->setAdapter($this->getAdapter());
         if (method_exists($migration, MigrationInterface::INIT)) {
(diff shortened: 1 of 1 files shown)


dereuromark:
I see what you mean, but now we lost IDE and stan analyses of it, no?
Before one could follow the method into the actual code, now it seems like this is not possible anymore.

MasterOdin:
The functions don't exist in the Abstract(Migration|Seed) or MigrationInterface so there's nothing to link to for an IDE even if it did work, hence the annotations above each line in master (which only works in JetBrains IDEs). In VSCode for example, I get greeted with when looking at the those lines:

Screenshot 2020-08-14 11 11 15

PR #1858, opened 2020-08-10
use quoteTableName when caching table name operations

MasterOdin:
This rewrites the `createdTables` cached to use the quoted table names, instead of just the table names as provided to the function. This fixes an issue where if a user on postgres created a table `table1` and then did `hasTable('public.table1')`, the cache would be missed as it would contain just `table1`, even though both things refer to the same thing. Interestingly, this showed an issue where the schema table (phinxlog) was being added somewhat inconsistently to the cache such that on dropping the database / schema, and then recreating it, the phinxlog table was entering the cache in slightly different ways, and so had to clear the cache on the reset of things (which makes sense anyway) or else the phinxlog table would complain about not existing when the system attempted to refer to it.

### DIFF
--- a/src/Phinx/Db/Adapter/AbstractAdapter.php
+++ b/src/Phinx/Db/Adapter/AbstractAdapter.php
@@ -365,6 +365,7 @@ abstract class AbstractAdapter implements AdapterInterface
      */
     protected function addCreatedTable($tableName)
     {
+        $tableName = $this->quoteTableName($tableName);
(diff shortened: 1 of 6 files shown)


MasterOdin:
For coveralls, it might be a good idea to maybe use the windows build of Travis to test sqlserver and upload the results to coveralls, so that PRs that modify any lines in the SQLServerAdapter do not fail that check always.

PR #1859, opened 2020-08-10
update gh action to run on all branches

MasterOdin:
Updates the GH action workflow to run on all branches. As is, for in-progress work of mine, I cannot use GH actions to see if the test suite would pass or not, instead having to create a PR on my own repo to get the results. This greatly diminishes the valuableness of GH to me before I try opening a PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Phinx CI
 on:
   push:
     branches:
-      - master
+      - '*'
   pull_request:
     branches:
       - '*'


othercorey:
@ADmad Phinx has other odd CI behavior. AppVeyor runs a push and pr build for everything although it seems to have the same settings as cakephp.

ADmad:
@othercorey You don't want AppVeyor to run for every push/pr?  Personally I gave up on dealing with appveyor config long time ago.

If dealing with appveyor is problematic then the builds for windows/sqlserver could be setup on Github Actions itself.

othercorey:
I just don't understand why it runs differently than cakephp.

PR #1871, opened 2020-08-23
add createSchema and dropSchema methods on MigrationInterface

### LINKED ISSUE #1870: Add schema creation to migration API
Migrations can create databases and tables but not schemas. It is also not valid to create a schema implicitly using the create table API, e.g. `$this->table('foo.bar')->create()`.

>PDOException: SQLSTATE[3F000]: Invalid schema name: 7 ERROR:  schema "foo" does not exist

MasterOdin:
Closes #1870 

Given that there's no docs right now for `createDatabase` either, did not elect to add them here (as not sure where to place them). For now, it should at least make it a bit easier for user discovery.

PR made against 0.13 due to breaking change on the MigrationInterface

### DIFF
--- a/src/Phinx/Migration/MigrationInterface.php
+++ b/src/Phinx/Migration/MigrationInterface.php
@@ -218,6 +218,28 @@ interface MigrationInterface
     public function dropDatabase($name);
 
     /**
+     * Creates schema.
+     *
+     * This will thrown an error for adapters that do not support schemas.
+     *
+     * @param string $name Schema name
+     * @return void
+     * @throws \BadMethodCallException
+     */
+    public function createSchema($name);
+
+    /**
+     * Drops schema.
+     *
+     * This will thrown an error for adapters that do not support schemas.
+     *
+     * @param string $name Schema name
+     * @return void
+     * @throws \BadMethodCallException
+     */
+    public function dropSchema($name);
+
+    /**
      * Checks to see if a table exists.
      *
      * @param string $tableName Table name
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -262,6 +262,22 @@ abstract class AbstractMigration implements MigrationInterface
     /**
      * @inheritDoc
      */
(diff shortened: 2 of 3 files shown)


src/Phinx/Migration/MigrationInterface.php
```diff
@@ -217,6 +217,28 @@ public function createDatabase($name, $options);
...
+    /**
+     * Creates schema.
+     *
+     * This will thrown an error for adapters that do not support schemas.
+     *
+     * @param string $name Schema name
+     * @return void
+     * @throws \BadMethodCallException
+     */
+    public function createSchema($name);
```
dereuromark, line 229: If we add those freshly, couldnt we add the typehints? Or should we keep them off to be in sync with the other methods?

MasterOdin: I would say that the functions type hinting within the interface be consistent throughout the file.

I do support in 0.13 to make the BC break here and add the type hints throughout the interface, especially given the larger BC break in #1872 already potentially going into 0.13. Alternatively, could hold off till 0.14.

dereuromark: Yeah, makes sense 
What do others think?

dereuromark: So merging now, but I agree that 0.13 could and should add those typehints.

Bilge:
It seems this is not available in seed files.

MasterOdin:
The `createDatabase` and `dropDatabase` methods are also not available on the `SeedInterface`.

Just for my own edification, what's the use-case on wanting to create / drop schemas as part of seeding, and not the migration process?

Bilge:
See #2161.

Aside, I don't understand why seeding and migration don't share the same API.

MasterOdin:
For #2161, the end result of what you'd want is that:
1. phinx itself doesn't care about the migration table's existence when running seeds
2. that you would have a method to easily create the schema for the phinx migration table

> Aside, I don't understand why seeding and migration don't share the same API.

Agreed now that I've looked at them more closely.

PR #1872, opened 2020-08-26
Set column null by default unless identity

### LINKED ISSUE #1817: Phinx default for null columns contravenes DBMS default
By default, Phinx creates columns as `NOT NULL`, but the default in Postgres is `NULL`.  Consider the following query.

```sql
ALTER TABLE public.foo
    ADD COLUMN bar timestamp with time zone;
```

This query creates the column `bar` but does not specify any `NULL` options. By default it creates the column permitting nulls. In contrast, the following Phinx query will do the opposite:

```php
$this->table('foo')->addColumn('bar', 'timestamp')->update();
```

Not only does this betray expectations, it also causes errors if there's existing rows.

>PDOException: SQLSTATE[23502]: Not null violation: 7 ERROR:  column "bar" contains null values

I believe these defaults should be aligned to avoid unpleasant surprises.

MasterOdin:
Closes #1817 

This makes it so that newly added columns are `NULL` by default, instead of `NOT NULL` except in the case of a column that uses the `identity` option, and does not also set a `null` option.

This is definitely BC breaking, and the release notes should make a big note that all previously written migrations will have to be updated to take into account the new null default, unless the authors were already explicit about the null status within their migrations.

### DIFF
--- a/src/Phinx/Db/Adapter/AbstractAdapter.php
+++ b/src/Phinx/Db/Adapter/AbstractAdapter.php
@@ -312,7 +312,7 @@ abstract class AbstractAdapter implements AdapterInterface
             ];
 
             $table = new Table($this->getSchemaTableName(), $options, $this);
-            $table->addColumn('version', 'biginteger')
+            $table->addColumn('version', 'biginteger', ['null' => false])
                 ->addColumn('migration_name', 'string', ['limit' => 100, 'default' => null, 'null' => true])
                 ->addColumn('start_time', 'timestamp', ['default' => null, 'null' => true])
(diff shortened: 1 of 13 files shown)


dereuromark:
Sounds good to me. 
Any objections from others?

sitedyno:
#1718 :ping_pong:

Bilge:
Looks like you missed https://github.com/cakephp/phinx/blob/2922b17953f4a0910efb4ba5ae7cb528eea5718e/src/Phinx/Db/Adapter/AbstractAdapter.php#L314

PR #1873, opened 2020-08-26
run preFlightCheck and postFlightCheck immediately after migration

### LINKED ISSUE #1818: Forgetting to call save()/create()/update() leaves migration log in a bad state
Writing a simple Phinx query but forgetting to call one of the `save()/create()/update()` leaves the migration log in an invalid state. The migration will fail with an error similar to the following.

>RuntimeException: Migration has pending actions after execution! 

However, Phinx will still write to the migration log as if the migration completed successfully. Subsequent migrations will not run the migration again, because it thinks it has already completed. Subsequent rollbacks will fail because the changes do not exist.

MasterOdin:
Closes #1818

This makes it so that the `preFlightCheck` and `postFlightCheck` are now run immediately before and immediately after the migration is run. This gives the advantage that one could run stuff against the adapter if they wanted to in the preFlightCheck, and that the `postFlightCheck` will run before the migration is logged as having run. This is especially beneficial in the latter's case given that by default, it will throw an exception if the user has incomplete table operations.

While this does slightly change the order things and run in, and that the checks are run within the migration timer, it's not really BC breaking enough to warrant necessarily going in the 0.13.x line only.

### DIFF
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -376,14 +376,12 @@ class Manager
             ' <info>' . $migration->getVersion() . ' ' . $migration->getName() . ':</info>' .
             ' <comment>' . ($direction === MigrationInterface::UP ? 'migrating' : 'reverting') . '</comment>'
         );
-        $migration->preFlightCheck($direction);
 
         // Execute the migration and log the time elapsed.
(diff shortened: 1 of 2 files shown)


dereuromark:
Sounds reasonable to me.

PR #1875, opened 2020-09-01
make utf8mb4_unicode_ci default collation for new mysql tables

### LINKED ISSUE #1763: Move to utf8mb4_unicode_ci for default MySQL collation
A proposal for Phinx 0.13.x release, as this is a breaking change.

Currently, the default collation for a table is `utf8_general_ci` (https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/MysqlAdapter.php#L243). This is a legacy collation that has a broken utf8 implementation that should be avoided. While `utf8mb4_general_ci` does exist and is "faster" as it takes various shortcuts on sorting (which can resort in wrong sorts in a number of cases), this speed-up is almost negligible on any modern machine that there's no reason to suggest it. As such, it should be that `utf8mb4_unicode_ci` is used as the default and then for anyone who wants something different, an override option is provided.

MasterOdin:
closes #1763

This makes `utf8mb4_general_ci` the default collation for MySQL tables over `utf8_general_ci`. The prior default made sense 7+ years ago when computing power for servers was more limited, but nowadays, the performance gained is increasingly minimal over the necessary deal of handling the greater nuances of capitalization of different languages and their characters by default that the original `utf8` schema lacked for MySQL.

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -170,8 +170,8 @@ specified under the ``environments`` nested collection. For example:
             user: root
             pass: ''
             port: 3306
-            charset: utf8
-            collation: utf8_unicode_ci
+            charset: utf8mb4
+            collation: utf8mb4_unicode_ci
 
 would define a new environment called ``production``.
 
--- a/docs/en/migrations.rst
+++ b/docs/en/migrations.rst
@@ -478,7 +478,7 @@ Option     Description
 comment    set a text comment on the table
 row_format set the table row format
 engine     define table engine *(defaults to ``InnoDB``)*
-collation  define table collation *(defaults to ``utf8_general_ci``)*
+collation  define table collation *(defaults to ``utf8mb4_unicode_ci``)*
 signed     whether the primary key is ``signed``  *(defaults to ``true``)*
 ========== ===========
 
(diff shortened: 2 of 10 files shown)


MasterOdin:
Something I failed to think about within the original PR is that for a primary key within InnoDB for MySQL, the max number of bytes for 5.7 is 255 chars for utf8 (which use 3 bytes per character) and 191 for utf8mb4. This makes transitioning a bit more difficult for the default case of something like:

```php
        $table = $this->table('table1', ['id' => false, 'primary_key' => ['column1']]);

        $table->addColumn('column1', 'string')
            ->addColumn('column2', 'integer')
            ->create();
```

Should the default of string be made `191` instead of `255`? Only if it's the primary key? Require explicitly setting the length for primary key strings?

docs/en/configuration.rst
```diff
@@ -171,7 +171,7 @@ specified under the ``environments`` nested collection. For example:
             pass: ''
             port: 3306
             charset: utf8
-            collation: utf8_unicode_ci
+            collation: utf8mb4_general_ci
```
dereuromark, line 174: Is there a reason not to keep unicode? At least that's the default I am usually using.

othercorey: cakephp defaults to utf8mb4 if that makes a difference here.

garas: Lines above should be 
```yaml
charset: utf8mb4
collation: utf8mb4_unicode_ci
```

See https://stackoverflow.com/a/766996/4228563.

`utf8mb4` supports more characters (e.g. emojis) than `utf8`. 

`unicode` is more universal for sorting and comparison in wide range of languages than `general`.

othercorey: Agree that unicode is best

MasterOdin:
For right now, I'm leaving this open, but not necessarily coming back to it for a period as I would like to re-examine how other migration software handle default limits for string / varchar type, especially as it might concern to primary keys.

dereuromark:
Just use

charset: utf8mb4
collation: utf8mb4_unicode_ci

as commented above and we can merge this

MasterOdin:
@dereuromark done. I decided to just insert a note in the documentation about having to explicitly set the primary key length for MySQL 5.7 and below when using the `string` type with `utf8mb4_unicode_ci`, instead of adjusting the default limit, as I view that as being a much larger BC break as it would probably mean changing the default for all adapters and such to keep them equal. Hopefully 5.7 is fazed out at a decent clip for 8.0+ such that this isn't an issue for too long.

Said note should probably be replicated into the changelog notes as well on release of 0.13 just to increasingly drive awareness of that fact.

PR #1882, opened 2020-09-23
Mysql - Ensure a default of CURRENT_TIMESTAMP(3) is not quoted

rlweb:
Allow for the following:
```
$this->table('exampleFailure', [
	'id' => false,
])
	->addColumn('created_at', 'timestamp', [
		'null' => false,
		'default' => 'CURRENT_TIMESTAMP(3)',
		'limit' => 3,
		'after' => 'displayType',
	])
	->create();
```
which before gave the following "Syntax error or access violation: 1067 Invalid default value for 'created_at'" as it was a quoted string of


dereuromark:
Shouldnt this target master as part of a fix instead of the next major?

Issue #1661, opened 2019-12-17
Index column sort order


JRDuncan:
Phinx does not allow you to specify the sort order on Index columns.

    public function change()
    {
        $table = $this->table('users');
        $table->addColumn('email', 'string')
              ->addColumn('username','string')
              ->addIndex(['email','username'], ['order' => ['email' => 'DESC', 'username' => 'ASC']])
              ->create();
    }

dereuromark:
PR welcome

nohaycubetto:
Second this.

PR #1884, opened 2020-09-25
Index column sort order #1661

JRDuncan:
#1661


Dockerfile
```diff
@@ -22,6 +22,11 @@ RUN docker-php-ext-install \
 # composer
 RUN curl -sS https://getcomposer.org/installer | php && \
 	  mv composer.phar /usr/local/bin/composer
+
+#xdebug
+RUN pecl install xdebug
```
dereuromark, line 27: Why do we need xdebug?

JRDuncan: I'm closing this and opening a different PR

PR #1890, opened 2020-09-29
trigger deprecation notice when using default_database

MasterOdin:
This was deprecated with the 0.12 release, but was only deprecated within the release notes. This adds a deprecation notice into the codebase that is triggered on usage to hopefully help ensure users are not surprised when this does actually get removed.

### DIFF
--- a/tests/Phinx/Config/ConfigTest.php
+++ b/tests/Phinx/Config/ConfigTest.php
@@ -103,6 +103,20 @@ class ConfigTest extends AbstractConfigTest
     }
 
     /**
+     * @covers \Phinx\Config\Config::getDefaultEnvironment
+     */
+    public function testDeprecatedDefaultDatabase()
+    {
+        $configArray = $this->getConfigArray();
+        $configArray['environments']['default_database'] = 'production';
+        $config = new Config($configArray);
+
+        $this->expectDeprecation();
+        $this->expectExceptionMessage('default_database in the config has been deprecated since 0.12, use default_environment instead.');
+        $this->assertEquals('production', $config->getDefaultEnvironment());
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


tests/Phinx/Config/ConfigTest.php
```diff
@@ -95,10 +95,24 @@ public function testReturnsEmptyArrayWithEmptyDataDomain()
...
+     */
+    public function testDeprecatedDefaultDatabase()
     {
         $configArray = $this->getConfigArray();
         $configArray['environments']['default_database'] = 'production';
         $config = new Config($configArray);
+
+        $this->expectDeprecation();
+        $this->expectExceptionMessage('default_database in the config has been deprecated since 0.12, use default_environment instead.');
         $this->assertEquals('production', $config->getDefaultEnvironment());
```
MasterOdin, line 116: It looks like the deprecated warning in the code actually stops phpunit from continuing with testing, and so there's no way to validate that the function actually does return `production` it looks like.

dereuromark: You can check how CakePHP using a function wrapper does it, this seems to work well so far.

MasterOdin: Ah, that makes sense. Added another testcase which does that wrapper to validate the value works as expected, as well as kept existing test to validate the deprecation notice is raised.

othercorey: @MasterOdin You just need to use the deprecated() wrapper here. I believe.

MasterOdin: There is no `deprecated` wrapper available within phinx codebase at the moment. Doing so would probably require adding a new dev dependency to some other cakephp package for this, which is probably not worth while for a singular test.

PR #1899, opened 2020-10-01
make default PKs for mysql unsigned by default

### LINKED ISSUE #1780: Make primary key unsigned by default
Given that #1763 is a breaking change on default options for table creation, it's probably also a good opportunity to also make the primary key column unsigned by default. This was discussed in #657. The [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html) recommend doing it (though strangely enough it's not in their starting example):

> Use the UNSIGNED attribute if possible to allow a greater range. For example, if you use TINYINT, the maximum permissible sequence number is 127. For TINYINT UNSIGNED, the maximum is 255.

MasterOdin:
closes #1780

This changes it so that the default PK for a table (either by not specifying one at all, or just setting the `id` option for the table) is unsigned by default, instead of signed. This gives users extra space to store IDs in their databases, and is generally considered the "right" definition to have for an auto incrementing column for mysql, with some amount of other ORM / DB stuff out there doing this as well.

The principle downside is that foreign key columns will have to be specified with `"signed" => false` now, but I think is balanced out by the upsides of doing this.

### DIFF
--- a/docs/en/migrations.rst
+++ b/docs/en/migrations.rst
@@ -479,11 +479,11 @@ comment    set a text comment on the table
 row_format set the table row format
 engine     define table engine *(defaults to ``InnoDB``)*
 collation  define table collation *(defaults to ``utf8mb4_unicode_ci``)*
-signed     whether the primary key is ``signed``  *(defaults to ``true``)*
+signed     whether the primary key is ``signed``  *(defaults to ``false``)*
 ========== ===========
 
-By default the primary key is ``signed``.
-To simply set it to unsigned just pass ``signed`` option with a ``false`` value:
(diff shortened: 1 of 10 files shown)


othercorey:
> The principle downside is that foreign key columns will have to be specified with "signed" => false now, but I think is balanced out by the upsides of doing this.

This seems like a significant change. Why don't users who want an unsigned range specify it instead of adding this dependency? The default INT range is very large.

dereuromark:
Is there any downside of this?
This seems like the right way to approach this for most keys, if not explicitly configured otherwise.

PR #1904, opened 2020-10-01
add ability to set generic PDO attributes

### LINKED ISSUE #1844: Be able to specify PDO::ATTR_ options for connection
This is a:
* [x] enhancement

* Phinx Version: 0.11.4 hash: 10e8c3f6b09111f4c0ef128d5a43375e15286cc5
* Platform and Target: nginx, MySQL 5.7, PHP 7.2

### What you did
I need to specify `\PDO::ATTR_` options for MySQL connection, but Phinx only allow `\PDO::MYSQL_ATTR_` options.

#### Example
```
\PDO::ATTR_PERSISTENT => true,
\PDO::ATTR_CASE => \PDO::CASE_LOWER,
\PDO::ATTR_ERRMODE => \PDO::ERRMODE_EXCEPTION,
\PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES utf8',
\PDO::MYSQL_ATTR_USE_BUFFERED_QUERY> = false,
```

### What happened
There's no way to do it currently.

### What you expected to happen
I expect to be able to specify  `\PDO::ATTR_` options when connecting to the database. It would also be nice to have it in the docs for both `\PDO::ATTR_` and `\PDO::MYSQL_ATTR_` options.

MasterOdin:
Closes #1844

This allows specifying a list of PDO options that are then set against the connection using the `setAttribute` method. This follows a similar design of the adapter specific options of checking against a specific prefix. This is BC breaking in that it reserves `attr_` for options, but there is no option with that value so it should not affect anyone's code unless they were adding extraneous things to their config.

I've also added documentation on setting these options, both the generic ones and the adapter specific ones with a simple example, as I could not find a mention of it before in the docs.

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -334,7 +334,7 @@ For example, to set the above example options:
 
     $config = [
         "environments" => [
-            "developement" => [
+            "development" => [
                 "adapter" => "mysql",
                 # other adapter settings
                 "attr_case" => \PDO::ATTR_CASE,
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -321,6 +321,31 @@ Phinx currently supports the following database adapters natively:
 * `SQLite <http://www.sqlite.org/>`_: specify the ``sqlite`` adapter.
 * `SQL Server <http://www.microsoft.com/sqlserver>`_: specify the ``sqlsrv`` adapter.
 
+For each adapter, you may configure the behavior of the underlying PDO object by setting in your
+config object the lowercase version of the constant name. This works for both PDO options
+(e.g. ``\PDO::ATTR_CASE`` would be ``attr_case``) and adapter specific options (e.g. for MySQL
+you may set ``\PDO::MYSQL_ATTR_IGNORE_SPACE`` as ``mysql_attr_ignore_space``). Please consult
+the `PDO documentation <https://www.php.net/manual/en/book.pdo.php>`_ for the allowed attributes
+and their values.
+
+For example, to set the above example options:
+
+.. code-block:: php
+
+    $config = [
+        "environments" => [
+            "developement" => [
+                "adapter" => "mysql",
+                # other adapter settings
+                "attr_case" => \PDO::ATTR_CASE,
+                "mysql_attr_ignore_space" => 1,
+            ],
+        ],
+    ];
+
+By default, the only attribute that Phinx sets is ``\PDO::ATTR_ERRMODE`` to ``PDO::ERRMODE_EXCEPTION``. It is
+not recommended to override this.
+
 SQLite
 `````````````````
 
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -87,7 +87,11 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
 
             foreach ($adapterOptions as $key => $option) {
                 if (strpos($key, 'attr_') === 0) {
-                    $db->setAttribute(constant('\PDO::' . strtoupper($key)), $option);
+                    $pdoConstant = '\PDO::' . strtoupper($key);
+                    if (!defined($pdoConstant)) {
+                        throw new \UnexpectedValueException('Invalid PDO attribute: ' . $key . '(' . $pdoConstant . ')');
+                    }
+                    $db->setAttribute(constant($pdoConstant), $option);
                 }
             }
         } catch (PDOException $e) {
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -76,9 +76,20 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
      */
     protected function createPdoConnection($dsn, $username = null, $password = null, array $options = [])
     {
+        $adapterOptions = $this->getOptions();
+
+        if (!isset($adapterOptions['attr_errmode'])) {
+            $adapterOptions['attr_errmode'] = PDO::ERRMODE_EXCEPTION;
+        }
+
         try {
             $db = new PDO($dsn, $username, $password, $options);
-            $db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+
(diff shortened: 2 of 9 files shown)


garas:
Array with real PDO constant keys would be better than error prone string keys.

```php
"attributes" => [
    \PDO::ATTR_CASE => \PDO::CASE_LOWER,
    \PDO::MYSQL_ATTR_IGNORE_SPACE => 1,
],
```

MasterOdin:
I disagree with that. Going that route with using the constant names for keys would lock out JSON and YAML configuration files from using this part of phinx easily as you would be forced to use magic number keys which to me is way worse than strings. For your example, this would be the equivalent JSON object:

```json
{
    "8": 1,
    "1009": 1
}
```

Additionally, it would probably mean that phinx would have to hardcode in the list of attributes for specific drivers so that you don't send `\PDO::MYSQL_ATTR_IGNORE_SPACE` to postgresql driver for example, and the docs say this about avoiding that scenario:

> Using driver-specific attributes with another driver may result in unexpected behaviour.

garas:
I agree now, I was not familiar enough with how adapter specific attributes are handled.

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -76,9 +76,20 @@ protected function verboseLog($message)
      */
     protected function createPdoConnection($dsn, $username = null, $password = null, array $options = [])
     {
+        $adapterOptions = $this->getOptions();
+
+        if (!isset($adapterOptions['attr_errmode'])) {
+            $adapterOptions['attr_errmode'] = PDO::ERRMODE_EXCEPTION;
+        }
```
othercorey, line 83:
```suggestion
        $adapterOptions = $this->getOptions() + [
            'attr_errormode' => PDO::ERRMODE_EXCEPTION;
        ];
```

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -76,9 +76,20 @@ protected function verboseLog($message)
...
+
         try {
             $db = new PDO($dsn, $username, $password, $options);
-            $db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+
+            foreach ($adapterOptions as $key => $option) {
+                if (strpos($key, 'attr_') === 0) {
+                    $db->setAttribute(constant('\PDO::' . strtoupper($key)), $option);
```
othercorey, line 90: We check if the constant exists and throw a useful exception if not.

src/Phinx/Db/Adapter/PdoAdapter.php
othercorey: There doesn't seem to be reason to have this up here. Move it down to where the attributes are used.

src/Phinx/Db/Adapter/PdoAdapter.php
othercorey: It might be easier to simply check if any of the options are valid constants and set them if they are instead of checking for attr_ and validating it.

src/Phinx/Db/Adapter/PdoAdapter.php
MasterOdin: It was put above to minimize the amount of code under the try block.

For your suggestion, just to confirm you want phinx to mandate that `attr_errmode` MUST be set to `\PDO::ERRMODE_EXCEPTION`, even if the user sets it to something else for some reason or other.

src/Phinx/Db/Adapter/PdoAdapter.php
MasterOdin: I still need to make a check on the string prefix, to avoid including `mysql_attr_` and `sqlsrv_attr_` values here, which are considered valid constants, but should not be set within this context.

This is also the same set-up as we use for assigning the `mysql_attr_` and `sqlsrv_attr_` variables within their respective adapters, so you would like to see said exception thrown be added there as well?

src/Phinx/Db/Adapter/PdoAdapter.php
othercorey: No, this is merging the arrays.

src/Phinx/Db/Adapter/PdoAdapter.php
othercorey: I think an exception here is better than a php error.

docs/en/configuration.rst
```diff
@@ -321,6 +321,31 @@ Phinx currently supports the following database adapters natively:
...
+the `PDO documentation <https://www.php.net/manual/en/book.pdo.php>`_ for the allowed attributes
+and their values.
+
+For example, to set the above example options:
+
+.. code-block:: php
+
+    $config = [
+        "environments" => [
+            "developement" => [
```
dereuromark, line 337:
```suggestion
            "development" => [
```

src/Phinx/Db/Adapter/PdoAdapter.php
dereuromark: @othercorey The merge has the wrong order, otherwise looks good.

src/Phinx/Db/Adapter/PdoAdapter.php
MasterOdin: Done

src/Phinx/Db/Adapter/PdoAdapter.php
MasterOdin: May I ask about the preference over array merging for here instead of using `isset`? Is it just personal preference on the part of cakephp folks, or a technical reason for it as well? Just trying to get a handle for future PRs and avoiding repeated review comments.

othercorey: It seems the right order. You want it set only as the default.

We use merging for options defaults when possible because it is easier to add new ones and store the defaults in a property.

dereuromark: Ah ok.

@MasterOdin this is the right operator for assoc array merging, especially for performance and reliability.

MasterOdin: @othercorey fair enough. I will keep that in mind.

Anyway, suggestion committed.

dereuromark:
Looks good.

PR #1905, opened 2020-10-11
Allow setting limit option for big int for mysql

MasterOdin:
This makes `biginteger` work as the other integer types for mysql where it has a default, but you can override it.

Opening this PR here just so I don't forget about this as I noticed this while working on a project that used phinx. Will add a test shortly.


MasterOdin:
@dereuromark ready for review. Tests added not only for `biginteger` but the other two as well.

dereuromark:
Looks good to me.

Issue #1903, opened 2020-10-01
Add support for include clause when adding an index


MasterOdin:
It would be nice for phinx to support the `include` clause for indexes, which is available minimally in [SQLServer](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/create-indexes-with-included-columns?view=sql-server-ver15) and [PostgreSQL](https://www.postgresql.org/docs/11/indexes-index-only-scans.html). For Postgres, having this allows for creation of a "covering index" which is useful for tables where the information does not change often, and where the included columns are part of the payload, but not necessarily part of what you are searching for, giving a nice speed-bump, while separating it from the actual index machinery.

The raw sql is usually something of the form `CREATE INDEX tab_x_y ON tab(x) INCLUDE (y);`.

For phinx, the expected syntax might be something like:

```php
$this->table('users');
    ->addColumn('email', 'string')
    ->addColumn('firstname','string')
    ->addColumn('lastname','string')
    ->addIndex(['email'], ['include' => ['firstname' , 'lastname']])
    ->create();
```

PR #1908, opened 2020-10-16
Add support for include clause when adding an index #1903

JRDuncan:
Resolves https://github.com/cakephp/phinx/issues/1903

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -1249,15 +1249,15 @@ class PostgresAdapter extends PdoAdapter
             }
 
             return $ret;
-        }, $columnNames);	
+        }, $columnNames);
 
         return sprintf(
             'CREATE %s INDEX %s ON %s (%s);',
             ($index->getType() === Index::UNIQUE ? 'UNIQUE' : ''),
             $this->quoteColumnName($indexName),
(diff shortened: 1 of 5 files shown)


src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1282,12 +1282,15 @@ protected function getIndexSqlDefinition(Index $index, $tableName)
             return $ret;
         }, $columnNames);
 
+        $includedColumns = $index->getInclude() ? sprintf(' INCLUDE ([%s])', implode('],[', $index->getInclude())) : '';
```
MasterOdin, line 1285: Given the space in the query below, probably don't need the space before `INCLUDE` here

src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -1251,12 +1251,15 @@ protected function getIndexSqlDefinition(Index $index, $tableName)
             return $ret;
         }, $columnNames);
 
+        $includedColumns = $index->getInclude() ? sprintf(' INCLUDE ("%s")', implode('","', $index->getInclude())) : '';
```
MasterOdin, line 1254: Given the space in the query below, probably don't need the space before `INCLUDE` here

JRDuncan (commented):
Removed unncessary space before "Include"

PR #1910, opened 2020-10-17
Add MySQL FIRST support.

### LINKED ISSUE #655: MysqlAdapter does not support altering a column's position to `FIRST`
[MySQL syntax](https://dev.mysql.com/doc/refman/5.1/en/alter-table.html) supports reordering columns using the `AFTER [column_name]` syntax. Phinx similarly supports this, but _only_ this and not the alternate `FIRST` syntax:

[MysqlAdapter.php L402](https://github.com/robmorgan/phinx/blob/0ccff10/src/Phinx/Db/Adapter/MysqlAdapter.php#L402):

``` php
$sql .= ' AFTER ' . $this->quoteColumnName($column->getAfter());
```

[MysqlAdapter.php L454](https://github.com/robmorgan/phinx/blob/0ccff10/src/Phinx/Db/Adapter/MysqlAdapter.php#L454):

``` php
$after = $newColumn->getAfter() ? ' AFTER ' . $this->quoteColumnName($newColumn->getAfter()) : '';
```

The full MySQL syntax allows for the use of the `FIRST` keyword in place of `AFTER [column_name]` but this is not currently supported in Phinx. There is no way to alter a column to be the "first" column in the table without moving the desired column to the second position, then moving whatever the first column is to the second position. That's a truly ugly prospect.

The above code samples could be improved to recognize a symbolic value, such as `true`, or `FIRST` to indicate that the `ALTER` SQL generated should place the column first. For example:

``` php
protected function afterClause($column)
{
  $after = $column->getAfter();
  if (empty($after)) { // Matches false, empty string, null
    return '';
  } elseif ($after === true) { // Strong comparison! Or $after === 'FIRST' if you prefer.
      return ' FIRST';
  } else {
      return ' AFTER ' . $this->quoteColumnName($after);
  }
}
```

This function could be used (consistently) in place of the existing ternary operators:

``` php
// L402:
$sql .= $this->afterClause($column);

// L454:
$after = $this->afterClause($column);
```

I'd be happy to submit a PR if this looks promising.

Thanks.

beporter:
Closes #655. (Only 5 years later.)

This is a MySQL-only syntax (as is the `AFTER` syntax to begin with) so no other adapters are relevant.

Since the Phinx repo isn't labeled with `hacktoberfest`, could I kindly request adding the `hacktoberfest-accepted` to this PR upon approval please?

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -77,6 +77,8 @@ class MysqlAdapter extends PdoAdapter
 
     public const TYPE_YEAR = 'year';
 
+    public const FIRST = 'FIRST';
+
     /**
      * {@inheritDoc}
      *
@@ -481,11 +483,30 @@ class MysqlAdapter extends PdoAdapter
             $this->getColumnSqlDefinition($column)
         );
 
-        if ($column->getAfter()) {
-            $alter .= ' AFTER ' . $this->quoteColumnName($column->getAfter());
-        }
+        $alter .= $this->afterClause($column);
 
         return new AlterInstructions([$alter]);
+    }
+
+    /**
+     * Exposes the MySQL syntax to arrange a column `FIRST`.
+     *
+     * @param Column $column The column being altered.
+     *
+     * @return string The appropriate SQL fragment.
+     */
+    protected function afterClause(Column $column)
+    {
+        $after = $column->getAfter();
+        if (empty($after)) {
+            return '';
+        }
+
+        if ($after === self::FIRST) {
+            return ' FIRST';
+        }
+
+        return ' AFTER ' . $this->quoteColumnName($after);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -481,13 +483,30 @@ protected function getAddColumnInstructions(Table $table, Column $column)
...
+     * @param Column $column The column being altered.
+     *
+     * @return string The appropriate SQL fragment.
+     */
+    protected function afterClause(Column $column)
+    {
+        $after = $column->getAfter();
+        if (empty($after)) {
+            return '';
+        } elseif ($after === self::FIRST) {
```
dereuromark, line 503: no else needed on early returns.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -661,6 +661,16 @@ public function testAddColumnWithDefaultLiteral()
...
     }
 
+    public function testAddColumnFirst()
+    {
+        $table = new \Phinx\Db\Table('table1', [], $this->adapter);
+        $table->save();
+        $table->addColumn('new_id', 'integer', ['after' => MysqlAdapter::FIRST])
+              ->save();
+        $rows = $this->adapter->fetchAll('SHOW COLUMNS FROM table1');
+        $this->assertTrue($rows[0]['Field'] === 'new_id');
```
dereuromark, line 671: assertSame() would probably be better here.

src/Phinx/Db/Adapter/MysqlAdapter.php
beporter: I have no idea what you're talking about. 😝

dereuromark (approved):

beporter:
Thanks for the (speedy!) feedback, Mark. Always a pleasure!

dereuromark:
Let's make this 5 years, and 13 minutes, shall we? ;)

Issue #1693, opened 2020-01-17
PostgreSQL adapter can't use smallserial type


badmansan:
```php
$smallSerial = (new \Phinx\Db\Table\Column())->setName('id')
	->setType(\Phinx\Db\Adapter\AdapterInterface::PHINX_TYPE_SMALL_INTEGER)
	->setIdentity(TRUE);

$this->table('test', ['id' => FALSE, 'primary_key' => ['id']])
	->addColumn($smallSerial)
	->create();
```
The result table contains column ```id``` with type ```integer``` instead ```smallint```.
This is because of [this](https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/PostgresAdapter.php#L1132) code:
```php
$buffer[] = $column->getType() == 'biginteger' ? 'BIGSERIAL' : 'SERIAL';
```

Can any one fix this?

dereuromark:
Why are you not using normal integers?
Are u able to make a pr here?

badmansan:
> Why are you not using normal integers?
My reference table is very small, but related table is very large.

> Are u able to make a pr here?
Sure, [here](https://imgur.com/a/L4T0Lrl)

dereuromark:
My tip for you: Don't nano-optimize.
It only creates problems without any benefit for you.
In this case you also don't gain anything.

badmansan:
May be you right)

PR #1915, opened 2020-10-23
add support for smallserial alias for postgres

MasterOdin:
Debated on whether to make this against master or 0.next, and decided for 0.next as this is a breaking chang, albeit a small one.

Closes (properly) #1693

Using the column type `smallinteger` with `identity = true` will produce `SMALLSERIAL` data type. This mirrors the usage of `biginteger` producing `BIGSERIAL`.

### DIFF
--- a/docs/en/migrations.rst
+++ b/docs/en/migrations.rst
@@ -838,7 +838,7 @@ Option    Description
 values    Can be a comma separated list or an array of values
 ========= ===========
 
-For ``integer`` and ``biginteger`` columns:
+For ``smallinteger``, ``integer`` and ``biginteger`` columns:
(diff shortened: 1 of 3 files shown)


MasterOdin:
Docs need to be added for this, but not sure where best to place them. Should I make a new section to describe this behavior, or just append it to column options as a note?

MasterOdin:
@dereuromark any word on PR?

dereuromark (approved):

dereuromark:
Is it worth its own section? Otherwise appending sounds reasonable for now.

MasterOdin:
@dereuromark done

Issue #1922, opened 2020-11-11
Is there any way to run seeds just once per class?


Marcin-TA:
Hi, 

I noticed that when I run seeds, Phinx doesn't check if the seeds have been already run. Phinx just runs it again and again ...n. Is there any reason why Phinx does do it? If I won't add e.g. new record to my table then I would like to add just one record, not multiple records. It works totally opposite to migration.

dereuromark:
Seeding is demo data import IMO.
As such, providing here also "migration history" seems usually overkill.

See https://github.com/cakephp/phinx/issues/981

dereuromark:
I think similar points have been raised before, and there is some merit to it.
Having the data versioned more easily along with the current time in structure would sure help.
Right now it is always a snapshot of demo data on the very last state of migrations
And you can easily forget those when adding migrations. At some point they might stop working then.

As for the duplication:
I think there is an easy way to solve this with the current way fixtures work:

Besides the existing hasTable() method, we could provide hasData() and based on that you could return early:
```php
    public function run()
    {
        if ($this->hasData()) {
            return; // Records > 0 found in DB, demo data available
        }
        
        ...
```
Or with specific data maybe as well:
```php
    public function run()
    {
        if ($this->hasData(['key' => 'my-key'])) {
            return; // Records > 0 found in DB for this key, this demo data available
        }
        
        ...
```

If we didn't want such specific methods, we could also add an Interface `OnlyOnceInterface` or alike 
and those seeds would internally check then if there is already demo data in those tables.

What do people think?

Marcin-TA:
I would keep the same logic which has migration. So if the seeds were added then store a record on phinxlog table, but with seed file name and DateTime. That's all.

Marcin-TA:
> I think similar points have been raised before, and there is some merit to it.
> Having the data versioned more easily along with the current time in structure would sure help.
> Right now it is always a snapshot of demo data on the very last state of migrations
> And you can easily forget those when adding migrations. At some point they might stop working then.
> 
> As for the duplication:
> I think there is an easy way to solve this with the current way fixtures work:
> 
> Besides the existing hasTable() method, we could provide hasData() and based on that you could return early:
> 
> ```
>     public function run()
>     {
>         if ($this->hasData()) {
>             return; // Records > 0 found in DB, demo data available
>         }
>         
>         ...
> ```
> 
> Or with specific data maybe as well:
> 
> ```
>     public function run()
>     {
[...]

PR #1926, opened 2020-11-13
Allow a way to insert demo data only for empty table.

dereuromark:
This would solve https://github.com/cakephp/phinx/issues/1922 , or somehow at least would allow to easily check for non empty table and skip the demo data input (on 2nd + run).

What do you think?
@MasterOdin 

Trying to go the interface way or some other more clean way might not be so trivial, but we could also always try to add that on top.
On problem also is that seeds can have dependencies to other seeds, so it becomes less and less easy without also storing migration data for seeds. And at that point we are back to just saying: Use migrations for both data and structure then.

### DIFF
--- a/src/Phinx/Seed/AbstractSeed.php
+++ b/src/Phinx/Seed/AbstractSeed.php
@@ -174,7 +174,7 @@ abstract class AbstractSeed implements SeedInterface
     public function hasData(string $tableName): bool
     {
         $table = $this->getAdapter()->quoteTableName($tableName);
-        $countQuery = $this->getAdapter()->query('SELECT COUNT(*) FROM ' . $table);
+        $countQuery = $this->getAdapter()->query('SELECT COUNT(*) as count FROM ' . $table);
         $res = $countQuery->fetchAll();
 
         return $res[0]['count'] > 0;
--- a/src/Phinx/Seed/AbstractSeed.php
+++ b/src/Phinx/Seed/AbstractSeed.php
@@ -162,11 +162,21 @@ abstract class AbstractSeed implements SeedInterface
      */
     public function insert($table, $data)
     {
-        // convert to table object
-        if (is_string($table)) {
-            $table = new Table($table, [], $this->getAdapter());
-        }
-        $table->insert($data)->save();
+        $tableInstance = new Table($table, [], $this->getAdapter());
+        $tableInstance->insert($data)->save();
+    }
+
+    /**
+     * @param string $tableName
+     *
+     * @return bool
+     */
+    public function hasData(string $tableName): bool
+    {
+        $countQuery = $this->getAdapter()->query('SELECT COUNT(*) FROM ' . $tableName);
+        $res = $countQuery->fetchAll();
+
+        return $res[0]['count'] > 0;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Phinx/Seed/AbstractSeed.php
```diff
@@ -162,11 +162,21 @@ public function fetchAll($sql)
...
+    }
+
+    /**
+     * @param string $tableName Table name.
+     *
+     * @return bool
+     */
+    public function hasData(string $tableName): bool
+    {
+        $countQuery = $this->getAdapter()->query('SELECT COUNT(*) FROM ' . $tableName);
```
MasterOdin, line 176: This will need to call down to the adapter to quote the table name (and potential schema prefix).

src/Phinx/Seed/AbstractSeed.php
```diff
@@ -162,11 +162,22 @@ public function fetchAll($sql)
...
+        $tableInstance = new Table($table, [], $this->getAdapter());
+        $tableInstance->insert($data)->save();
+    }
+
+    /**
+     * @param string $tableName Table name.
+     *
+     * @return bool
+     */
+    public function hasData(string $tableName): bool
```
MasterOdin, line 174: Should be this be added to the SeedInterface?

MasterOdin:
I agree that this method can be useful in the context of seeds (I can immediately rewrite one of my seeds to use it), and also agree that these sorts of utility methods are as far as should go in extending the Seed class. I do not think that we want to implement some complex set-up on figuring out whether seeds should be run or not when you type `seed:run` and that is an exercise best left to the end-user as there's a multitude of ways seeds can be run. For data that must exist in the DB at runtime in both testing and production, a migration is 100% the way to go there.

Personally, I use seeds to create test data for my DB. At the top of my seeds, I have a `if (hasData) { clearData(); }` type setup since I want my seeds to move my DB back to pristine test data from however I muddle it as I work on features or bugfixes. Having to work around some sort of migration like manager would just be a nuisance in that view.

dereuromark:
It doesnt work though
the count comes always back as 1, even if nothing is in the db.

I hate to have to mock so much
I would rather use fixtures and real DB testing :)

Yeah, we should ideally keep it simple for now.
I do not want to add it to the interface as this is just a convenience function and also, adding something to the interface would require a major at this time.

MasterOdin:
Thinking on this, we probably want to make the `hasData` a utility function off the AdapterInterface that the Seed can then call (similar to other methods like "hasTable"). In this way, the adapter can handle the implementation details of running this as there is no guarantee that `query()` will return a \PDOStatement like structure, and I am loathe to introduce that here.

> I do not want to add it to the interface as this is just a convenience function and also, adding something to the interface would require a major at this time.

Well, we're releasing 0.13 in a month right, so not like that's too far off.

dereuromark:
Fair enough.
If you have a better way of implementing this, go ahead.

dereuromark:
That said:
If you completely go the interface way, we could just as well allow my initial idea of implementing a OnlyOnEmptyTableInterface
or alike
And if the seed implements this interface, it knows to only run for empty one automatically.

Wouldnt that be even cleaner? Or would this be too inflexible then compared to the method?

MasterOdin:
Sure, but will probably be a week+ at least before I have time.

MasterOdin:
I would probably be fine in both having an abstract class that one could extend off of that implements this functionality automatically, but would still want the utility function as within my uses, I check for row existence to then clean out the tables affected, and then proceed to run my seed as I want to reset my DB back to pristine test condition.

PR #1927, opened 2020-11-24
Avoid error

dereuromark:
> 1) Migrations\Test\TestCase\MigrationsTest::testMigrateErrors
Failed asserting that exception of type "Error" matches expected exception "Exception". Message was: "Call to a member function getName() on bool" at
/.../vendor/robmorgan/phinx/src/Phinx/Db/Adapter/SQLiteAdapter.php:696

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -693,6 +693,9 @@ PCRE_PATTERN;
             // as the alternative is unwinding all possible table constraints which
             // gets messy quickly with CHECK constraints.
             $columns = $this->getColumns($tableName);
+            if (!$columns) {
+                return $state;
+            }
             $finalColumnName = end($columns)->getName();
             $sql = preg_replace(
                 sprintf(


MasterOdin:
Is this covering for a deeper bug in `getColumns` function, as far as I am aware, should not be possible for that function to ever return an empty array as you cannot have a table without any columns.

dereuromark:
Very possible, yes.

MasterOdin:
Looking at the Travis test results from cakephp/migrations#480, does not look like any failed with the test indicated above? I'll look into running the test locally, but just want to make sure it's not PHP version specific.

dereuromark:
No this happens when you have a not found Migration.
Quite the edge case.

MasterOdin (changes requested):
Looked at this and it looks like this is an error when you have a migration that does `addColumn` followed by `update()` for a table that does not exist. Phinx will then run the `getAddColumnInstructions` method as expected, but `getColumns` will return empty set (because table does not exist), which then causes issues elsewhere. The change in an earlier 0.12 release "broke" things in that instead of a `\PDOException` being thrown, now it's a fatal error. The `PDOException` was probably not super great anyway for debugging purposes (as it throws on copying data over into temp table from non-existing table), but better than the fatal error which just makes it look like phinx is messing up.

So, I think the change here to `SQLiteAdapter` is right and good and restores us back to throwing the PDOException and can add a test like this to catch this regression (feel free to change my horrible function name):

```
    public function testPdoExceptionUpdateNonExistingTable()
    {
        $this->expectException(\PDOException::class);
        $table = new \Phinx\Db\Table('non_existing_table', [], $this->adapter);
        $table->addColumn('column', 'string')->update();
    }
```

Issue #1824, opened 2020-06-25
Support pre and post migrations for zero-downtime deployments


Bilge:
It is common to run migrations at deploy time, and those migrations are often split into pre- and post-deploy migrations. Pre-deploy migrations affect the current version of the app for a few moments before the new version is deployed. Post-deploy migrations are only applied after the new version has been deployed. This is to ensure no errors occur during a live, zero-downtime deployment.

For example, new columns are usually added during pre-deploy, because if those columns are not present the moment the new version goes live, errors will occur because the app will try to access non-existent columns. Deleted columns should often be removed during post-deploy, because if they are removed whilst the current app version is running, errors will occur because the current version may still be trying to access them.

To support these two types of migrations, there needs to be a mechanism to mark a given migration is either *pre* or *post* and the migration command must accept an option to toggle which migration subgroup is being targeted. This changes migration command behaviour from one-pass two two-pass style, where the command must be invoked twice - once for each migration subgroup - to perform the complete migration.

### Key discussion points 

1. Should we implement this feature?
2. How should migrations be marked as _pre_ and _post_?
3. Should a new migration command be added or should the existing command be repurposed to support this behaviour with additional arguments or options?

rquadling:
An opinion ... my team looked at how it would be best to handle this sort of thing and the conclusion we came to was to put this process into the workflow and devops pipelines, rather than in the production pipeline where it could have negative impact or attempting to coerce pipeline aspects into Phinx.

To that end we came up with a silly acronym - ProMiReD (you can google that and see the top non-ad link! Rare to be number 1 on Google for anything useful!).

It may not be a perfect generic solution for all, but it is one that works for us.

We did explore the idea of tagging before and after migrations, but invariably, by the time you've chained in code changes and repeatables, the idea that Phinx alone can solve this problem is demonstrated as non-viable. At least with our team.

But having implemented ProMiReD, we have commit pipelines that reject PRs that affect more than one of migrations, repeatables, or code.

The advantage is that each task deals with 1 aspect of the reason for the change. Smaller tasks. Easier to review and test. Easier to deploy. Speed up all round.

So. Like I said, an opinion.

Bilge:
[...]

Issue #1619, opened 2019-10-02
Conditional migration


matheusab:
Hello, is it possible to conditionally execute a migration? 

I'm working on a software that communicates with a few databases of different types. Depending on its type, (e.g. X or Y) a migration should or should not be executed
Example:

DB type| Migration 1 | Migration 2 | Migration 3
------------ | ------------ | ------------ | ------------ | 
X | &#9745; |  | &#9745; 
Y | &#9745; | &#9745;  

I'm currently doing it this way:
```php
<?php

class Migration2 extends AbstractConditionalMigration
{
    
    public function change()
    {
        if ($this->shouldExecute()) {
            // migration script
        }
    }

    public function isTypeX(): bool
    {
        return false;
    }

    public function isTypeY(): bool
    {
        return true;
    }
}

```

where "shouldExecute" is a method implemented by `AbstractConditionalMigration` that verifies the type of the target environment and, based on the returned value of  `isTypeX()` and `isTypeY()`, whether the migration should be executed.

The AbstractConditionalMigration class:
```php
use Phinx\Migration\AbstractMigration;

abstract class AbstractConditionalMigration extends AbstractMigration
{
    public abstract function isTypeX(): bool;

    public abstract function isTypeY(): bool;

    protected function shouldExecute(): bool
    {
        $env = $this->getEnvironment();

        $shouldExecute = false;

        if ($this->isTypeX() && $this->isEnvTypeX($env)) {
            $shouldExecute = true;
        }

        if ($this->isTypeY() && $this->isEnvTypeY($env)) {
            $shouldExecute = true;
        }
        return $shouldExecute;
    }

   // ...
}
```

The thing is: for every migration class, I have to copy and paste that condition.
Also, it will generate a row for that migration in the database, even though it was not supposed to be executed.

matheusab:
Just noticed that it may be related to https://github.com/cakephp/phinx/issues/1044.

dereuromark:
In general I like the idea, as sometimes you are otherwise used to use inline if/else switches etc.

rquadling:
You're almost there.

In your phinx.php config file (or yaml or whatever), change
`'migration_base_class' => AbstractMigration::class,`
to
`'migration_base_class' => AbstractConditionalMigration::class,`

Then copy the standard template from src/Phinx/Migration/Migration.template.php.dist to somewhere in your codebase.

Change the phinx.php config to include:
```
        'templates' => [
[...]

Issue #1044, opened 2017-02-25
RFC: Ability to tag migrations and execute migrations by tag.


rquadling:
Scenario:

Multi-developer team, developing branches of code and migrations for a product.

Each branch will have multiple migrations created over the time that the branch is in development. Some of these migrations will need to be run before the code is deployed, other afterwards. Not all migrations are known prior to the development of the branch, so, as such, they are interleaved.

e.g.

AddNewTable (run before code deploy)
IntroduceNewColumnForRenaming (run before code deploy)
AddTriggersToPopulateNewColumn (run before code deploy)
DropTriggersAndOldColumn (run after code deploy)
AddSecondNewTable (run before code deploy)

Whilst it is possible to manually rename the migrations versions, this is still a problem for automatic deployment tools in that the migrations have to be manually processed within the deployment process.

In addition to migrating, rolling back migrations in the right order in terms of code deployment is quite complicated.

The proposal is to incorporate migration tagging, allowing you to tag migrations when they are created and filter migrations when they are executed.

**Migration creation**
Tagging would start with the team agreeing upon a default set of tag values. These would be added to the `phinx.yml` (or equivalent) and saved in the project repo for all developers to share. There will be 2 types of tag sets: SingleTag (no more than one tag may be selected from this set), MultiTag (any number of tags may be selected).

When you create a migration, you can supply an optional `--tag` parameter, and supply any number of tags.

Each tag will be validated against the tag definitions such that you cannot choose more than 1 tag from a SingleTag set, and that all tags are valid.

Any invalid tags will abort the creation of the migration.

Once validated, the tags would be presented to the migration template creation logic, similarly to the migration name and version.

The expected route for the template would be to add the tags as a comma separated value stored in a new migration class level constant `TAGS`.

For those using their own migration templates and/or migration template creation classes, then it will be their responsibility to upgrade their templates and classes to process the supplied tags.

**Migration execution**
Currently, any unexecuted migration is a target for migrating. An additional `--tag` parameter will be used to filter migrations based upon those that have _**all**_ the tags supplied. More tags means a more specific filter. At the time of writing this, I am not sure if an option should exist to allow the tags to be treated as `OR` rather than `AND`. My personal opinion is that the this feature only makes sense by using an `AND` approach.

So, as an example, say you have several new tables and permissions to migrate, adding these commands to your deployment pipeline would allow you to define the migration order in a consistent way:

```
phinx breakpoint
phinx migrate --tag BEFORE CREATE
[...]

Issue #529, opened 2015-03-26
Ability to separate pre and post-deploy migrations


acoulton:
We generally aim for zero-downtime migrations/deployments, which usually means splitting migrations to ensure that the database is always in a valid state for either old or new production code.

For example, new columns need to be added before the code is deployed, but code needs to be deployed before old columns can be removed.

As standard, we run migrations before deployment (and take care to write safe migrations for that context). When we need to run a destructive migration we usually change the code and deploy without adding a migration, leaving a `@todo` tag somewhere to come back later and do any cleanup on the DB.

The downside is that this leaves work for later, makes it harder to see the relationship between application and DB changes in source control, and makes pull requests etc harder to review.

It would be great instead if there was a way to write the pre and post deploy migrations at the same time, but tag them so that phinx could run them in two batches. We would then have a deploy flow like:
- `bin/phinx migrate --phase=pre-deploy`
- deploy code and do whatever checks that the deploy was happy, potentially waiting to see it in production a while
- `bin/phinx migrate --phase=post-deploy`

This could be implemented something like:

``` php
class AbstractMigration
{
  public function shouldApplyInPhase($phase)
  {
    // by default migration will run on the next migrate command, as now.
    return TRUE;
  }
}

class RemoveAColumn extends AbstractMigration
{
  public function shouldApplyInPhase($phase)
  {
    return $phase === 'post-deploy';
  }
}

class Phinx\Migration\Manager {
  public function migrate($environment, $version)
  {
      //...
      foreach($migrations as $migration) {
            if ($migration->getVersion() > $version) {
                break;
            }

           if (!$migration->shouldApplyInPhase($env->getMigrationPhase()) {
                continue;
           }

           if (!in_array($migration->getVersion(), $versions)) {
                $this->executeMigration($environment, $migration, MigrationInterface::UP);
            }
      }
  }
}
```

Alternatively, it could be done by including a prefix in the filename eg `20150326100000_post-deploy_remove_a_column.php` and filtering them at load-time.

This would only be relevant for migrate, for rollback it would be assumed the site had been taken offline for recovery.

pre and post deploy seem like the two most logical phases, but it would be easy to allow end users to use whatever phases suited them. For example you might define an extra 'maintenance-window' migration phase for more intensive cleanup tasks that should happen at some point but weren't required immediately post-deploy.

What do you think?

dereuromark:
Any feedback here from the team?

daidais:
[...]

PR #1939, opened 2020-12-08
Add new shouldExecute method to prevent a migration or seeder from being executed

AdrienPoupa:
The company I work for uses Phinx and we are pleased with it. However what's currently stopping us from using it in production is not being able to run migrations in two steps, before and after deployment.

This is an issue that's been raised several times already, in #529, #1044, #1619 and #1824 

I started using my own implementation of AbstractMigration and creating a custom template, but I realized that there was a key feature that I was missing if I did not want to wrap all my code in big `if` conditions in all my migrations: a way to stop the migrations from being executed at the Manager level, before it inserts data in phinxlog. This was rightly suggested here: https://github.com/cakephp/phinx/issues/1619#issuecomment-578732670

Even with `if` conditions as mentioned in #1619, the migration would still be recorded in the database and not tried again. What I needed was a way to make sure it would only be executed when the `shouldExecute` method returns true.

As I implemented it, a custom `AbstractMigration` implementation can override `shouldExecute` and perform any logic it wants to determine if the migration should run.

If not, it displays "skipped" in the console and stops there.

I was toying with the idea of changing how `preFlightCheck` works, and making it return a boolean and execute the migrations or not based on that, but it would break its existing behaviour, and I was not sure what to do with `postFlightCheck`, resulting in an inconsistency between the two methods.

While I was at it, I also refactored the console status message display to use a common method to display all the similar migration/seed status messages.

I believe this method would provide a simple, implementation agnostic solution for pre/post deployments and possibly other uses such as determining if a migration should run based on migration attributes, time, tags, environment, etc.

Please let me know if you have any comments! Thank you :)

### DIFF
--- a/docs/en/seeding.rst
+++ b/docs/en/seeding.rst
@@ -75,6 +75,14 @@ The Init Method
 The ``init()`` method is run by Phinx before the run method if it exists. This
 can be used to initialize properties of the Seed class before using run.
 
+The Should Execute Method
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``shouldExecute()`` method is run by Phinx before executing the seed.
+This can be used to prevent the seed from being executed at this time. It always
+returns true by default. You can override it in your custom ``AbstractSeed``
+implementation.
+
 Foreign Key Dependencies
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/en/migrations.rst
+++ b/docs/en/migrations.rst
@@ -158,6 +158,14 @@ The ``init()`` method is run by Phinx before the migration methods if it exists.
 This can be used for setting common class properties that are then used within
 the migration methods.
 
+The Should Execute Method
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``shouldExecute()`` method is run by Phinx before executing the migration.
+This can be used to prevent the migration from being executed at this time. It always
+returns true by default. You can override it in your custom ``AbstractMigration``
+implementation.
+
 Executing Queries
 -----------------
 
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -336,4 +336,18 @@ abstract class AbstractMigration implements MigrationInterface
             }
         }
     }
+
+    /**
+     * Checks to see if the migration should be executed.
+     *
+     * Returns true by default.
+     *
+     * You can use this to prevent a migration from executing.
+     *
+     * @return bool
+     */
+    public function shouldExecute()
+    {
+        return true;
+    }
 }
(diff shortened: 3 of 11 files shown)


dereuromark:
Does this also make https://github.com/cakephp/phinx/pull/1926 hook not necessary?

AdrienPoupa:
If I'm not mistaken #1926 is about seeds. But yes I could add the same check in the seed function to keep it consistent

dereuromark:
Given that we chance interface contracts, we also should put this into 0.13 (.next) branch as target

docs/en/seeding.rst
```diff
@@ -75,6 +75,14 @@ The Init Method
 The ``init()`` method is run by Phinx before the run method if it exists. This
 can be used to initialize properties of the Seed class before using run.
 
+The Should Execute Method
+~~~~~~~~~~~~~~~
```
dereuromark, line 79: Not enough ~ right?

AdrienPoupa: It's the same amount as the block above:

```
The Init Method
~~~~~~~~~~~~~~~
```

I can change it though

AdrienPoupa: I see what you mean, updated it

AdrienPoupa:
This MR is now targeting `0.next` and I applied the changes to Seeders.

stickler-ci complains about migrations but I guess it should ignore the namespaces there since most migrations do not have a namespace?

AdrienPoupa:
Then, about #1926 or even the pre/post deployment requests. I was thinking this would enable us to maybe use traits that define a specific behaviour for `shouldExecute`. For example, I think your use case could done as follows:

```php
trait SeedOnce 
{
    /**
     * @param string $tableName Table name.
     *
     * @return bool
     */
    public function hasData(string $tableName): bool
    {
        $table = $this->getAdapter()->quoteTableName($tableName);
        $countQuery = $this->getAdapter()->query('SELECT COUNT(*) as count FROM ' . $table);
        $res = $countQuery->fetchAll();

        return $res[0]['count'] > 0;
    }

    /**
     * Checks to see if the seed should be executed.
     *
     * Returns true by default.
     *
     * You can use this to prevent a seed from executing.
     *
     * @return bool
     */
    public function shouldExecute()
    {
        return !$this->hasData($this->seedOnceTable);
    }
}
```

```php
class CustomSeeder extends AbstractSeeder
{
    use SeedOnce;

    $seedOnceTable = 'posts';
}
```

I know traits often do not have a good reputation but in this case I think they could be handy :)

dereuromark:
It is even fine to just have the trait add the hasData() convenience method
Having to code

    public function shouldExecute()
    {
        return !$this->hasData($this->seedOnceTable);
    }

is still not too much work :)

AdrienPoupa:
You're right. However, using it as a trait (that could be bundled here) means that we do not have to add the `hasData` function to the `AbstractSeeder` and only have it when necessary.

AdrienPoupa:
Hello, I wonder if there are any modifications that should be made before it can be merged to the next branch?

AdrienPoupa:
Hello, I don't want to overstep but do we have an idea as to when it could be merged?

MasterOdin:
I'm fine with this, as it provides a nice generic abstraction over what #1926 was doing that works for both migrations and seeds.

AdrienPoupa:
Thanks a lot!

PR #1946, opened 2020-12-30
Add limit option for primary key migrations

olivertappin:


MasterOdin (changes requested):
The limit option for the default primary key should only be added for the `MysqlAdapter`. None of the other adapters have a concept of setting a limit for the integer type, making it a no-opt in those cases.

While MySQL also is finally moving away from the concept of "integer limits" (because it didn't actually do anything) in the 8.x release, I do think this change is good for making it easier to work with older installations that will probably be around for a long time yet.

PR #1951, opened 2021-01-21
Provide $input, $output, and $context variable to bootstrap script

martin-rueegg:
Currently, it is not possible to use the `$output` or `$input` of the command when running the bootstrap script. Which requires you to just `echo` any progress/diagnostic output or create your own output handler but then you would also have to parse the original verbosity parameters and so forth.

Passing the `$input` and `$output` parameters from the [`bootstrap()`-method](https://github.com/cakephp/phinx/blob/9457e3879b0abc02a4022941a6a535141889432e/src/Phinx/Console/Command/AbstractCommand.php#L105) will make them available in the included script as `$input` and `$output` respectively.

That would be enough for just output handling. But it would also be great to be able to access the command object itself, so there is also access to the config (via `getConfig()` and all the other information one might need during bootstrapping. Hence the `$context` parameter.

Making the parameters optional is not necessarily desired but required for backward-compatibility to be maintained. Although the [`loadPhpFile()`-method](https://github.com/cakephp/phinx/blob/9457e3879b0abc02a4022941a6a535141889432e/src/Phinx/Util/Util.php#L243) is only called by the above mentioned `bootstrap()`-method within the project, it has the public access modifier set (for obvious reasons) and hence it is part of the current API. This is also the reason why the fourth parameter is called `$context`, rather than e.g. `$command` as it could be something entirely different if used in a different context (e.g. when `loadPhpFile()` is called from within a migration, it could contain the migration object instead. Or even an array of options!

### DIFF
--- a/src/Phinx/Util/Util.php
+++ b/src/Phinx/Util/Util.php
@@ -10,6 +10,8 @@ namespace Phinx\Util;
 use DateTime;
 use DateTimeZone;
 use Exception;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
 
 class Util
 {
@@ -235,12 +237,15 @@ class Util
      * Takes the path to a php file and attempts to include it if readable
      *
      * @param string $filename Filename
+     * @param \Symfony\Component\Console\Input\InputInterface|null $input Input
+     * @param \Symfony\Component\Console\Output\OutputInterface|null $output Output
+     * @param \Phinx\Console\Command\AbstractCommand|mixed|null $context Context
      *
      * @throws \Exception
      *
      * @return string
      */
-    public static function loadPhpFile($filename)
+    public static function loadPhpFile($filename, ?InputInterface $input = null, ?OutputInterface $output = null, $context = null)
     {
         $filePath = realpath($filename);
         if (!file_exists($filePath)) {
@@ -257,6 +262,9 @@ class Util
         if (!$isReadable) {
             throw new Exception(sprintf("Cannot open file %s \n", $filename));
         }
+
+        // prevent this to be propagated to the included file
+        unset($isReadable);
 
         include_once $filePath;
 
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -418,3 +418,15 @@ setting External Variables to modify the config will not work because the config
 
     paths:
         bootstrap: 'phinx-bootstrap.php'
+
+Within the bootstrap script, the following variables will be available:
+
+.. code-block:: php
+
+    /**
+     * @var string $filename The file name as provided by the configuration
+     * @var string $filePath The absolute, real path to the file
+     * @var \Symfony\Component\Console\Input\InputInterface $input The executing command's input object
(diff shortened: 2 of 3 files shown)


martin-rueegg:
Not quite sure how to write a test here, as none of the bootstrap functionality is currently tested.

One option would be to add a test bootstrap script that would create a global array containing the provided arguments that can then be evaluated by the test?

src/Phinx/Util/Util.php
```diff
@@ -235,12 +237,15 @@ public static function glob($path)
...
+     * @param \Symfony\Component\Console\Input\InputInterface|null $input Input
+     * @param \Symfony\Component\Console\Output\OutputInterface|null $output Output
+     * @param \Phinx\Console\Command\AbstractCommand|mixed|null $context Context
      *
      * @throws \Exception
      *
      * @return string
      */
-    public static function loadPhpFile($filename)
+    public static function loadPhpFile($filename, InputInterface $input = null, OutputInterface $output = null, $context = null)
```
dereuromark, line 248: cs sniff not complaining?
Isnt this `?InputInterface` etc?

dereuromark (approved):

martin-rueegg:
Thanks for approving, awesome! 

I noticed I just missed the new release cycle - any idea when you might be doing the next one?

Since I'm using this feature during deployment already, it would obviously be very convenient to not have to use a vcs repository in my composer to get the snapshot ...

Thanks for your awesome work!

dereuromark:
We can have an appending patch pretty soon the next days if necessary.
They are pretty cheap to do after all.

martin-rueegg:
That would be highly appreciated. Thanks!

MasterOdin (approved):

martin-rueegg:
Hi @dereuromark 

Just gently asking, if you have already a date for the next release as we are now moving to production and would really be grateful if the patch would be released.

Thank you!

martin-rueegg:
For anyone needing this feature before the next release cycle, I've found a work-around:

```
use Phinx\Console\Command\Migrate;
use Phinx\Console\PhinxApplication;
use Symfony\Component\Console\Application;
use Symfony\Component\Console\Output\ConsoleOutput;
use Symfony\Component\Console\Output\OutputInterface;

if ( ! isset( $context ) && $GLOBALS['app'] instanceof PhinxApplication )
{
    $property = new \ReflectionProperty( Application::class, "runningCommand" );
    $property->setAccessible( true );
    $context = $property->getValue( $GLOBALS['app'] );

    unset( $property );
}

if ( ! isset( $output ) && $context instanceof Migrate )
{
    $output = $context->getManager()->getOutput();
}

if ( ! isset( $input ) && $context instanceof Migrate )
{
    $input = $context->getManager()->getInput();
}

```

dereuromark:
Isnt it already released as 0.12.6 ? :)

martin-rueegg:
You are right! Last time I checked it was not yet released - I believe. But apparently you did a release right after my [comment](https://github.com/cakephp/phinx/pull/1951#issuecomment-799873267)?

Anyways, fantastic it is released now! Thanks a lot.

PR #1952, opened 2021-01-25
tinyinteger fixed

munavvard2:
#1755 the issue in release 2.x not yet fixed

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1042,7 +1042,7 @@ class MysqlAdapter extends PdoAdapter
                 return ['name' => 'bigint', 'limit' => $limit ?: 20];
             case static::PHINX_TYPE_BOOLEAN:
             case static::INT_TINYINT_BOOLEAN:
-            return ['name' => 'tinyint', 'limit' => 1];
+                return ['name' => 'tinyint', 'limit' => 1];
             case static::PHINX_TYPE_UUID:
                 return ['name' => 'char', 'limit' => 36];
             case static::PHINX_TYPE_YEAR:
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -72,6 +72,7 @@ class MysqlAdapter extends PdoAdapter
     public const INT_MEDIUM = 16777215;
     public const INT_REGULAR = 4294967295;
     public const INT_BIG = 18446744073709551615;
+    public const INT_TINYINT_BOOLEAN = 1;
 
     public const BIT = 64;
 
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1040,6 +1041,7 @@ public function getSqlType($type, $limit = null)
             case static::PHINX_TYPE_BIG_INTEGER:
                 return ['name' => 'bigint', 'limit' => $limit ?: 20];
             case static::PHINX_TYPE_BOOLEAN:
+            case static::INT_TINYINT_BOOLEAN:
```
dereuromark, line 1044: Here we only use PHINX_ types usually

dereuromark: It means the above logic should be changed probably

        if ($type === 'tinyint' && $limit === 1) {
            $type = static::PHINX_TYPE_BOOLEAN;
            $limit = null;
        }

munavvard2: Okey, but i think it should be its own phinx type, shouldn't it?
It won't fall in any case, i will have to put it in default case, as it's generating tinyinteger, not tinyint.

dereuromark:
@MasterOdin Did we forget something here in the past?

src/Phinx/Db/Adapter/MysqlAdapter.php
dereuromark: I don't quite follow.
Tinyint(1) === boolean
Those are the same thing.

munavvard2: Absolutely same thing, but when migrations/seeds are generated it gets tinyinteger, not tinyint.

dereuromark: And tinyinteger is being converted to "tinyint" which then becomes boolean.
Do you have a specific error/issue currently?
This code works. I also use tinyint(1) in mysql.

munavvard2: i created an initial seed for my database and it generated all booleans as tinyintegers
`            ->addColumn('is_active', 'tinyinteger', [
                'default' => '0',
                'limit' => null,
                'null' => false,
            ])`

like that,
then i migrated that and it gave me error 
`Phinx\Db\Adapter\UnsupportedColumnTypeException: Column type "tinyinteger" is not supported by MySQL. in /var/www/html/<proj>/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php:964
`

and by debugging i found that type is not "tinyint", its "tinyinteger".

so by referring :
https://github.com/cakephp/phinx/blob/9457e3879b0abc02a4022941a6a535141889432e/src/Phinx/Db/Adapter/MysqlAdapter.php#L78

i added INT_TINYINT_BOOLEAN

munavvard2: using 
`"cakephp/migrations": "2.4.2",`

dereuromark: Is this supposed to be a boolean?
Then why not using

    ->addColumn('is_active', 'boolean', [ 'default' => '0', 'null' => false])

?

munavvard2: That's generated by `migrations seed`.
Absolutely, it should use boolean,
But while creating seed. It's giving tinyinteger. Instead of "boolean"

dereuromark: Isn't seed creation part of https://github.com/cakephp/migrations and should be reported and fixed there?
I don't see an issue in this repository right now.

PR #1957, opened 2021-02-23
 Add limit option for primary key migrations

MasterOdin:
Finishes up #1946, removing the setting from the non-mysql adapters, and adds a test case for it, which only runs on MySQL 5.7 and older as MySQL 8 removed integer limits.


MasterOdin:
@dereuromark codecov failure would be that this test is only run on PHP 7.2 which is not reporting its statistics.

dereuromark:
What can we do?

MasterOdin:
I modified the test so that it run on MySQL 8 (as it's just a null-op then), so codecov is happy now and I'd say this is ready to merge.

Might be worth looking to see if you could upload the results of PHP 7.2 runs to codecov (it should happily merge the results)?

PR #1962, opened 2021-02-24
Add optional $params argument to query/execute adapter methods

### LINKED ISSUE #1954: Allow parameters to be bound to migration SQL statements
It would be preferable to use the convenience methods (that add to the migration timer) rather than having to manually pull out the connection (from `$this->getAdapter()->getConnection()`) and prepare PDO statements when migrating data. This would avoid having to cast, sanitise, and/or validate values used in schema alterations.

https://github.com/cakephp/phinx/blob/9457e3879b0abc02a4022941a6a535141889432e/src/Phinx/Migration/AbstractMigration.php#L196
https://github.com/cakephp/phinx/blob/9457e3879b0abc02a4022941a6a535141889432e/src/Phinx/Migration/AbstractMigration.php#L204

 [Phinx's minimum PHP version](https://github.com/cakephp/phinx/wiki#version-and-branch-overview) is listed as 5.6 so the signatures could be made variadic:
```php
public function execute($sql, ...$params)
```
so developers can choose either:
```php
$this->execute($sql, ...$params);
```
or:
```php
$this->execute($sql, $param1, $param2, $param3);
```

MasterOdin:
Closes #1954

This allows a new optional parameter to the `query` and `execute` adapter methods to allow you to pass in an array of parameters to be used in a prepared query on the underlying PDOAdapter. It's assume that custom adapters people may implement will also have some mechanism of utilizing a concept of prepared queries. To maintain backwards compatibility, if the optional parameter is empty, then we use the direct `exec()` and `query()` methods on the connection to maintain support for executing multiple queries in one go.

The aim here was to keep the return result for the two methods consistent with that's there, just adding the optional parameter. This is then also exposed upwards in the `MigrationInterface` and `SeedInterface` classes.


MasterOdin:
This is marked as a draft as I've still got some amount of work to do to get this to work, and add tests across all adapters, but wanted input from @dereuromark, @garas, @Roy-Orbison that this would work for everyone's needs / expectations.

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1161,6 +1161,37 @@ public function testQueryBuilder()
...
+
+        $this->adapter->insert($table->getTable(), [
+            'string_col' => null,
+        ]);
+
+        $this->adapter->insert($table->getTable(), [
+            'int_col' => 23,
+        ]);
+
+        $countQuery = $this->adapter->query('SELECT COUNT(*) FROM table1 WHERE int_col > ?', [5]);
```
Roy-Orbison, line 1184: `COUNT(*) AS c` and then `$this->assertEquals(3, $res[0]['c']);`?

Roy-Orbison:
This will cover most use cases.

All other problems/features I raised are voided by that fact that, when using the connection directly, one can call the `startCommandTimer()`, etc. methods themself. Exposing a wrapped connection or something to automate that timer would make the code unnecessarily complex.

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1161,6 +1161,37 @@ public function testQueryBuilder()
...
+
+        $this->adapter->insert($table->getTable(), [
+            'string_col' => null,
+        ]);
+
+        $this->adapter->insert($table->getTable(), [
+            'int_col' => 23,
+        ]);
+
+        $countQuery = $this->adapter->query('SELECT COUNT(*) as c FROM table1 WHERE int_col > ?', [5]);
```
Roy-Orbison, line 1184: &lt;pedantism>You've got a lowercase `as` but other keywords are upper.&lt;/pedantism>

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -215,11 +218,14 @@ public function getQueryBuilder()
      *
      * @param string $sql SQL
      *
-     * @return \PDOStatement
+     * @return \PDOStatement|false
```
dereuromark, line 221: Can we do |null instead please? The false part also due to PHP legacy is such an antipattern, and it wouldnt allow for future return types added as easily as just `?object`

Roy-Orbison: Isn't that a BC break because this emulates `PDO::query()`?

dereuromark: 0.next can do bc breaks no? If that means cleaning up the API, thats probably a good thing.
But if this is too low level, then never mind.

Roy-Orbison: I wouldn't bother. Cleaning up your own API :+1:, cleaning up PHP's :grimacing:.

MasterOdin: I agree with @Roy-Orbison here that I think it would be good to follow PHP's antipattern here just so that it's consistent with how one would expect `PDO::query()` to work (and similar comment to `PDO::execute()`.

MasterOdin:
@dereuromark alright, finally returned to this and got it working, and should be ready for review, with fixed test case, and have also added to the docs for this.

docs/en/migrations.rst
```diff
@@ -193,6 +195,11 @@ Queries can be executed with the ``execute()`` and ``query()`` methods. The
                 // query()
                 $stmt = $this->query('SELECT * FROM users'); // returns PDOStatement
                 $rows = $stmt->fetchAll(); // returns the result as an array
+
+                // using prepared queries
+                $count = $this->execute('DELETE FROM users WHERE id = ?', [5]);
+                $stmt = $this->query('SELECT * FROM users WHERE id > 5'); // returns PDOStatement
```
Roy-Orbison, line 201: Is this meant to be parameterized like the previous statement?

MasterOdin: Yes 🤦. Fixed in #1994.

PR #1969, opened 2021-03-12
Add geometry column to sql server adapter

tedce:
Differentiate between geometry and geography columns in sql server


MasterOdin:
This should be put into 0.13 as it is a breaking change in behavior where anyone using SQLServer and `PHINX_TYPE_GEOMETRY` will need to replace it with `PHINX_TYPE_GEOGRAPHY`.

dereuromark:
Can you check if I can merge this? As the rebase doesn't seem to retrigger any tests

MasterOdin:
Test suite passes fine (except for codecov), though there are no tests around geometry / geography in sqlserver to begin with.

PR #1974, opened 2021-04-17
Postgres allow change from int to bool and char to uuid

jsm222:
For boolean start by dropping defaults since its value is not a bool,
then convert all non-zero values to boolean true.
For uuid add USING:uuid


dereuromark:
Looks good

pabloelcolombiano:
Hi @dereuromark 
this fix will enable Passbolt to support PostgreSQL, any idea on the timeline to get this shipped on a tag? Just to know how to prioritize. Thanks in advance for the feedback!

Issue #1972, opened 2021-04-15
SQL Server adapter still uses text and ntext types, despite deprecation


atbigelow:
According to MS's documentation for SQL Server 2017, `ntext` and `text` are deprecated and pending removal. It has been official guidance to use `varchar(max)` and `nvarchar(max)` for years instead of these types.

The use of these old types also requires SQL to be written to cast or convert when comparing strings, further encouraging the need for this to be changed.

The adapter should be changed to generate `varchar(max)` and `nvarchar(max)` instead of `text` and `ntext`.

PR #1977, opened 2021-04-29
SQL Server adapter uses nvarchar(max) instead of ntext and always updates default constraint

atbigelow:
The first commit of this PR handles #1972, which is to remove the use of NTEXT and replace it with NVARCHAR(MAX). The existing NTEXT and TEXT parsing to the Phinx TEXT type still exists for compatibility reasons.

The second commit forces default constraints to always be dropped and readded, as SQL Server requires this to be removed if the column is ever changed.

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -609,9 +609,6 @@ SQL;
     protected function getChangeColumnInstructions($tableName, $columnName, Column $newColumn)
     {
         $columns = $this->getColumns($tableName);
-        $changeDefault =
-            $newColumn->getDefault() !== $columns[$columnName]->getDefault() ||
-            $newColumn->getType() !== $columns[$columnName]->getType();
 
         $instructions = new AlterInstructions();
 
@@ -621,9 +618,7 @@ SQL;
             );
         }
 
-        if ($changeDefault) {
-            $instructions->merge($this->getDropDefaultConstraint($tableName, $newColumn->getName()));
-        }
+        $instructions->merge($this->getDropDefaultConstraint($tableName, $newColumn->getName()));
 
         $instructions->addPostStep(sprintf(
             'ALTER TABLE %s ALTER COLUMN %s %s',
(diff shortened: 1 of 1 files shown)


src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1109,17 +1103,23 @@ public function getSqlType($type, $limit = null)
...
+    public function getPhinxType($sqlType, $char_length = null)
     {
         switch ($sqlType) {
             case 'nvarchar':
             case 'varchar':
-                return static::PHINX_TYPE_STRING;
+                if ($char_length == -1) {
+                    // max char length
+                    return static::PHINX_TYPE_TEXT;
+                } else {
```
dereuromark, line 1120: the else is not necessary with early returns.

atbigelow: I went ahead and removed that.

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -471,7 +471,8 @@ public function getColumns($tableName)
                    ->setIdentity($columnInfo['identity'] === '1')
                    ->setComment($this->getColumnComment($columnInfo['table_name'], $columnInfo['name']));
 
-            if (!empty($columnInfo['char_length'])) {
+            if ($type != self::PHINX_TYPE_TEXT && !empty($columnInfo['char_length'])) {
```
dereuromark, line 474: !==

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1109,16 +1103,22 @@ public function getSqlType($type, $limit = null)
...
      *
      * @param string $sqlType SQL Type definition
+     * @param int|null $char_length Column length value
      *
      * @throws \Phinx\Db\Adapter\UnsupportedColumnTypeException
      *
      * @return string Phinx type
      */
-    public function getPhinxType($sqlType)
+    public function getPhinxType($sqlType, $char_length = null)
```
dereuromark, line 1112: charLength

dereuromark:
Any update? Or shall we close this?

atbigelow:
Went ahead and updated these. Sorry on the delay.

dereuromark:
Ping @MasterOdin

dereuromark:
I wonder if we might need a test case here to back things up.

PR #1979, opened 2021-05-13
Expose hasPrimaryKey method in Table object

MasterOdin:
This adds a new method to the `Table` object to allow checking for the existence of a primary key. This mirrors the existing `hasForeignKey` function on the object.

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -189,6 +189,19 @@ class Table
     }
 
     /**
+     * Checks to see if a primary key exists.
+     *
+     * @param string|string[] $columns Column(s)
+     * @param string|null $constraint Constraint names
+     *
+     * @return bool
+     */
+    public function hasPrimaryKey($columns, $constraint = null)
+    {
+        return $this->getAdapter()->hasPrimaryKey($this->getName(), $columns, $constraint);
+    }
+
+    /**
      * Changes the comment of the database table.
      *
      * @param string|null $comment New comment string, or null to drop the comment


MasterOdin:
@dereuromark I guess the ideal way to add a test for this would be to do something like:

```
public function testHasPrimaryKey() {
    $mock = $this->createMock(AbstractAdapter::class);
    $mock->expects($this->once())->method('hasPrimaryKey')->with('test', 'col', null)->toReturn(true);
    $table = new Table('test', [], $mock);
    $this->assertTrue($table->hasPrimaryKey('col'));
}
```

The adapters themselves have tests for `hasPrimaryKey` so probably don't need to replicate that logic here. Thoughts?

src/Phinx/Db/Table.php
```diff
@@ -188,6 +188,19 @@ public function changePrimaryKey($columns)
...
 
+    /**
+     * Checks to see if a primary key exists.
+     *
+     * @param string|string[] $columns Column(s)
+     * @param string|null $constraint Constraint names
+     *
+     * @return bool
+     */
+    public function hasPrimaryKey($columns, $constraint = null)
```
dereuromark, line 199: You could add a type for the 2nd param it seems. Same for return value

MasterOdin: I would not want to add type hinting to just one function in this file for fear of overall consistency. Perhaps on a release of 0.13, could do a quick release of 0.14 which just adds type hints where possible across the entire codebase, and then going forward type hints are always used?

Could also land such work in 0.13, but I'd almost at this point just say I'll get back to and land #1962 and then release instead of having it drag out too much longer.

dereuromark: Consistency sure is a point.
But this is a new public API method that is added.
And I think the benefits of types in place weigh higher than the consistency argument (from user perspective anyway).
I agree that any 0.13 semantic major release would need to clean up all possible types.

PR #1988, opened 2021-06-02
allow setting column to false to ignore it in addTimestamps

### LINKED ISSUE #1985: Allow addTimestamps to skip one of the columns
In some scenarios, the developer may not want both timestamp columns. I instinctively tried to supply `false`, but it results in an error caused by an empty field name.

My suggestion is either to:
- throw an error when a non-string/null value is supplied to either field names
- threat `false` as "skip this column" (and, of course, throw an error if both are false)

I would greatly support the latter (and probably provide a PR if greenlighted) since I see that happening a lot around my current project. A simple example: a logging table never gets updated entries, so `updated_at` makes no sense. The current method to achieve that is calling `addColumn()` by hand.

MasterOdin:
Closes #1985 

This allows using the boolean `false` for the first two arguments to `addTimestamps` and `addTimestampsWithTimezone` to have it skip adding that column to the table.

### DIFF
--- a/tests/Phinx/Db/TableTest.php
+++ b/tests/Phinx/Db/TableTest.php
@@ -13,19 +13,24 @@ use PHPUnit\Framework\TestCase;
 
 class TableTest extends TestCase
 {
+    public function provideAdapters()
+    {
+        return [[new SqlServerAdapter([])], [new MysqlAdapter([])], [new PostgresAdapter([])], [new SQLiteAdapter([])]];
+    }
+
     public function provideTimestampColumnNames()
     {
         $result = [];
-        $adapters = [new SqlServerAdapter([]), new MysqlAdapter([]), new PostgresAdapter([]), new SQLiteAdapter([])];
+        $adapters = $this->provideAdapters();
         foreach ($adapters as $adapter) {
             $result = array_merge(
                 $result,
                 [
-                    [$adapter, null, null, 'created_at', 'updated_at', false],
-                    [$adapter, 'created_at', 'updated_at', 'created_at', 'updated_at', true],
-                    [$adapter, 'created', 'updated', 'created', 'updated', false],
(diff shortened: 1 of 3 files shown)


tests/Phinx/Db/TableTest.php
```diff
@@ -136,6 +141,72 @@ public function testAddTimestamps(AdapterInterface $adapter, $createdAtColumnNam
...
+
+        $columns = [];
+
+        foreach ($actions as $action) {
+            $columns[] = $action->getColumn();
+        }
+
+        $this->assertEquals(1, count($columns));
+
+        $this->assertEquals('created_at', $columns[0]->getName());
```
dereuromark, line 163: We should use assertSame() for scalar values where possible.

tests/Phinx/Db/TableTest.php
```diff
@@ -136,6 +141,72 @@ public function testAddTimestamps(AdapterInterface $adapter, $createdAtColumnNam
...
+        $this->assertEquals('timestamp', $columns[0]->getType());
+        $this->assertEquals('CURRENT_TIMESTAMP', $columns[0]->getDefault());
+        $this->assertEquals(false, $columns[0]->getTimezone());
```
dereuromark, line 166: assertFalse()

igorsantos07:
Damn, I don't remember the last time I created an issue in an OS repo and the idea got accepted AND done in less than a few months. I didn't even have time to get psychologically prepared to make my "first" PR into the project :(((

Congratz you all! hahaha

tests/Phinx/Db/TableTest.php
MasterOdin: changed.

tests/Phinx/Db/TableTest.php
MasterOdin: changed.

PR #1991, opened 2021-06-16
Access static fields statically

InvisibleSmiley:
Allow subclasses to override the fields

Example use case: Map ENUM to VARCHAR


dereuromark:
Looks good to me

MasterOdin (approved):
This looks to be the only place in codebase where we were using `self` instead of `static`.

PR #1995, opened 2021-06-17
Add type hints to codebase

MasterOdin:
This PR will add type hints to all possible places in the codebase. I debated on adding `declare(strict_types=1);` at the same time, but not sure if should be done at the same time, or separately. Putting this up in draft form to keep track of progress, and hopefully help ensure no one else spends time on doing this concurrently.

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -11,6 +11,7 @@ use Closure;
 use InvalidArgumentException;
 use Phinx\Db\Adapter\SQLiteAdapter;
 use Phinx\Util\Util;
+use Psr\Container\ContainerInterface;
 use RuntimeException;
 use Symfony\Component\Yaml\Yaml;
 use UnexpectedValueException;
@@ -366,7 +367,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getContainer(): ?\Psr\Container\ContainerInterface
+    public function getContainer(): ?ContainerInterface
     {
         if (!isset($this->values['container'])) {
             return null;
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -41,7 +41,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     protected $values = [];
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $configFilePath;
 
@@ -258,7 +258,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getConfigFilePath(): string
+    public function getConfigFilePath(): ?string
     {
         return $this->configFilePath;
     }
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -60,9 +60,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the Yaml File
      * @throws \RuntimeException
-     * @return \Phinx\Config\Config
+     * @return self
      */
-    public static function fromYaml(string $configFilePath): Config
+    public static function fromYaml(string $configFilePath): self
     {
         if (!class_exists('Symfony\\Component\\Yaml\\Yaml', true)) {
             // @codeCoverageIgnoreStart
@@ -88,9 +88,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the JSON File
      * @throws \RuntimeException
-     * @return \Phinx\Config\Config
+     * @return self
      */
-    public static function fromJson(string $configFilePath): Config
+    public static function fromJson(string $configFilePath): self
     {
         if (!function_exists('json_decode')) {
             // @codeCoverageIgnoreStart
@@ -114,9 +114,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the PHP File
      * @throws \RuntimeException
-     * @return \Phinx\Config\Config
+     * @return self
      */
-    public static function fromPhp(string $configFilePath): Config
+    public static function fromPhp(string $configFilePath): self
     {
         ob_start();
         /** @noinspection PhpIncludeInspection */
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -60,9 +60,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the Yaml File
      * @throws \RuntimeException
-     * @return self
+     * @return \Phinx\Config\ConfigInterface
      */
-    public static function fromYaml(string $configFilePath): self
+    public static function fromYaml(string $configFilePath): ConfigInterface
     {
         if (!class_exists('Symfony\\Component\\Yaml\\Yaml', true)) {
             // @codeCoverageIgnoreStart
@@ -88,9 +88,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the JSON File
      * @throws \RuntimeException
-     * @return self
+     * @return \Phinx\Config\ConfigInterface
      */
-    public static function fromJson(string $configFilePath): self
+    public static function fromJson(string $configFilePath): ConfigInterface
     {
         if (!function_exists('json_decode')) {
             // @codeCoverageIgnoreStart
@@ -114,9 +114,9 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @param string $configFilePath Path to the PHP File
      * @throws \RuntimeException
-     * @return self
+     * @return \Phinx\Config\ConfigInterface
      */
-    public static function fromPhp(string $configFilePath): self
+    public static function fromPhp(string $configFilePath): ConfigInterface
     {
         ob_start();
         /** @noinspection PhpIncludeInspection */
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -49,7 +49,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @param array $configArray Config array
      * @param string|null $configFilePath Config file path
      */
-    public function __construct(array $configArray, $configFilePath = null)
+    public function __construct(array $configArray, ?string $configFilePath = null)
     {
         $this->configFilePath = $configFilePath;
         $this->values = $this->replaceTokens($configArray);
@@ -62,7 +62,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @throws \RuntimeException
      * @return \Phinx\Config\Config
      */
-    public static function fromYaml($configFilePath)
+    public static function fromYaml(string $configFilePath): Config
     {
         if (!class_exists('Symfony\\Component\\Yaml\\Yaml', true)) {
             // @codeCoverageIgnoreStart
@@ -90,7 +90,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @throws \RuntimeException
      * @return \Phinx\Config\Config
      */
-    public static function fromJson($configFilePath)
+    public static function fromJson(string $configFilePath): Config
     {
         if (!function_exists('json_decode')) {
             // @codeCoverageIgnoreStart
@@ -116,7 +116,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @throws \RuntimeException
      * @return \Phinx\Config\Config
      */
-    public static function fromPhp($configFilePath)
+    public static function fromPhp(string $configFilePath): Config
     {
         ob_start();
         /** @noinspection PhpIncludeInspection */
@@ -138,7 +138,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getEnvironments()
+    public function getEnvironments(): ?array
     {
         if (isset($this->values) && isset($this->values['environments'])) {
             $environments = [];
@@ -157,7 +157,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getEnvironment($name)
+    public function getEnvironment(string $name): ?array
     {
         $environments = $this->getEnvironments();
 
@@ -187,7 +187,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function hasEnvironment($name)
+    public function hasEnvironment(string $name): bool
     {
         return $this->getEnvironment($name) !== null;
     }
@@ -195,7 +195,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getDefaultEnvironment()
+    public function getDefaultEnvironment(): string
     {
         // The $PHINX_ENVIRONMENT variable overrides all other default settings
         $env = getenv('PHINX_ENVIRONMENT');
@@ -242,7 +242,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getAlias($alias)
+    public function getAlias($alias): ?string
     {
         return !empty($this->values['aliases'][$alias]) ? $this->values['aliases'][$alias] : null;
     }
@@ -250,7 +250,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getAliases()
+    public function getAliases(): array
     {
         return !empty($this->values['aliases']) ? $this->values['aliases'] : [];
     }
@@ -258,7 +258,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getConfigFilePath()
+    public function getConfigFilePath(): string
     {
         return $this->configFilePath;
     }
@@ -267,7 +267,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @inheritDoc
      * @throws \UnexpectedValueException
      */
-    public function getMigrationPaths()
+    public function getMigrationPaths(): array
     {
         if (!isset($this->values['paths']['migrations'])) {
             throw new UnexpectedValueException('Migrations path missing from config file');
@@ -281,23 +281,10 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     }
 
     /**
-     * Gets the base class name for migrations.
-     *
-     * @param bool $dropNamespace Return the base migration class name without the namespace.
-     * @return string
-     */
-    public function getMigrationBaseClassName($dropNamespace = true)
-    {
-        $className = !isset($this->values['migration_base_class']) ? 'Phinx\Migration\AbstractMigration' : $this->values['migration_base_class'];
-
-        return $dropNamespace ? (substr(strrchr($className, '\\'), 1) ?: $className) : $className;
-    }
-
-    /**
      * @inheritDoc
      * @throws \UnexpectedValueException
      */
-    public function getSeedPaths()
+    public function getSeedPaths(): array
     {
         if (!isset($this->values['paths']['seeds'])) {
             throw new UnexpectedValueException('Seeds path missing from config file');
@@ -311,12 +298,25 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     }
 
     /**
+     * Gets the base class name for migrations.
+     *
+     * @param bool $dropNamespace Return the base migration class name without the namespace.
+     * @return string
+     */
+    public function getMigrationBaseClassName(bool $dropNamespace = true): string
+    {
+        $className = !isset($this->values['migration_base_class']) ? 'Phinx\Migration\AbstractMigration' : $this->values['migration_base_class'];
+
+        return $dropNamespace ? (substr(strrchr($className, '\\'), 1) ?: $className) : $className;
+    }
+
+    /**
      * Gets the base class name for seeders.
      *
      * @param bool $dropNamespace Return the base seeder class name without the namespace.
      * @return string
      */
-    public function getSeedBaseClassName($dropNamespace = true)
+    public function getSeedBaseClassName(bool $dropNamespace = true): string
     {
         $className = !isset($this->values['seed_base_class']) ? 'Phinx\Seed\AbstractSeed' : $this->values['seed_base_class'];
 
@@ -354,7 +354,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * {@inheritdoc}
      */
-    public function getDataDomain()
+    public function getDataDomain(): array
     {
         if (!isset($this->values['data_domain'])) {
             return [];
@@ -366,7 +366,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritDoc
      */
-    public function getContainer()
+    public function getContainer(): ?\Psr\Container\ContainerInterface
     {
         if (!isset($this->values['container'])) {
             return null;
@@ -380,7 +380,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @return string
      */
-    public function getVersionOrder()
+    public function getVersionOrder(): string
     {
         if (!isset($this->values['version_order'])) {
             return self::VERSION_ORDER_CREATION_TIME;
@@ -394,7 +394,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      *
      * @return bool
      */
-    public function isVersionOrderCreationTime()
+    public function isVersionOrderCreationTime(): bool
     {
         $versionOrder = $this->getVersionOrder();
 
@@ -421,7 +421,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @param array $arr Array to replace
      * @return array
      */
-    protected function replaceTokens(array $arr)
+    protected function replaceTokens(array $arr): array
     {
         // Get environment variables
         // Depending on configuration of server / OS and variables_order directive,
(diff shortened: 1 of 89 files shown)


dereuromark:
I would hold off on the strict_types part, as this can create some more pitfalls where currently auto-casting works (to not break more than necessary).

src/Phinx/Db/Table/Index.php
```diff
@@ -59,12 +59,12 @@ class Index
...
      *
-     * @param string[] $columns Columns
+     * @param string|string[] $columns Columns
      * @return $this
      */
-    public function setColumns($columns)
+    public function setColumns($columns): Index
     {
-        $this->columns = $columns;
+        $this->columns = is_string($columns) ? [$columns] : $columns;
```
MasterOdin, line 67: This change makes this method equivalent to how `setColumns` works in the `ForeignKey` class.

src/Phinx/Db/Table/Table.php
```diff
@@ -73,10 +73,12 @@ public function getOptions()
...
      * @param array $options The options for the table creation
-     * @return void
+     * @return $this
      */
-    public function setOptions(array $options)
+    public function setOptions(array $options): Table
     {
         $this->options = $options;
+
+        return $this;
```
MasterOdin, line 82: The other classes under this namespace all return their class instance at the conclusion of `setOptions`, so adding this for consistency.

src/Phinx/Db/Action/ChangeColumn.php
```diff
@@ -55,7 +55,7 @@ public function __construct(Table $table, $columnName, Column $column)
      * @param mixed $options Additional options for the column
```
dereuromark, line 55: I wonder if we can check more of those "mixed" ones and fix up the docblock to be more specific.

src/Phinx/Wrapper/TextWrapper.php
```diff
@@ -153,7 +153,7 @@ public function getSeed($env = null, $target = null, $seed = null)
      * @param mixed $target Target version, or 0 (zero) fully revert (optional)
```
dereuromark, line 153: Like here, too

src/Phinx/Db/Table/Column.php
```diff
@@ -76,37 +76,37 @@ class Column
     /**
      * @var mixed|null
      */
-    protected $default;
+    protected $default = null;
```
dereuromark, line 79: Thats not necessary, no value is null.

MasterOdin: Alright, was not sure as some parameters had an explicit `null` set versus others which just had the typedoc. Can remove them throughout the codebase?

src/Phinx/Config/Config.php
```diff
@@ -60,9 +60,9 @@ public function __construct(array $configArray, $configFilePath = null)
      *
      * @param string $configFilePath Path to the Yaml File
      * @throws \RuntimeException
-     * @return \Phinx\Config\Config
+     * @return \Phinx\Config\ConfigInterface
```
MasterOdin, line 63: Weakened this to `\Phinx\Config\ConfigInterface` as I believe if someone extends off `Config` and runs this function, then it would return an instance of the extended class and not `Config` via the usage of the `static` keyword on initializing the class.

src/Phinx/Db/Action/AddColumn.php
```diff
@@ -37,10 +37,10 @@ public function __construct(Table $table, Column $column)
      * @param \Phinx\Db\Table\Table $table The table to add the column to
      * @param string $columnName The column name
      * @param mixed $type The column type
-     * @param mixed $options The column options
-     * @return \Phinx\Db\Action\AddColumn
+     * @param array $options The column options
+     * @return self
```
MasterOdin, line 41: phpcs did not like it having the return type annotation matching the type hint. However, I think both might be wrong here given the static binding, not sure how to resolve? replace `@return self` with `@return static` and remove the inline type hint until phinx exclusively uses PHP 8.0? This is the same with other files in the `Action` sub-folder.

Though could maybe just remove the usage of `static` here and use `new AddColumn` and not have this class be extendable given it doesn't use an interface.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -417,31 +417,31 @@ public function hasPrimaryKey($tableName, $columns, $constraint = null);
...
      * Converts the Phinx logical type to the adapter's SQL type.
      *
-     * @param string $type Type
-     * @param int|null $limit Limit
+     * @param string|\Phinx\Util\Literal $type Type
+     * @param float|null $limit Limit
      * @return array
      */
-    public function getSqlType($type, $limit = null);
+    public function getSqlType($type, ?float $limit = null): array;
```
MasterOdin, line 444: This was changed from `int` to `float` to accommodate the `MysqlAdapter::INT_BIG` constant which has a value of `18446744073709551615` where that is treated as a `float` within PHP. Typing this (or having the integer typedoc) is then technically incorrect, and would throw a type error if we were to turn on `declare(strict_types=1);` and then `MysqlAdapter::INT_BIG` was used. The best option perhaps would be to type as `float`, and then just document that while it's a `float`, only integer like numbers should be used and decimal values will not work. The problem though is that now integers passed into the function are implicitly converted to floats (e.g. `4` -> `4.0`) which is a bit odd.

Could also maybe insert a:
```
if ($limit <= PHP_INT_MAX) {
    $limit = (int)$limit;
}
```
in the implemented functions to retain the current behavior best we can while having proper typehints?

src/Phinx/Db/Action/ChangeColumn.php
MasterOdin: Fixed a couple of these. We can probably resolve a number of the `mixed` values throughout the codebase, but I'd rather not bloat this PR more than it already is and treat it more in the spirit of "mechanical replacement" and then do targeted replacements of the remaining mixed values.

src/Phinx/Db/Table/Index.php
MasterOdin: Though the handling of the `$columns` value is very inconsistent across the codebase. A number of the `AlterInstruction` functions within the database adapters accepts a `string $column` property while some accept a `Column $column`, and then elsewhere we have `string|string[] $columns` or just `string[] $columns`. A follow-up PR would perhaps be good to work on making this more consistent throughout.

src/Phinx/Db/Table/Column.php
dereuromark: The code sniffer (at least mine does^^) should remove the overhead `= null` usually. 
We might want to add e.g. https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Classes/PropertyDefaultValueSniff.php on top

src/Phinx/Db/Action/AddColumn.php
dereuromark: Return types should never be self/static until we are fully PHP8, just omit them, then the docs here are easy enough.
The sniffer usually should check that no other return type is inside the scope of the method, and then a return type is not needed and superfluous.

MasterOdin: Not sure I follow on what you want to see for this specific function or elsewhere. For any case where the return docstring is `@return self` or `@return static`, just omit the inline return type altogether?

dereuromark: The functional return type, yes. No return types for self or static references, as they can be purely CS checked.

src/Phinx/Config/Config.php
```diff
@@ -366,7 +366,7 @@ public function getDataDomain()
     /**
      * @inheritDoc
      */
-    public function getContainer()
+    public function getContainer(): ?\Psr\Container\ContainerInterface
```
othercorey, line 369: Add a `use` statement for this.

src/Phinx/Config/ConfigInterface.php
```diff
@@ -108,28 +108,28 @@ public function getTemplateClass();
      *
      * @return \Psr\Container\ContainerInterface|null
      */
-    public function getContainer();
+    public function getContainer(): ?\Psr\Container\ContainerInterface;
```
othercorey, line 111: Add a `use` statement for this

src/Phinx/Console/Command/AbstractCommand.php
```diff
@@ -44,7 +44,7 @@ abstract class AbstractCommand extends Command
     protected const DEFAULT_SEED_TEMPLATE = '/../../Seed/Seed.template.php.dist';
 
     /**
-     * @var \Phinx\Config\ConfigInterface
+     * @var \Phinx\Config\ConfigInterface|null
```
othercorey, line 47: Can you actually have a null config?

src/Phinx/Db/Action/AddColumn.php
othercorey: The doc return can can be `static` here. `self` isn't correct for `new static`.

src/Phinx/Db/Action/AddForeignKey.php
```diff
@@ -41,9 +41,9 @@ public function __construct(Table $table, ForeignKey $fk)
      * @param string|string[] $referencedColumns The columns in the referenced table
      * @param array $options Extra options for the foreign key
      * @param string|null $name The name of the foreign key
-     * @return \Phinx\Db\Action\AddForeignKey
+     * @return self
```
othercorey, line 44: Use `static` type here too

src/Phinx/Db/Action/AddForeignKey.php
```diff
@@ -41,9 +41,9 @@ public function __construct(Table $table, ForeignKey $fk)
...
      */
-    public static function build(Table $table, $columns, $referencedTable, $referencedColumns = ['id'], array $options = [], $name = null)
+    public static function build(Table $table, $columns, $referencedTable, $referencedColumns = ['id'], array $options = [], ?string $name = null): AddForeignKey
```
othercorey, line 46: Can omit the return type here I think.

src/Phinx/Db/Action/AddIndex.php
```diff
@@ -38,20 +38,16 @@ public function __construct(Table $table, Index $index)
      * @param \Phinx\Db\Table\Table $table The table to add the index to
      * @param mixed $columns The columns to index
      * @param array $options Additional options for the index creation
-     * @return \Phinx\Db\Action\AddIndex
+     * @return self
```
othercorey, line 41: `static` here too

src/Phinx/Db/Adapter/AdapterFactory.php
```diff
@@ -64,16 +64,16 @@ public static function instance()
      * Add or replace an adapter with a fully qualified class name.
```
othercorey, line 64: This description should be updated.

src/Phinx/Db/Adapter/AdapterFactory.php
```diff
@@ -140,9 +140,9 @@ public function registerWrapper($name, $class)
      *
      * @param string $name Name
      * @throws \RuntimeException
-     * @return string
+     * @return string|object
```
othercorey, line 143:
```suggestion
     * @return object|string
```

src/Phinx/Db/Adapter/AdapterFactory.php
```diff
@@ -118,16 +118,16 @@ public function getAdapter($name, array $options)
      * Add or replace a wrapper with a fully qualified class name.
      *
      * @param string $name Name
-     * @param string $class Class
+     * @param string|object $class Class
```
othercorey, line 121:
```suggestion
     * @param object|string $class Class
```

src/Phinx/Db/Adapter/AdapterFactory.php
```diff
@@ -64,16 +64,16 @@ public static function instance()
      * Add or replace an adapter with a fully qualified class name.
      *
      * @param string $name Name
-     * @param string $class Class
+     * @param string|object $class Class
```
othercorey, line 67:
```suggestion
     * @param object|string $class Class
```

src/Phinx/Db/Adapter/AdapterFactory.php
```diff
@@ -52,7 +52,7 @@ public static function instance()
     /**
      * Class map of adapters wrappers, indexed by name.
      *
-     * @var string[]
+     * @var (string|object)[]
```
othercorey, line 55:
```suggestion
     * @var object[]|string[]
```

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -72,76 +72,76 @@ interface AdapterInterface
...
 
     /**
      * Sets the console input.
      *
      * @param \Symfony\Component\Console\Input\InputInterface $input Input
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function setInput(InputInterface $input);
+    public function setInput(InputInterface $input): AdapterInterface;
```
othercorey, line 122: The function should not have a return type for $this hints.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -72,76 +72,76 @@ interface AdapterInterface
...
 
     /**
      * Sets the console output.
      *
      * @param \Symfony\Component\Console\Output\OutputInterface $output Output
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function setOutput(OutputInterface $output);
+    public function setOutput(OutputInterface $output): AdapterInterface;
```
othercorey, line 137: Remove return type hint for `$this`.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -160,97 +160,97 @@ public function getColumnForType($columnName, $type, array $options);
      * @param string $direction Direction
      * @param string $startTime Start Time
      * @param string $endTime End Time
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function migrated(MigrationInterface $migration, $direction, $startTime, $endTime);
+    public function migrated(MigrationInterface $migration, string $direction, string $startTime, string $endTime): AdapterInterface;
```
othercorey, line 165: Same here.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -160,97 +160,97 @@ public function getColumnForType($columnName, $type, array $options);
...
 
     /**
      * Toggle a migration breakpoint.
      *
      * @param \Phinx\Migration\MigrationInterface $migration Migration
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function toggleBreakpoint(MigrationInterface $migration);
+    public function toggleBreakpoint(MigrationInterface $migration): AdapterInterface;
```
othercorey, line 173: And here.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -160,97 +160,97 @@ public function getColumnForType($columnName, $type, array $options);
...
 
     /**
      * Set a migration breakpoint.
      *
      * @param \Phinx\Migration\MigrationInterface $migration The migration target for the breakpoint set
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function setBreakpoint(MigrationInterface $migration);
+    public function setBreakpoint(MigrationInterface $migration): AdapterInterface;
```
othercorey, line 188: Same here. Remove return type hint for $this

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -160,97 +160,97 @@ public function getColumnForType($columnName, $type, array $options);
...
 
     /**
      * Unset a migration breakpoint.
      *
      * @param \Phinx\Migration\MigrationInterface $migration The migration target for the breakpoint unset
-     * @return \Phinx\Db\Adapter\AdapterInterface
+     * @return $this
      */
-    public function unsetBreakpoint(MigrationInterface $migration);
+    public function unsetBreakpoint(MigrationInterface $migration): AdapterInterface;
```
othercorey, line 196: Same here. Remove return type hint for $this

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -268,14 +268,14 @@ public function execute($sql, array $params = []);
...
-    public function executeActions(Table $table, array $actions);
+    public function executeActions(Table $table, array $actions): void;
 
     /**
      * Returns a new Query object
      *
      * @return \Cake\Database\Query
      */
-    public function getQueryBuilder();
+    public function getQueryBuilder(): \Cake\Database\Query;
```
othercorey, line 278: Add a `use` statement for this.

src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -417,31 +417,31 @@ public function hasPrimaryKey($tableName, $columns, $constraint = null);
...
      */
-    public function isValidColumnType(Column $column);
+    public function isValidColumnType(Column $column): bool;
 
     /**
      * Converts the Phinx logical type to the adapter's SQL type.
      *
-     * @param string $type Type
-     * @param int|null $limit Limit
+     * @param string|\Phinx\Util\Literal $type Type
```
othercorey, line 440:
```suggestion
     * @param \Phinx\Util\Literal|string $type Type
```

src/Phinx/Db/Table/Column.php
```diff
@@ -64,54 +64,54 @@ class Column
...
      * @var bool
      */
     protected $identity = false;
 
     /**
-     * @var int
+     * @var int|null
      */
-    protected $seed;
+    protected $seed = null;
```
othercorey, line 89: There's no need to default to null for any of these.

src/Phinx/Db/Table/Index.php
```diff
@@ -39,42 +39,42 @@ class Index
     /**
      * @var string|null
      */
-    protected $name;
+    protected $name = null;
```
othercorey, line 42: No need to default to null

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -36,19 +36,19 @@ abstract class AbstractMigration implements MigrationInterface
     protected $version;
 
     /**
-     * @var \Phinx\Db\Adapter\AdapterInterface
+     * @var \Phinx\Db\Adapter\AdapterInterface|null
      */
-    protected $adapter;
+    protected $adapter = null;
```
othercorey, line 41: No need to default to null

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -185,95 +185,95 @@ public function setMigratingUp($isMigratingUp)
...
+    public function query(string $sql, array $params = [])
     {
         return $this->getAdapter()->query($sql, $params);
     }
 
     /**
      * @inheritDoc
      */
-    public function getQueryBuilder()
+    public function getQueryBuilder(): \Cake\Database\Query
```
othercorey, line 212: Add a `use` statement.

src/Phinx/Migration/Manager/Environment.php
```diff
@@ -29,14 +29,14 @@ class Environment
     protected $options;
 
     /**
-     * @var \Symfony\Component\Console\Input\InputInterface
+     * @var \Symfony\Component\Console\Input\InputInterface|null
      */
-    protected $input;
+    protected $input = null;
```
othercorey, line 34: No need to default to null

src/Phinx/Migration/MigrationInterface.php
```diff
@@ -42,91 +42,91 @@ interface MigrationInterface
...
 
     /**
      * Sets the input object to be used in migration object
      *
      * @param \Symfony\Component\Console\Input\InputInterface $input Input
-     * @return \Phinx\Migration\MigrationInterface
+     * @return $this
      */
-    public function setInput(InputInterface $input);
+    public function setInput(InputInterface $input): MigrationInterface;
```
othercorey, line 62: Remove return type for `$this` hints

src/Phinx/Seed/SeedInterface.php
```diff
@@ -33,59 +33,66 @@ interface SeedInterface
...
 
     /**
      * Sets the database adapter.
      *
      * @param \Phinx\Db\Adapter\AdapterInterface $adapter Database Adapter
-     * @return \Phinx\Seed\SeedInterface
+     * @return $this
      */
-    public function setAdapter(AdapterInterface $adapter);
+    public function setAdapter(AdapterInterface $adapter): SeedInterface;
```
othercorey, line 51: Remove return type for `$this` hints

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -1058,7 +1058,7 @@ public function testIntegerColumnLimit()
               ->save();
         $columns = $table->getColumns();
         $sqlType = $this->adapter->getSqlType($columns[1]->getType(), $columns[1]->getLimit());
-        $this->assertEquals($this->usingMysql8() ? 11 : $limit, $sqlType['limit']);
+        $this->assertEquals($this->usingMysql8() ? 11.0 : $limit, $sqlType['limit']);
```
othercorey, line 1061: This feels odd.

othercorey (changes requested):
Reviewed syntax. Left a couple open questions about valid types, but just from a logic point of view.

dereuromark:
Shall we continue/finish this for the next branch?

src/Phinx/Console/Command/AbstractCommand.php
MasterOdin: As the config is initialized in its own function, yes. It's why this was necessary for a period of time:
```php
        /** @var \Phinx\Config\ConfigInterface|null $config */
        $config = $this->getConfig();
        if (!$config) {
            $this->loadConfig($input, $output);
        }
```

to make phpstan happy as `$this->getConfig()` was annotated as always returning `ConfigInterface`, but on first call to `bootstrap()` (which holds the above reference), it would be null.

MasterOdin:
@othercorey requested changes have been made.

@dereuromark the failing phpstan test is due to a bug in phpstan: https://github.com/phpstan/phpstan/issues/5369. Could maybe pin the used version to an older one that doesn't have this bug?

src/Phinx/Console/Command/Create.php
```diff
@@ -201,7 +201,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         // Get the alternative template and static class options from the command line, but only allow one of them.
+        /** @var string|null */
```
dereuromark, line 204: the variable is missing here

MasterOdin: Added.

The variable isn't strictly necessary as the IDE / phpstan will just apply whatever is there then to the variable in the next line, though having the variable is clearer.

dereuromark:
@othercorey What do you think about the float thingie? Should we keep as is, or do we want to internally work with integers here (as one intuitively would)?

othercorey:
Looks like `limit` is the same as `length` in cakephp. We cast length to int so doesn't seem like it would hurt to do that here. I don't think there are any fractional length types.

Is it possible to cast the limit when reflecting?

MasterOdin:
For the MySQL adapter, the following values are problematic if declaring a `int` limit:

https://github.com/cakephp/phinx/blob/26adc419257b84647602ec4e39e66b850116430b/src/Phinx/Db/Adapter/MysqlAdapter.php#L61

https://github.com/cakephp/phinx/blob/26adc419257b84647602ec4e39e66b850116430b/src/Phinx/Db/Adapter/MysqlAdapter.php#L68

https://github.com/cakephp/phinx/blob/26adc419257b84647602ec4e39e66b850116430b/src/Phinx/Db/Adapter/MysqlAdapter.php#L73-L74

These values all exceed the value of `INT_MAX` on 32-bit systems and `INT_BIG` exceeds `INT_MAX` on 64-bit systems. While these values are set to match the max size of the respective column types, given that these values exist purely as a shorthand to get say `bigint(20)` when doing `->column('foo', 'integer', ['limit' => MysqlAdapter::INT_BIG]`, could just arbitrarily cap them to be values less than `INT_MAX`, and then can declare the `limit` clause as an integer again.

This would be BC breaking, but in a way that I'm not sure anyone would notice. The specifics here are that:
* The values of the above constants would be different
* Assuming the new value of `TEXT_LONG` is now `2147483647` instead of `4294967295`, then values between `2147483647` and `4294967295` would be considered `TEXT_LONG` instead of `TEXT_MEDIUM` (this would apply to the other types in similar way)

However, again, given that these are supposed to be "magic constants" as it were and not to directly depend on the underlying value, should be fine to change?

othercorey:
> These values all exceed the value of `INT_MAX` on 32-bit systems and `INT_BIG` exceeds `INT_MAX` on 64-bit systems. While these values are set to match the max size of the respective column types, given that these values exist purely as a shorthand to get say `bigint(20)` when doing `->column('foo', 'integer', ['limit' => MysqlAdapter::INT_BIG]`, could just arbitrarily cap them to be values less than `INT_MAX`, and then can declare the `limit` clause as an integer again.

Looks like this is implemented differently than `length` in cakephp (which is just reflecting the mysql metadata). This should probably be implemented differently, but that's a separate issue.

I leave it up to you. I was just reviewing the type conventions and noticed the type changed in an odd way there.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
MasterOdin: Changed back to `limit` being an integer.

src/Phinx/Db/Adapter/AdapterInterface.php
MasterOdin: This was reverted, such that this function expects an integer again.

MasterOdin:
@othercorey @dereuromark I followed through with what I talked about above, giving the `MysqlAdapter` constants values pinned at `PHP_INT_MAX` so that `setLimit` can be an integer again.

> Looks like this is implemented differently than length in cakephp (which is just reflecting the mysql metadata). This should probably be implemented differently, but that's a separate issue.

That's the expected usage of `limit` (or its alias `length`). The constants just exist as a way of specifying a different type for `integer`, `string`, and `blob` base types on mysql. When actually fetching columns from the DB, it's expected that their actual limit / length will be an integer well below `PHP_INT_MAX` as expected.

Issue #1996, opened 2021-06-27
Cannot define the display width for mediumints with MysqlAdapter


yohanleafheart:
Greetings all, 

There is a bug/missing implementation on treating `mediumint(n)` for the MysqlAdapter. If you go to `phinx/src/Phinx/Db/Adapter/MysqlAdapter.php:1005`, you will see on the case statements that there are handles for tiny,small and bigints, so that you can specify the limit, and it will write the query as, for example, `tinyint(2)`. This works for all special mysql types, except `mediumint`.

The fix appears to be easy, just add another case statement and a procedure for when we don't pass the limit (mediumint is the only one with a value on the `$limit` array.

dereuromark:
Are you willing to make a PR here.

yohanleafheart:
Hey dereu, I am. I will make the change and do a pull request

PR #1997, opened 2021-06-28
Adding the mediumint type for the MySql adapter

yohanleafheart:
This Pull requests solves issue #1996. Correctly adding the mediumint type to mysql. Together on this patch, one of the PHPUnit tests was changed to ignore case, due to differences between MySQL and MariaDB

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -47,6 +47,7 @@ class MysqlAdapter extends PdoAdapter
         self::PHINX_TYPE_INTEGER => true,
         self::PHINX_TYPE_TINY_INTEGER => true,
         self::PHINX_TYPE_SMALL_INTEGER => true,
+        self::PHINX_TYPE_MEDIUM_INTEGER => true,
         self::PHINX_TYPE_BIG_INTEGER => true,
         self::PHINX_TYPE_FLOAT => true,
         self::PHINX_TYPE_DECIMAL => true,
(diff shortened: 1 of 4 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1000,6 +1001,8 @@ public function getSqlType($type, $limit = null)
                 return ['name' => 'bit', 'limit' => $limit ?: 64];
             case static::PHINX_TYPE_SMALL_INTEGER:
                 return ['name' => 'smallint', 'limit' => $limit ?: 6];
+            case static::PHINX_TYPE_MEDIUM_INTEGER:
```
dereuromark, line 1004: Shouldnt this be above small? 8, 6, 4

yohanleafheart: We can change, but since we are returning on each case check, should not make a difference, should it?

yohanleafheart: hahaha Know the feeling mate. Now, dumb question from someone who is doing their first github pull request, do I need to do another commit on my fork and another pull request or can I edit here to fix it?

dereuromark: A commit on top is fine

yohanleafheart: Done. I also moved the big int that was after the integer statement.

PR #2006, opened 2021-08-01
Update MysqlAdapter.php

CrazyTapok-bit:
When working with PDO in PDO :: ATTR_DEFAULT_FETCH_MODE => PDO :: FETCH_ASSOC mode, an error pops up. Cannot use an object of type stdClass as an array

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -470,7 +470,8 @@ class MysqlAdapter extends PdoAdapter
     {
         $rows = $this->fetchAll(sprintf('SHOW COLUMNS FROM %s', $this->quoteTableName($tableName)));
         foreach ($rows as $column) {
-            if (strcasecmp($column['Field'], $columnName) === 0) {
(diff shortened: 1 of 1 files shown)


MasterOdin (changes requested):
There's probably a number of places that phinx will fail on when using `PDO::FETCH_ASSOC`, both within mysql as well as within other adapters.

A better solution might be to modify:
https://github.com/CrazyTapok-bit/phinx/blob/a69d66de09b0cddc2c0bb97c853a68b46b7c50fb/src/Phinx/Db/Adapter/PdoAdapter.php#L225-L236

to something like:
```php
    public function fetchRow($sql, $mode = PDO::FETCH_DEFAULT)
    {
        return $this->query($sql)->fetch($mode);
    }

    /**
     * @inheritDoc
     */
    public function fetchAll($sql, $mode = PDO::FETCH_DEFAULT)
    {
        return $this->query($sql)->fetchAll($mode);
    }
```

and then wherever phinx internally calls those two functions, pass `PDO::FETCH_ASSOC` as the second argument.

dereuromark:
This will require a different strategy, closing for now.

PR #2012, opened 2021-09-02
Copy GA CI triggers from cakephp

othercorey:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,15 @@
...
+on:
+  push:
+    branches:
+      - 'master'
+      - '0.next'
+  pull_request:
+    branches:
+      - '*'
+  schedule:
+    - cron: '0 0 * * *'
```
dereuromark, line 12: I started to remove cron from my plugins
If there is no input after xx days, it will deactivate the whole CI pipeline and no further PRs would work anymore
It would require manual activation again, quite annoying.

othercorey: Ok, removed due to low commit rate.

MasterOdin:
This is a reverse of #1859 and as stated there, I like having GH actions running on all pushes as it allows me to validate the changes before I make a PR. The commit rate on this repo is also low enough and the suite fast enough that I'm not sure what's the gain for removing it?

dereuromark:
What's the issue of using draft PRs instead? would that work and run CI?

dereuromark:
Given that this repo is indeed not very high in commits, I am fine either way.
It just feels weird to have xx runs extra on each PR commit, as it is fully duplicated there.

othercorey:
It's strange to have a different setup just for phinx, but I'll let someone else approve and merge the PR.

dereuromark:
As long as this repo remains this "dead" I guess we can skip the filtering for now.

PR #2013, opened 2021-09-04
Allow specifying template style for create command

### LINKED ISSUE #2009: Include alternative migration template for up/down dichotomy
By default, the `create` command produces a migration template based on the `change()` method. However, I always delete this and replace it with `up()`/`down()` instead. Although the _create_ command has support for custom templates, they're not particularly intuitive to write, so I propose shipping an additional template with Phinx to support the up/down dichotomy since it may be popular enough to warrant it, given that it allows us to write raw SQL migrations. In that case, we could pass `--sql`, `--rawsql` or some other command line switch to use the alternative template.

MasterOdin:
This PR adds a new config option and runtime flag for allowing to specify the template style you want to use, either `change` or `up_down`, with the former being the default value. This can be used by people who want the older style of `up` and `down` functions over the `change` function.

This PR is against `0.next` due to breaking changes of:

* Renaming const in `Create` console class.
* Renaming the filename for the migration template.

Closes #2009.


MasterOdin:
Codecov failing seems like a bug on their end as all lines in this changeset are covered by tests.

dereuromark:
Any other breaking changes we want to include and ship?
Maybe in the next few weeks.

Bilge:
Is this documented? I can't find any reference to `template` in the [documentation](https://book.cakephp.org/phinx/0/en/index.html). 

If I run `vendor/bin/phinx create -t up_down Foo` it gives me:

>In Create.php line 220:
>  The alternative template file "up_down" does not exist

src/Phinx/Console/Command/Create.php
```diff
@@ -57,6 +58,8 @@ protected function configure(): void
 
         // Allow the migration path to be chosen non-interactively.
         $this->addOption('path', null, InputOption::VALUE_REQUIRED, 'Specify the path in which to create this migration');
+
+        $this->addOption('style', null, InputOption::VALUE_REQUIRED, 'Specify the style of migration to create');
```
dereuromark, line 62: This creates a BC issue for https://github.com/cakephp/migrations/actions/runs/3583067226/jobs/6028049748 it seems for new release of supporting 0.13

Could we just default to one somewhere maybe? Not having it required?

PR #2016, opened 2021-09-13
Fixes on php8.1

CriztianiX:


src/Phinx/Config/Config.php
```diff
@@ -430,9 +430,11 @@ protected function replaceTokens(array $arr)
             }
         }
 
+        $configFilePath = $this->getConfigFilePath();
+
         // Phinx defined tokens (override env tokens)
         $tokens['%%PHINX_CONFIG_PATH%%'] = $this->getConfigFilePath();
-        $tokens['%%PHINX_CONFIG_DIR%%'] = dirname($this->getConfigFilePath());
+        $tokens['%%PHINX_CONFIG_DIR%%'] = is_null($configFilePath) ? '' : dirname($configFilePath);
```
dereuromark, line 437: !== null

CriztianiX: done

src/Phinx/Config/Config.php
```diff
@@ -455,6 +455,7 @@ protected function recurseArrayForTokens($arr, $tokens)
             }
             if (is_string($value)) {
                 foreach ($tokens as $token => $tval) {
+                    $tval = $tval === null ? '' : $tval;
                     $value = str_replace($token, $tval, $value);
```
MasterOdin, line 459: Could simplify with a null coalescing within where we use `$tval`.

```suggestion
                    $value = str_replace($token, $tval ?? '', $value);
```

MasterOdin (changes requested):

PR #2026, opened 2021-10-12
Fix setting default for MySQL Blob, Geometry, Json, and Text

### LINKED ISSUE #1999: MySQL default value BLOB, TEXT, GEOMETRY or JSON
MySQL version 8.0.22

Update script:
```php
$users = $this->table('users');
$users->addColumn('test', 'json', ['default' => '{}']);
$users->update();
```
SQL:
```sql
ALTER TABLE `users`  ADD `test` JSON NOT NULL DEFAULT '{}';
```
Not great error by MySQL (it is not true, just wrong syntax):
`BLOB, TEXT, GEOMETRY or JSON column 'test' can't have a default value`

Documentation:
https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html

`The BLOB, TEXT, GEOMETRY, and JSON data types can be assigned a default value only if the value is written as an expression, even if the expression value is a literal`

Expected syntax:
```sql
ALTER TABLE `users`  ADD `test` JSON NOT NULL DEFAULT ('{}');
```

MasterOdin:
Fixes #1999 

This PR adds special casing to the four mentioned types that wraps their default value in an expression (adds `('` and `')` around the default value), such that it gets set. Technically this only works under MySQL 8+, with prior versions throwing an exception. The thrown exception has not changed either in type or message, so I would not consider this a BC break.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,14 +89,14 @@ jobs:
         if [[ ${{ matrix.db-type }} == 'mysql' ]]; then export MYSQL_DSN='mysql://root:root@127.0.0.1/phinx'; fi
         if [[ ${{ matrix.db-type }} == 'pgsql' ]]; then export PGSQL_DSN='pgsql://postgres:postgres@127.0.0.1/phinx'; fi
 
-        if [[ ${{ matrix.php-version }} == '7.2' ]]; then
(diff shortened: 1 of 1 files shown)


MasterOdin:
Actually, it looks like code paths are purely additive, so fine to just run code coverage only on php 8.0 (which uses latest version of MySQL and SQLite, the only DBs that version gate) for now.

PR #2032, opened 2021-10-15
Support using mysql int constants on same types

### LINKED ISSUE #1973: INT_BIG gets interpreted as 1.844674407371E+19 instead of 18446744073709551615
This addColumn:

`            ->addColumn('id', 'biginteger', [
                'null' => false,
                'limit' => MysqlAdapter::INT_BIG,
                'identity' => 'enable',
            ])
`
results in this error:

`you have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '1.844674407371E+19)`

When I debug it, stepping into the addColumn function, the debugger shows that $option['limit'] is in fact `1.844674407371E+19`

I've had this code for months, deployed it, used these tables, all without this problem.  I'm not sure why it suddenly started now.  I'm guessing it's a php setting, not necessarily a problem with phinx.

If it is a setting in php, phinx checking that setting and throwing a warning would be nice.

Yes, I've googled this, but "php converting int to scientific notation" gives a lot of pages on what scientific notation is, not how to prevent `int`s from being converted to it.

I tried using 18446744073709551615 instead of using the const, same issue.  I tried using it as a string, it caused other problems.

To work around it, I had to step it down to `integer` type and `MysqlAdapter::INT_REGULAR`, not what I was hoping for.

### LINKED ISSUE #2030: Broken limit value for biginteger in alter clause
Migration code
```
        $this->table('table')
            ->changeColumn('exp', 'biginteger',
                [
                    'null' => false,
                    'limit' => \Phinx\Db\Adapter\MysqlAdapter::INT_BIG,
                ])
            ->update();
```
Generated SQL:
```
ALTER TABLE `table` CHANGE `exp` `exp` BIGINT(1.844674407371E+19) NOT NULL;
```
It shouldn't convert limit to float format... MariaDB won't accept such request. Dropping `limit` param doesn't help. Also shouldn't it be converted to `BIGINT(20)`?
Tried on version 0.12.8 and 0.12.9.

MasterOdin:
Fixes #1973
Fixes #2030 

This PR adjusts how `MysqlAdapter::getSqlType` function such that we can now pass the `INT_*` constants to work with their corresponding types and get the default limit instead of whatever the value of `INT_*` constant was. For example, `MysqlAdapter::getSqlType(AdapterInterface::PHINX_TYPE_BIG_INTEGER, MysqlAdapter::INT_BIG)` will return `['bigint', 20]` as opposed to `['bigint', MysqlAdapter::INT_BIG]`. Behavior remains the same for the `AdapterInterface::PHINX_TYPE_INTEGER` type and how it handles the various `INT_*` constants.

### DIFF
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -884,9 +884,8 @@ class MysqlAdapterTest extends TestCase
         $this->assertNull($rows[1]['Default']);
     }
 
-    public function sqlTypeProvider()
+    public function sqlTypeIntConversionProvider()
     {
-        // TODO: add tests for missing types
         return [
           // tinyint
           [AdapterInterface::PHINX_TYPE_TINY_INTEGER, null, 'tinyint', 4],
@@ -916,10 +915,10 @@ class MysqlAdapterTest extends TestCase
     }
 
     /**
-     * @dataProvider sqlTypeProvider
+     * @dataProvider sqlTypeIntConversionProvider
      * The second argument is not typed as MysqlAdapter::INT_BIG is a float, and all other values are integers
      */
-    public function testGetSqlType(string $type, $limit, string $expectedType, int $expectedLimit)
+    public function testGetSqlTypeIntegerConversion(string $type, $limit, string $expectedType, int $expectedLimit)
     {
         $sqlType = $this->adapter->getSqlType($type, $limit);
         $this->assertSame($expectedType, $sqlType['name']);


MasterOdin:
@dereuromark @markstory I cannot see why stickler-ci is failing. Clicking on the details link takes me to a not found page that states "The repository you were looking for could not be found.". I am logged in via GH.

othercorey:
> @dereuromark @markstory I cannot see why stickler-ci is failing. Clicking on the details link takes me to a not found page that states "The repository you were looking for could not be found.". I am logged in via GH.

Comment refers to a TODO task "add tests for missing types"

We don't generally allow TODO in comments since it isn't tracked.

PR #2033, opened 2021-10-15
Fix up as per PHPStan/Psalm

dereuromark:
These are fixes need for Migrations integration
https://github.com/cakephp/migrations


dereuromark:
I fixed up also a few more PHPStan level 7
71 more issues until we can use that level (and 147 for level 8 left).

Issue #2037, opened 2021-10-26
MySQL Smallint not extracting correct limit from SHOW TABLES


oprose:
I've recently updated from an older version to 0.12.9, and one of the migrations we used to build our schema used the setColumns/getColumns pair to effectively create a copy of a table. As these have been removed from newer versions, I've needed to recreate their actions to recreate the target table structure.

What I've discovered is that the getColumns method now seems to set an invalid limit value for smallint (and possibly medium and tiny columns too), as it returns the maximum value for those types instead of the target length.

Example Table Structure
`CREATE TABLE a (
 i INT(10) UNSIGNED,
 i2 INT(11) UNSIGNED,
 s SMALLINT(5) UNSIGNED
);`

$table->getColumns() returns:
`
[
 0 => ['name' => 'i', 'type' => 'integer', 'limit' => 10 ...],
 1 => ['name' => 'i2', 'type' => 'integer', 'limit' => null ...],
 2 => ['name' => 's', 'type' => 'smallint', 'limit' => 65535 ...]
]
`

The limit on column s should be 5 I believe, but I'm not ruling out user error yet. I think the problem might be MysqlAdapter::getPhinxType(), in the switch starting on line 1086.

PR #2038, opened 2021-10-27
Fix MysqlAdapter::getColumns returning physical limit instead of display width for int columns

oprose:
See cakephp/phinx#2037


MasterOdin (approved):
I'm good with this change (after removing the commented out `$limit` assignment lines.

PR #2040, opened 2021-11-13
Maintain indexes / triggers after altering sqlite tables

### LINKED ISSUE #1290: [SQLite Adapter] Triggers are dropped on table/column modification
When modifying tables or columns in SQLite, a lot of operations require a new temporary table to be made. The structure and the data is then moved from one table to another. Currently this does not bring the table triggers along with it.

There is a closed PR that demonstrates one way of fixing the issue: https://github.com/cakephp/phinx/pull/1284
However it was deemed that an alteration of the SQLite adapter is the right way to go about fixing it so it was closed. Creating an issue to highlight the problem and maybe somebody will want to re-write the adapter.

MasterOdin:
PR fixes a bug where altering sqlite tables that had indexes or triggers would drop the indexes or triggers. This was because indexes and triggers are not contained in the `CREATE TABLE` sql that we used to create a table copy the indexes and triggers would be get copied over, and so on dropping the original table it would delete the indexes and triggers, and the rename would not bring them back.

This PR adds new post steps of recreating the indexes and triggers after we rename the temporary table we created back to its original name.

Fixes #1290

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -832,12 +832,23 @@ PCRE_PATTERN;
                 $state['selectColumns']
             );
 
+            $rows = $this->fetchAll(
+                sprintf(
+                    "SELECT * FROM sqlite_master WHERE `type` = 'index' OR `type` = 'trigger' AND tbl_name = %s",
+                    $this->quoteValue($tableName)
+                )
+            );
+
             $this->execute(sprintf('DROP TABLE %s', $this->quoteTableName($tableName)));
             $this->execute(sprintf(
                 'ALTER TABLE %s RENAME TO %s',
                 $this->quoteTableName($state['tmpTableName']),
                 $this->quoteTableName($tableName)
             ));
+
+            foreach ($rows as $row) {
+                $this->execute($row['sql']);
+            }
 
             return $state;
         });
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -834,7 +834,14 @@ PCRE_PATTERN;
 
             $rows = $this->fetchAll(
                 sprintf(
-                    "SELECT * FROM sqlite_master WHERE `type` = 'index' OR `type` = 'trigger' AND tbl_name = %s",
+                    "
+                        SELECT *
+                        FROM sqlite_master
+                        WHERE
+                            (`type` = 'index' OR `type` = 'trigger')
+                            AND tbl_name = %s
+                            AND sql IS NOT NULL
+                    ",
                     $this->quoteValue($tableName)
                 )
             );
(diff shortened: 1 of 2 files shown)


dereuromark:
Should we finish up the 0.12 milestone with last release and then do 0.13?
I wonder if we can make 0.12 already Symf6 and PHP8.1 compatible, that would probably also be best

MasterOdin:
@dereuromark I thought that https://github.com/cakephp/migrations/pull/517 would need to be merged before releasing 0.13, but I'm not sure the status of that PR.

MasterOdin:
I do plan to finish this up with the caveat of disabling it for renaming or dropping columns as those requiring potentially rewriting the trigger or index. I figure it's at least a step in the right direction, and document these cases for now.

dereuromark:
Shall we wait on this for the next patch release? Or release already as is?

MasterOdin:
This is not yet ready for release (as evidenced by the broken tests), and I have not had time to come back to this yet. We should not have this PR hold up a new release.

ndm2:
Sorry for unsolicitedly sliding in here, but I need this in order to get stuff in `cakephp/migrations` fixed/tested 🥲

MasterOdin:
@ndm2 Do you think the PR is ready to merge now? Given that all existing tests are passing, and two new tests added for coverage on this, I think should be safe?

ndm2:
It looks pretty safe... famous last words.

This is assuming that the SQL in the master table is safe to run as-is, and that indices and triggers are always being scrubbed on table drops.

Maybe a look into temporary triggers is advised, I'm not sure how exactly they function and how they are stored?

MasterOdin:
This follows the official method for doing these kinds of operations as described in the [sqlite docs](https://sqlite.org/lang_altertable.html#otheralter), so I'm not particularly concerned on getting invalid SQL from the `sqlite_master` table, or that triggers / indexes aren't removed on dropping a table.

From that doc, they do mention views, and not sure if we have a test case for that.

ndm2:
Does Phinx support views at all? Altering a table could of course affect a view, but I don't think there's much that Phinx can do, as views are arbitrary select statements, and aren't tied to specific tables (which has the effect that they're also not being deleted when tables are dropped).

MasterOdin:
Phinx doesn't natively support views, and I'm not sure what sqlite (or other DBs) do if you drop a table that a view references. In theory, with a SQL statement parser, it might be possible to apply migrations and such to view statements automatically when you rename a table or column or whatever, but that's out of scope for this bugfix.

ndm2:
Maybe it wasn't that safe after all. I see this breaking when columns are being removed as part of a rollback, where it will try to re-add indices for removed columns :/

But I'm a bit irritated, as I thought that reversed plans would drop indices before dropping columns 🤔

MasterOdin:
Can you open a new bug report with a migration that demonstrates that bug?

ndm2:
Will do once I've figured if it's maybe something I or `cakephp/migrations` is doing wrong.

Issue #2052, opened 2021-12-31
Adding support for the `mode` query parameter in Sqlite (PHP 8.1)


swiffer:
PHP Sqlite allows setting `mode` and `cache` parameters. While adding support for `cache` was straight forward I struggled a little bit with the mode parameter and the options that are currently possible.

Currently for a sqlite connection you can set `name` and `memory` options. If `memory` has been set to anything `name` is overwritten to `:memory:`.

https://github.com/cakephp/phinx/blob/master/tests/Phinx/Config/ConfigTest.php#L344

What is the benefit of having the memory option and not specifying `:memory:` in the `name` directly?

PHP 8.1+ adds support for using `mode` query parameter when creating a connection. Wouldn't it be better to keep the `name` or use the name specified in `memory` if not `true` as name for the database?

https://github.com/cakephp/phinx/blob/master/src/Phinx/Db/Adapter/SQLiteAdapter.php#L159

```php
                if($options['memory'] !== true && PHP_VERSION_ID >= 80100) {
                    $options['name'] = $options['memory'];
                    $params[] = 'mode=memory';
                } else {
                    $options['name'] = static::MEMORY;
                }
```

MasterOdin:
Looking at the PHP docs for [PDO SQLite](https://www.php.net/manual/en/ref.pdo-sqlite.php), I do not see either of these options mentioned. Could you provide a link to somewhere that talks about them, and what their meaning and values could be?

othercorey:
It's not part of PDO specifically, they just added support for URI filenames. That is required to use filenames with options when opening the db.

https://www.sqlite.org/sharedcache.html

MasterOdin:
From the above example, what does `sqlite:foo.db?mode=memory` do? Is it to allow opening multiple different shared caches for in-memory databases?

othercorey:
That is creating a unique in-memory db using the name `foo.db` so you can open and share caches with multiple dbs.

othercorey:
The more common use case will simply be `file::memory:?cache=shared`

MasterOdin:
For `cache=shared` for a non inline memory DB, do we need to use `file:` there as well? When should `file:` be prefixed automatically?

othercorey:
So, all URI filenames have to be prefixed with `file:`. Support for this is disabled by default, but PDO enables it as long as open_basedir isn't configured.

[...]

PR #2053, opened 2021-12-31
support cache query parameter

swiffer:
Adds support for the `cache` query parameter - the same has been added in CakePHP already and e.g. allows using the same in memory database when running a CakePHP App TestSuite via Migrator.

CakePHP also adds support for a named inmemory database via the `mode` query parameter. Adding support for this one needs some discussion in #2052

Something similar has been done for CakePHP already here: https://github.com/cakephp/cakephp/pull/16189


dereuromark:
@MasterOdin Is this OK to be merged as is?

MasterOdin:
@dereuromark It would be good to come up with a way to test this that isn't just try opening a bunch of sqlite connections with different parameters on 8.1+. Could we expose a `public function getDsn()` function for all adapters? Could also make the function private, and then test it via reflection, but either way, I think just making sure we're putting the string together correctly is going to be necessary to make sure we don't have regressions here as the functionality is somewhat complicated.

dereuromark:
Sure, we can also check non private ways for the new 0.13 branch/release.
And keep things simple for last minor version of 0.12

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -155,11 +155,26 @@ public function connect()
...
+                throw new RuntimeException('Memory must not be set when cache or mode are.');
+            } elseif (PHP_VERSION_ID >= 80100) {
+                $params = [];
+                if (!empty($options['cache'])) {
+                    $params[] = 'cache=' . $options['cache'];
+                }
+                if (!empty($options['mode'])) {
+                    $params[] = 'mode=' . $options['mode'];
+                }
+                $dsn = 'sqlite:file:' . $options['name'] . '?' . implode('&', $params);
```
MasterOdin, line 170: Are we requiring that `name` always be set if using `mode` or `cache`? Does sqlite support `sqlite:file?...`? I'd think so, so we should probably throw a warning here if `name` is unset, or maybe default to `phinx`, e.g. `$options['name'] ?? 'phinx'`?

MasterOdin (changes requested):

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -155,11 +155,26 @@ public function connect()
...
             $options = $this->getOptions();
 
-            // use a memory database if the option was specified
-            if (!empty($options['memory']) || $options['name'] === static::MEMORY) {
-                $dsn = 'sqlite:' . static::MEMORY;
+            if (PHP_VERSION_ID < 80100 && (!empty($options['mode']) || !empty($options['cache']))) {
+                throw new RuntimeException('SQLite URI support requires PHP 8.1.');
+            } elseif ((!empty($options['mode']) || !empty($options['cache'])) && !empty($options['memory'])) {
+                throw new RuntimeException('Memory must not be set when cache or mode are.');
+            } elseif (PHP_VERSION_ID >= 80100) {
```
MasterOdin, line 162: We shouldn't be forcing everyone on PHP 8.1+ into using `mode` and `cache`. They should be able to use the existing `name` / `memory` only options if they want.

```suggestion
            } elseif (PHP_VERSION_ID >= 80100 && (!empty($options['mode']) || !empty($options['cache'])) {
```

MasterOdin (changes requested):

src/Phinx/Db/Adapter/SQLiteAdapter.php
swiffer: it is supported with an empty filename:

https://github.com/php/php-src/pull/7662/files#diff-0223830cccc48601cf8cb19d5bd6dc243d44b24f2c0aca2776c7f15d0ced8878R13

what would a user expect if it's not set via config - i guess leaving it empty would be right?

src/Phinx/Db/Adapter/SQLiteAdapter.php
swiffer: changed that

src/Phinx/Db/Adapter/SQLiteAdapter.php
MasterOdin: Makes sense to me, so `$options['name'] ?? ''` should be fine?

swiffer: done

swiffer:
> @dereuromark It would be good to come up with a way to test this that isn't just try opening a bunch of sqlite connections with different parameters on 8.1+. Could we expose a `public function getDsn()` function for all adapters? Could also make the function private, and then test it via reflection, but either way, I think just making sure we're putting the string together correctly is going to be necessary to make sure we don't have regressions here as the functionality is somewhat complicated.

is someone from your end picking this up or can you give a little more guidenance on the way this should be implemented?

MasterOdin (approved):

MasterOdin:
> is someone from your end picking this up or can you give a little more guidenance on the way this should be implemented?

This would be done in a follow-up PR, no changes necessary from you here.

PR #2057, opened 2022-01-04
improve tests for create command

LordSimal:
This PR will slightly improve the test coverage for the `Create` command.

You may also ask why I removed 2 sections related to path checking.
The reason behind that is the fact, that `getMigrationPaths()` already checks if the given (or default) path is present and valid.

If not it already throws an exception [here](https://github.com/cakephp/phinx/blob/master/src/Phinx/Console/Command/Create.php#L105) and therefore can never get to those other 2 checks.


dereuromark:
Prefer lowest was failing, my bad for merging this.

PR #2063, opened 2022-01-21
increase dependency versions

saeideng:
Refs https://github.com/cakephp/phinx/pull/2057/files


dereuromark:
We can also revert that commit otherwise, if we cannot find a good way to fix this up.

saeideng:
yes , we can revert it , until next time

PR #2065, opened 2022-01-21
improve test-coverage of Create Console Command

LordSimal:
Basically the same as #2057 but without removing code from the `src` directory

It seems like those are required for 7.2 prefer-lowest.


LordSimal:
Nope, still the same error...

dereuromark:
Maybe Tests are wrong or sth?

PR #2073, opened 2022-03-08
Bump phpstan

saeideng:


dereuromark:
Looks like it needs some adjusting

PR #2074, opened 2022-03-09
Remove out-of-date translated docs

MasterOdin:
The `es`, `fr`, and `jp` translations, while nice to have in theory, are so far out of date to probably be detrimental. Without either a larger contributor community around phinx, or a regular maintainer willing to dedicate to maintaining these docs, I see no reason to keep them and let them grow more and more out of date.

As such, this PR removes the pages as well as building them altogether from phinx. While it's regrettable, at least no one will risk getting docs that are flat out wrong for something phinx does or supports.


dereuromark (approved):
I am fine with it, but we also need to ask the rest of the team here.

aspirantzhang:
I have seen that .rst files can be translated "line by line" using .po files. I think this is a [good solution](https://docs.readthedocs.io/en/stable/guides/manage-translations-sphinx.html).  I wonder why CakePHP doesn't take this approach?
However, in the process of working with Chinese documents, I also found that files in other languages were missing a lot of content, so it might be a good idea to delete them all.

PR #2075, opened 2022-03-13
TEXT WRAPPER: Refactor shorthand ternary & add tests

sjokkateer:
Currently, within the `TextWrapper::class`, in the methods `TextWrapper::getStatus()`, `TextWrapper::getMigrate()`, `TextWrapper::getSeed()`, `TextWrapper::getRollback()` the following expression: `if ($env ?: $this->hasOption('environment'))` is used.

If I am correct, that would mean that the inner short hand ternary would first have to evaluate before the outer if would evaluate its result. Hence there would be two evaluations of the value's truthiness. This pull request removes the use of the shorthand ternary and replaces it by a logical or.

The added tests should confirm that equivalent behavior is maintained (although tests are written for one method implementation).

In case there has been anything I overlooked please inform me, or if you have any additional requests please let me know!

### DIFF
--- a/src/Phinx/Wrapper/TextWrapper.php
+++ b/src/Phinx/Wrapper/TextWrapper.php
@@ -73,7 +73,7 @@ class TextWrapper
     public function getStatus($env = null)
     {
         $command = ['status'];
-        if ($env ?: $this->hasOption('environment')) {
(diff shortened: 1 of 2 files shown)


dereuromark:
Makes sense

MasterOdin (approved):

PR #2077, opened 2022-03-19
remove faker document

### LINKED ISSUE #2076: Seeding documentation should replace fzaninotto/faker with fakerphp/faker
fzaninotto/faker has been abandoned since 2020-12-11. The Phinx seeding documentation should instead reference fakerphp/faker, the replacement that is in active development.

saeideng:
Close #2076


MasterOdin (approved):
Should be easy enough for someone to figure this out for themselves that we don't need explicit documentation on this.

PR #2082, opened 2022-03-29
make sure options are still appended when 'name' is empty

aa-matthias:
make sure options are still appended when 'name' is empty


dereuromark:
@saeideng what do you think?

PR #2083, opened 2022-03-29
make sure params are applied

swiffer:


dereuromark:
Looks like you need to fix up the phpstan baseline or alike.

dereuromark (approved):

PR #2090, opened 2022-04-25
Fix SqlServerAdapter returning empty string instead of null for column default

reeperbahnause:
The `SqlServerAdapter::parseDefault($default)` function returns `""` if `$default` is `null`
Because of the returned empty string, the column can not be **changed** to have an empty string as default value (no differnce detected)

Example for an existing column which has no effect:
```php
$this->table('tableName')->changeColumn('columnName', 'string', ['default' => '', 'null' => false])
```

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -477,16 +477,18 @@ class SqlServerAdapter extends PdoAdapter
     }
 
     /**
-     * @param string $default Default
+     * @param string|null $default Default
      * @return int|string|null
      */
     protected function parseDefault($default)
     {
+        // if a column is non-nullable and has no default, the default is null, otherwise
+        // it should be a string value that we parse below
         if ($default === null) {
-            $result = 'NULL';
-        } else {
-            $result = preg_replace(["/\('(.*)'\)/", "/\(\((.*)\)\)/", "/\((.*)\)/"], '$1', $default);
+            return null;
         }
+
+        $result = preg_replace(["/\('(.*)'\)/", "/\(\((.*)\)\)/", "/\((.*)\)/"], '$1', $default);
 
         if (strtoupper($result) === 'NULL') {
             $result = null;
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -482,7 +482,11 @@ class SqlServerAdapter extends PdoAdapter
      */
     protected function parseDefault($default)
     {
-        $result = preg_replace(["/\('(.*)'\)/", "/\(\((.*)\)\)/", "/\((.*)\)/"], '$1', $default);
+        if ($default === null) {
+            $result = 'NULL';
+        } else {
+            $result = preg_replace(["/\('(.*)'\)/", "/\(\((.*)\)\)/", "/\((.*)\)/"], '$1', $default);
+        }
 
         if (strtoupper($result) === 'NULL') {
             $result = null;
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -482,8 +482,9 @@ class SqlServerAdapter extends PdoAdapter
      */
     protected function parseDefault($default)
     {
-        // if a column is non-nullable and has no default, the default is null, otherwise
-        // it should be a string value that we parse below
+        // if a column is non-nullable and has no default, the value of column_default is null,
+        // otherwise it should be a string value that we parse below, including "(NULL)" which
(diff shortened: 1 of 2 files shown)


MasterOdin:
What version of SQLServer are you using, and how are you getting a column that has `$default === null`? Just want to understand where the current test I'd expect to catch this (https://github.com/cakephp/phinx/blob/48d8e1f686852882a319e6deff1ae6748d04ab5c/tests/Phinx/Db/Adapter/SqlServerAdapterTest.php#L530-L542) is failing.

reeperbahnause:
The SQL Server version is 15.0.2000

And the statement which was executed by phinx to get the current table structure was this (`SqlServerAdapter::getColumns()`):
```sql
SELECT DISTINCT TABLE_SCHEMA AS [schema], TABLE_NAME as [table_name], COLUMN_NAME AS [name], DATA_TYPE AS [type],
            IS_NULLABLE AS [null], COLUMN_DEFAULT AS [default],
            CHARACTER_MAXIMUM_LENGTH AS [char_length],
            NUMERIC_PRECISION AS [precision],
            NUMERIC_SCALE AS [scale], ORDINAL_POSITION AS [ordinal_position],
            COLUMNPROPERTY(object_id(TABLE_NAME), COLUMN_NAME, 'IsIdentity') as [identity]
        FROM INFORMATION_SCHEMA.COLUMNS
        WHERE TABLE_NAME = 'tableName'
        ORDER BY ordinal_position
```

MasterOdin:
Sorry, when I said "how are you getting the column", I meant the phinx command that you were using to create the column, or is the column coming from somewhere else initially?

I'd like to make sure we have a test here for this code change, just not sure what to suggest for such a test and trying to gain a better understanding of the issue.

reeperbahnause:
The tabe is initialy created like this:
```php
        $table->addColumn('login', 'string', array('limit' => 50))
            ->addColumn('firstname', 'string', array('limit' => 100))
            ->addColumn('lastname', 'string', array('limit' => 100))
            ->addColumn('status', 'string', array('limit' => 45))
            ->addColumn('hash', 'string', array('limit' => 255))
            ->addColumn('email', 'string', array('limit' => 255))
            ->addColumn('network', 'string', array('limit' => 255))
            ->addColumn('networkid', 'string', array('limit' => 255))
            ->addColumn('profileurl', 'string', array('limit' => 255))
            ->addColumn('imageurl', 'string', array('limit' => 255))
            ->addColumn('company', 'string', array('limit' => 255))
            ->addColumn('position', 'string', array('limit' => 255))
            ->addColumn('country', 'string', array('limit' => 255))
            ->addColumn('city', 'string', array('limit' => 255))
            ->addColumn('street', 'string', array('limit' => 255))
            ->addColumn('resettoken', 'string', array('limit' => 64, 'null' => true))
            ->create();
```

and the migration which did not have any effect was this:
```php
        $table->changeColumn('status', 'string', ['default' => '', 'null' => false])
            ->changeColumn('network', 'string', ['default' => '', 'null' => false])
            ->changeColumn('networkid', 'string', ['default' => '', 'null' => false])
            ->changeColumn('profileurl', 'string', ['default' => '', 'null' => false])
            ->changeColumn('imageurl', 'string', ['default' => '', 'null' => false])
            ->changeColumn('company', 'string', ['default' => '', 'null' => false])
            ->changeColumn('position', 'string', ['default' => '', 'null' => false])
            ->changeColumn('country', 'string', ['default' => '', 'null' => false])
            ->changeColumn('street', 'string', ['default' => '', 'null' => false])
            ->changeColumn('resettoken', 'string', ['default' => '', 'null'=>true])
            ->changeColumn('city', 'string', ['default' => '', 'null' => false])
        ->update();
```
Does that help?
If not let me know. I will get back in an hour.

reeperbahnause:
Can I provide any more information regarding this issue?

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -482,7 +482,11 @@ public function getColumns($tableName)
      */
     protected function parseDefault($default)
     {
-        $result = preg_replace(["/\('(.*)'\)/", "/\(\((.*)\)\)/", "/\((.*)\)/"], '$1', $default);
+        if ($default === null) {
```
dereuromark, line 485: Then docblock should state also null as possible param value

MasterOdin: Done in 593af76

MasterOdin (approved):
Sorry for the delay in getting to this @reeperbahnause! I've added a test case that verifies the change and made some slight tweaks, but the fix looks good. Seems like the bug was around that a `NOT NULL` column with no default would have a `null` value for `column_default`, while a `NULL` column with a null default would have `"(NULL)"` for `column_default`. I'm sure there's some important reason for this, but weird nevertheless.

This should be put out in a release soon.

PR #2092, opened 2022-05-10
Fix setting persistent option for PDO connection

paulermo:
This fixes the ability to use persistent connections by PDO. 
One can set generic PDO attributes (#1904) but the attribute - \PDO::ATTR_PERSISTENT - is useless while setting it by setAttribute after PDO object instantiation:

> Note:
> If you wish to use persistent connections, you must set PDO::ATTR_PERSISTENT in the array of driver options passed to the PDO constructor. If setting this attribute with [PDO::setAttribute()](https://www.php.net/manual/en/pdo.setattribute.php) after instantiation of the object, the driver will not use persistent connections.
[https://www.php.net/manual/en/pdo.connections.php](url)

From the point of view of usage, nothing changes: like other PDO arguments we specify `attr_persistent => true` config option. Before the PDO instatiation this attribute (if specified) is passed to driver options instead of setting in setAttribute after it.

It's done in Mysql, Postgres and SQLite adapters. SQL Server PDO driver does not support this specific attribute.

### DIFF
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -2214,4 +2214,16 @@ INPUT;
         $this->assertSame($expectedResponse['name'], $result['name'], "Type mismatch - got '{$result['name']}' when expecting '{$expectedResponse['name']}'");
         $this->assertSame($expectedResponse['limit'], $result['limit'], "Field upper boundary mismatch - got '{$result['limit']}' when expecting '{$expectedResponse['limit']}'");
     }
+
+    public function testPdoPersistentConnection()
+    {
+        $adapter = new MysqlAdapter(MYSQL_DB_CONFIG + ['attr_persistent' => true]);
+        $this->assertEquals(true, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
+    }
+
+    public function testPdoNotPersistentConnection()
+    {
+        $adapter = new MysqlAdapter(MYSQL_DB_CONFIG);
+        $this->assertEquals(false, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
+    }
 }
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -2218,12 +2218,12 @@ INPUT;
     public function testPdoPersistentConnection()
     {
         $adapter = new MysqlAdapter(MYSQL_DB_CONFIG + ['attr_persistent' => true]);
-        $this->assertEquals(true, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
+        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
     }
 
     public function testPdoNotPersistentConnection()
     {
         $adapter = new MysqlAdapter(MYSQL_DB_CONFIG);
-        $this->assertEquals(false, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
+        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
     }
 }
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -2218,12 +2218,12 @@ INPUT;
     public function testPdoPersistentConnection()
     {
         $adapter = new MysqlAdapter(MYSQL_DB_CONFIG + ['attr_persistent' => true]);
-        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
+        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
     }
 
     public function testPdoNotPersistentConnection()
     {
         $adapter = new MysqlAdapter(MYSQL_DB_CONFIG);
-        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
+        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
     }
 }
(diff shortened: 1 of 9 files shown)


tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2214,4 +2214,16 @@ public function testGetPhinxTypeFromSQLDefinition(string $sqlDefinition, array $
         $this->assertSame($expectedResponse['name'], $result['name'], "Type mismatch - got '{$result['name']}' when expecting '{$expectedResponse['name']}'");
         $this->assertSame($expectedResponse['limit'], $result['limit'], "Field upper boundary mismatch - got '{$result['limit']}' when expecting '{$expectedResponse['limit']}'");
     }
+
+    public function testPdoPersistentConnection()
+    {
+        $adapter = new MysqlAdapter(MYSQL_DB_CONFIG + ['attr_persistent' => true]);
+        $this->assertEquals(true, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2221:
```suggestion
        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

tests/Phinx/Db/Adapter/PostgresAdapterTest.php
```diff
@@ -2310,4 +2310,16 @@ public function testInvalidPdoAttribute()
         $this->expectExceptionMessage('Invalid PDO attribute: attr_invalid (\PDO::ATTR_INVALID)');
         $adapter->connect();
     }
+
+    public function testPdoPersistentConnection()
+    {
+        $adapter = new PostgresAdapter(PGSQL_DB_CONFIG + ['attr_persistent' => true]);
+        $this->assertEquals(true, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2317:
```suggestion
        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -2306,4 +2306,16 @@ public function testPdoExceptionUpdateNonExistingTable()
         $table = new \Phinx\Db\Table('non_existing_table', [], $this->adapter);
         $table->addColumn('column', 'string')->update();
     }
+
+    public function testPdoPersistentConnection()
+    {
+        $adapter = new SQLiteAdapter(SQLITE_DB_CONFIG + ['attr_persistent' => true]);
+        $this->assertEquals(true, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2313:
```suggestion
        $this->assertTrue($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2214,4 +2214,16 @@ public function testGetPhinxTypeFromSQLDefinition(string $sqlDefinition, array $
...
+    }
+
+    public function testPdoNotPersistentConnection()
+    {
+        $adapter = new MysqlAdapter(MYSQL_DB_CONFIG);
+        $this->assertEquals(false, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2227:
```suggestion
        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

tests/Phinx/Db/Adapter/PostgresAdapterTest.php
```diff
@@ -2310,4 +2310,16 @@ public function testInvalidPdoAttribute()
...
+    }
+
+    public function testPdoNotPersistentConnection()
+    {
+        $adapter = new PostgresAdapter(PGSQL_DB_CONFIG);
+        $this->assertEquals(false, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2323:
```suggestion
        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -2306,4 +2306,16 @@ public function testPdoExceptionUpdateNonExistingTable()
...
+    }
+
+    public function testPdoNotPersistentConnection()
+    {
+        $adapter = new SQLiteAdapter(SQLITE_DB_CONFIG);
+        $this->assertEquals(false, $adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT));
```
MasterOdin, line 2319:
```suggestion
        $this->assertFalse($adapter->getConnection()->getAttribute(\PDO::ATTR_PERSISTENT);
```

MasterOdin (changes requested):

MasterOdin (changes requested):
Does `attr_persistent` work for SQL Server? If not, would there be any harm in passing it as an option anyway?

If either of the above is "yes", I think it would make sense to lift the code setting this option out of the specific adapters into the `PdoAdapter::createPdoConnection` method, having it be:

```php
        $adapterOptions = $this->getOptions() + [
            'attr_errmode' => PDO::ERRMODE_EXCEPTION,
        ];
        if (isset($adapterOptions['attr_persistent'])) {
          $driverOptions[PDO::ATTR_PERSISTENT] = $adapterOptions['attr_persistent'];
        }
```

This way we are following DRY principles.

paulermo:
> Does attr_persistent work for SQL Server? If not, would there be any harm in passing it as an option anyway?

No, SQL Server does not support persistent connections. And that's why I have moved the setting of driver attribute out from PDOAdapter to specific MySQL, Postgres and SQLite. For SQL Server nothing changes in comparison to current behavior.

> This way we are following DRY principles.

Yes, that was my first solution (https://github.com/cakephp/phinx/pull/2092/commits/7a49fa25bed3b4128d0afe16481043cd2048d852), but I have found that it could break SQL server functionality and moved it out in https://github.com/cakephp/phinx/pull/2092/commits/2b860620b4d22eda9d821d43778c731a0f25fc9c.

MasterOdin (approved):

MasterOdin:
Alright, I found https://github.com/Microsoft/msphpsql/issues/65 which confirms that setting that will in fact cause the driver to throw an exception, so we cannot set it for all drivers. I'm fine with just ignoring it for mssql as opposed to throwing an exception like `cakephp/database` does.

othercorey:
We have a small discussions on this in the phinx-dev channel. Leaving it up to @markstory still, but @MasterOdin had some thoughts after I explained the naming alignment.

MasterOdin:
I've opened #2097 to discuss the branch switch and can carry this conversation there.

PR #2094, opened 2022-05-19
Add TrustServerCertificate to SqlServerAdapter

### LINKED ISSUE #2093: How to use sqlsrv 2019?
Hi,

using docker with mssql image _mcr.microsoft.com/mssql/server:2019-latest_

how to set up phinx with _TrustServerCertificate=true_ ?

Getting this error when trying to migrate 
`PDOException: SQLSTATE[08001]: [Microsoft][ODBC Driver 18 for SQL Server]SSL Provider: [error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:self signed certificate] in /var/www/api/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php:84`

ingLomeland:
Fixes #2093

This PR adds _TrustServerCertificate=true_ to the connection string. This is needed for sqlserver 2017 and 2019. Tested that phinx works with sqlsrv 2019 after this fix.

@markstory

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -75,7 +75,7 @@ class SqlServerAdapter extends PdoAdapter
             if (!empty($options['port'])) {
                 $dsn .= ',' . $options['port'];
             }
-            $dsn .= ';database=' . $options['name'] . ';MultipleActiveResultSets=false';
+            $dsn .= ';database=' . $options['name'] . ';MultipleActiveResultSets=false;TrustServerCertificate=true';
 
             $driverOptions = [];
 
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -77,9 +77,12 @@ class SqlServerAdapter extends PdoAdapter
             }
             $dsn .= ';database=' . $options['name'] . ';MultipleActiveResultSets=false';
 
-            // option to turn on TrustServerCertificate
-            if (isset($options['trust_server_certificate'])) {
-                $dsn .= ';TrustServerCertificate=' . $options['trust_server_certificate'];
+            // option to add additional connection options
+            // https://docs.microsoft.com/en-us/sql/connect/php/connection-options?view=sql-server-ver15
+            if (isset($options['dsn_options'])) {
+                foreach ($options['dsn_options'] as $key => $option) {
+                    $dsn .= ';' . $key . '=' . $option;
+                }
             }
 
             $driverOptions = [];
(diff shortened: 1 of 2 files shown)


garas:
`TrustServerCertificate=true` skips the trust chain validation, so the application connects even if the certificate can't be verified. Using `TrustServerCertificate=false` enforces certificate validation and is a best practice.

It should be enabled through configuration. For this, it needs some new configuration key like `dsn_options`, might be useful for other adapters too.

ingLomeland:
Hi, added it as a configuration option. Could you look at it again? @garas

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -77,6 +77,11 @@ public function connect()
             }
             $dsn .= ';database=' . $options['name'] . ';MultipleActiveResultSets=false';
 
+            // option to turn on TrustServerCertificate
+            if (isset($options['trust_server_certificate'])) {
+                $dsn .= ';TrustServerCertificate=' . $options['trust_server_certificate'];
```
markstory, line 82: Should we handle converting the boolean into a string? Having users define a string true is a bit counter-intuitive.

MasterOdin: I wonder if there's a way to generalize this so that we don't need to add handlers for [all possible additional allowed options](https://docs.microsoft.com/en-us/sql/connect/php/connection-options?view=sql-server-ver15) for SQLServer. We use the prefix `sqlsrv_attr_` for detecting PDO settings, perhaps we can do a similar prefix for DSN specific settings, or have a key that expects to point at an associative array of settings? This way, we don't need to have this same if statement setup for a number of different options, and in this case, I think having them be whatever the user inputs makes sense (e.g. a string) as that's what the settings page calls for.

ingLomeland: Added dsn_options

MasterOdin (approved):

tests/Phinx/Db/Adapter/SqlServerAdapterTest.php
```diff
@@ -50,6 +50,14 @@ public function testConnection()
         $this->assertSame(\PDO::ERRMODE_EXCEPTION, $this->adapter->getConnection()->getAttribute(\PDO::ATTR_ERRMODE));
     }
 
+    public function testConnectionWithDsnOptions()
+    {
+        $options = $this->adapter->getOptions();
+        $options['dsn_options'] = ['TrustServerCertificate' => 'true'];
+        $this->adapter->setOptions($options);
+        $this->assertInstanceOf('PDO', $this->adapter->getConnection());
+    }
```
MasterOdin, line 59: This test isn't overly valuable as it doesn't really test that the `dsn_options` were used in any way, but there's currently not a good way to do that. This method is also inline with how we test the effect of [port] missing as well:

https://github.com/cakephp/phinx/blob/48d8e1f686852882a319e6deff1ae6748d04ab5c/tests/Phinx/Db/Adapter/SqlServerAdapterTest.php#L62-L68

so it's not like this is the only place we have this. I'd like to address this in a separate PR, so I'm fine with merging this for now as is.

PR #2096, opened 2022-07-04
Add php 8.2 to github actions and update versions

othercorey:


.github/workflows/ci.yml
```diff
@@ -1,20 +1,30 @@
...
-on: ['push', 'pull_request']
+on:
+  push:
+    branches:
+      - '0.x'
+      - '0.next'
+      - '1.x'
+  pull_request:
+    branches:
+      - '*'
```
MasterOdin, line 11: It should be noted that using `*` will not work if the branch name includes a `/`. If you're looking to match _any_ branch name, you'd want to use `**`. Alternatively, remove the entire clause and just do:

```yaml
  pull_request:
```
and I think that'll work the same way as doing:

```yaml
  pull_request:
    branches:
      - '**'
```

othercorey: Hmm, that was just copied from our cakephp repo. I'll check it out.

MasterOdin: Relevant GH action docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

> `*`: Matches zero or more characters, but does not match the `/` character. For example, `Octo*` matches `Octocat`.
> `**`: Matches zero or more of any character.

othercorey: I'm willing to bet that the original workflow for cake was copied from somewhere. I wish they'd document the defaults better. All I see is a comment explaining the default activity types that triggers, although it's logical that no branch filter means all branches.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

Looks like we need to clean this up for all repos.

dereuromark: yeah, remove it alltogether, just keep the pull_request: part.

MasterOdin: Yeah, when GH actions was original released, I remember a lot of the original examples and early repos would do `*`, as I remember hitting this issue in some of my early usages of GH actions. It's definitely not the most intuitive thing, especially as I think the vast majority of repos probably don't use `/` in branch names and so would never hit this case.

PR #2101, opened 2022-07-08
Do not set code when recasting PDOException

### LINKED ISSUE #2098: PHP 8.1 : Type issue when PDOException is thrown
PDOException can return an int|mixed. This code is directly passed to InvalidArgumentException which requires 'code' to be an int. So this will crash in php 8.1 en therefor not show the real error.

https://github.com/cakephp/phinx/blob/15be19734a596051120ff34a27e6ba39d65743d0/src/Phinx/Db/Adapter/PdoAdapter.php#L96

MasterOdin:
Fixes #2098 

The [`PDOException`](https://www.php.net/manual/en/class.pdoexception.php) class has the signature `int|string $code`, which makes it incompatible with most other stdlib exceptions that only allow `int $code`. While we could probably make a custom exception class or whatever for the recast, I don't think it's worth the hassle, and looking at the DB ecosystem, some amount of it also just drops the code in favor of just using `0` (e.g. `cakephp/Database` in [`Database/Driver`](https://github.com/cakephp/cakephp/blob/4aa99f805fb39b17ac7760e68fc1074bd826de0c/src/Database/Driver.php#L133) and have for years that I'm not sure people will really miss it.

The PDOException recast has been happening for the entire existence of this project that I'm somewhat risk adverse to remove the recast.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -96,7 +96,7 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
             throw new InvalidArgumentException(sprintf(
                 'There was a problem connecting to the database: %s',
                 $e->getMessage()
-            ), $e->getCode(), $e);
+            ), 0, $e);
         }
 
         return $db;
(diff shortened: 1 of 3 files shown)


MasterOdin:
Having this target `0.next` as it is a breaking change assuming there's someone out there relying on what the recast exception code is for doing something.

PR #2102, opened 2022-08-23
Pass environment to seed from manager

martenb:


src/Phinx/Migration/Manager.php
```diff
@@ -865,7 +865,7 @@ protected function orderSeedsByDependencies(array $seeds)
      * @throws \InvalidArgumentException
      * @return \Phinx\Seed\AbstractSeed[]
      */
-    public function getSeeds()
+    public function getSeeds(string $environment)
```
dereuromark: param docblock needs adjustment

dereuromark (approved):
Looks good to me

MasterOdin:
@martenb could you rebase this against the `0.next` branch instead of `0.x` as while I'm not sure how many people have created custom managers, this is still editing the public interface, and so is a BC break.

As part of rebasing, please also add full type hints to all the new methods and such you've added.

martenb:
@MasterOdin rebased

src/Phinx/Seed/AbstractSeed.php
```diff
@@ -61,6 +66,24 @@ public function getDependencies(): array
...
+    {
+        $this->environment = $environment;
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getEnvironment()
```
MasterOdin:
```suggestion
    public function getEnvironment(): string
```

src/Phinx/Seed/SeedInterface.php
```diff
@@ -42,6 +42,20 @@ public function run(): void;
      */
     public function getDependencies(): array;
 
+    /**
+     * Sets the environment.
+     *
+     * @return \Phinx\Seed\SeedInterface
```
MasterOdin:
```suggestion
     * @return $this
```

src/Phinx/Seed/SeedInterface.php
```diff
@@ -42,6 +42,20 @@ public function run(): void;
...
+     */
+    public function setEnvironment(string $environment);
+
+    /**
+     * Gets the environment.
+     *
+     * @return string
+     */
+    public function getEnvironment();
```
MasterOdin:
```suggestion
    public function getEnvironment(): string;
```

MasterOdin (changes requested):
some minor nits, and then I think good to merge :+1:

MasterOdin (approved):

PR #2106, opened 2022-10-02
Update for Cake 5

ADmad:


LordSimal:
I gotta say I am not really sure why the test for `$stm->rowCount()` was working with an expected value of `1`.
According to https://stackoverflow.com/questions/4694850/pdo-sqlite-query-zero-result-issue PDO Drivers are not really consistent when it comes to returning the correct row count for `SELECT` queries.
As you can see https://onlinephp.io/c/8a760 the RowCount is always 0 for a SQLite PDO object and a prepared select statement.

LordSimal:
Had to drop `symfony/console:^3.4` support because prefer-lowest installed a version which was not compatible with 8.1 named parameters

LordSimal:
And now `vendor/bin/validate-prefer-lowest` told me other required packages can be narrowed down as well...

Issue #2112, opened 2022-10-07
Migrations only running when specifying a target


malle-pietje:
I've started using phinx for a project where we already have an existing schema with basic seeding running `PHP 8.1.2`. 

I've managed to create the initial migration and another test migration which both work fine as long as I specify a target. Without specifying the target none of the migrations are executed.

Here's the output without specifying a target:
```sh
$ vendor/bin/phinx migrate -e production --dry-run
Phinx by CakePHP - https://phinx.org.

using config file phinx.php
using config parser php
using migration paths 
 - /XXXXXXXXXXXXXXXXXX/db/migrations
using seed paths 
 - /XXXXXXXXXXXXXXXXXX/db/seeds
using environment production
using adapter mysql
using database captive_portal_v2
ordering by creation time

All Done. Took 0.0105s
```

This is the output when specifying the target:
```sh
$ vendor/bin/phinx migrate -e production -t 20221007143139
Phinx by CakePHP - https://phinx.org.

using config file phinx.php
using config parser php
using migration paths 
 - /XXXXXXXXXXXXXXXXXX/db/migrations
using seed paths 
 - /XXXXXXXXXXXXXXXXXX/db/seeds
using environment production
using adapter mysql
using database captive_portal_v2
ordering by creation time

 == 20221007110332 InitialMigration: migrating 
 == 20221007110332 InitialMigration: migrated 1.8860s

 == 20221007143139 MyNewMigration: migrating 
 == 20221007143139 MyNewMigration: migrated 0.0134s

All Done. Took 1.9104s
```
Is there something I'm missing here? Any additional details needed?

malle-pietje:
Here's the configuration file I'm using:
```php
<?php

use Dotenv\Dotenv;

/**
 * load the .env file
 */
$dotenv = Dotenv::createImmutable(__DIR__);
$dotenv->load();

return
    [
        'paths'         => [
            'migrations' => '%%PHINX_CONFIG_DIR%%/db/migrations',
            'seeds'      => '%%PHINX_CONFIG_DIR%%/db/seeds',
        ],
        'environments'  => [
            'default_migration_table' => 'phinxlog',
            'default_environment'     => 'production',
            'production'              => [
                'adapter' => $_ENV['DB_CONNECTION'],
                'host'    => $_ENV['DB_HOST'],
                'name'    => $_ENV['DB_DATABASE'],
                'user'    => $_ENV['DB_USERNAME'],
[...]

PR #2115, opened 2022-10-08
do not cast version variable to int

nrob81:
fix $version casing issue discussed here: https://github.com/cakephp/phinx/issues/2112#issuecomment-1272335100

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -63,7 +63,7 @@ EOT
     {
         $this->bootstrap($input, $output);
 
-        $version = (int)$input->getOption('target');
+        $version = $input->getOption('target');
         /** @var string|null $environment */
         $environment = $input->getOption('environment');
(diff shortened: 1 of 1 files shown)


nrob81:
I only deleted the (int) cast of $version variable in migrate command.
Later, in Migration/Manager.php it is checked agains null.

nrob81:
i installed php7.3, tried this and worked. (with the int cast)

malle-pietje:
Good catch😉 Will check this tomorrow with my dev environment running PHP 8.1.2 and report back here.

nrob81:
👍 thank you

malle-pietje:
@nrob81 After removing the `(int)` cast from version 0.13.0 the migrations work fine here. Thanks for that quick fix!

nrob81:
Thanks for the quick merge :)

malle-pietje:
It was a quick and dirty test. I've rolled back again to 0.12.0 for now until the next release is out.

PR #2117, opened 2022-10-09
Fix deprecation warning on Config::offsetGet function

### LINKED ISSUE #2110: Deprecation warning in PHP 8.1 with Phinx 0.13.0
<b>Deprecated</b>:  Return type of Phinx\Config\Config::offsetGet($id) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in <b>/var/www/html/elternsprechtag/vendor/robmorgan/phinx/src/Phinx/Config/Config.php</b> on line <b>510</b><br>

MasterOdin:
Fixes #2110 

Adds the `#[\ReturnTypeWillChange]` annotation to the `Config::offsetGet` function as to properly type it requires using the `mixed` keyword added in 8.1 and we still support 7.2+

### DIFF
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -507,6 +507,7 @@ class Config implements ConfigInterface, NamespaceAwareInterface
      * @throws \InvalidArgumentException
      * @return mixed
      */
+    #[\ReturnTypeWillChange]
     public function offsetGet($id)
     {
         if (!array_key_exists($id, $this->values)) {


src/Phinx/Config/Config.php
```diff
@@ -533,6 +536,7 @@ public function offsetExists($id): bool
      * @param mixed $id ID
      * @return void
      */
+    #[\ReturnTypeWillChange]
     public function offsetUnset($id): void
```
othercorey: I don't think this one will change.

src/Phinx/Config/Config.php
```diff
@@ -522,6 +524,7 @@ public function offsetGet($id)
      * @param mixed $id ID
      * @return bool
      */
+    #[\ReturnTypeWillChange]
```
othercorey: I don't think this one will change.

src/Phinx/Config/Config.php
```diff
@@ -495,6 +495,7 @@ protected function parseAgnosticDsn(array $options): array
      * @param mixed $value Value
      * @return void
      */
+    #[\ReturnTypeWillChange]
     public function offsetSet($id, $value): void
```
othercorey: I don't think this one will change.

MasterOdin: Yeah, total brain fart on doing this, only `configGet` needs it 🤦

PR #2123, opened 2022-10-11
Consistently test console commands' exit codes

MasterOdin:
PR adds asserts for exit codes across all console command tests. Before, they were only consistently in place for the `MigrateTest` and `StatusTest`.

Somewhat unrelated, I also modified two tests in `InitTest` so that the use of `chdir` is better wrapped so that if something were to throw an exception in the test, we won't be potentially left in a different working directory for the rest of the suite / tests.


dereuromark:
Nice work

Issue #2130, opened 2022-10-26
Altering tables can cause foreign key constraint violations when the SQLite adapter implicitly drops tables


ndm2:
When using SQLite, altering a table will cause an implicit table drop for various operations like adding or removing columns:

https://github.com/cakephp/phinx/blob/8891e74acec13ae57c5bc7346422a9c9781fe6cf/src/Phinx/Db/Adapter/SQLiteAdapter.php#L825-L849

Naturally this can cause a foreign key constraint violation in case data is present accordingly.

Example migration that reproduces the problem:

```php
$this
    ->table('articles')
    ->insert(['id' => 1])
    ->create();

$this
    ->table('comments')
    ->addColumn('article_id', 'integer')
    ->addForeignKey('article_id', 'articles', 'id', [
        'update' => ForeignKey::RESTRICT,
        'delete' => ForeignKey::RESTRICT
    ])
    ->insert(['id' => 1, 'article_id' => 1])
    ->create();

$this
    ->table('articles')
    ->addColumn('new_column', 'integer')
    ->update();
```

Adding the new column will cause a constraint violation, as it will drop the `articles` table at some point:

```
PDOException : SQLSTATE[23000]: Integrity constraint violation: 19 FOREIGN KEY constraint failed
 src/Phinx/Db/Adapter/PdoAdapter.php:194
```

PR #2131, opened 2022-10-26
Fix foreign key constraint violations on implicit table drops.

ndm2:
refs #2130

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -449,7 +449,7 @@ class SQLiteAdapter extends PdoAdapter
         if (!empty($primaryKey)) {
             $instructions->merge(
                 // FIXME: array access is a hack to make this incomplete implementation work with a correct getPrimaryKey implementation
-                $this->getDropPrimaryKeyInstructions($table, $primaryKey[0])
+                $this->getDropPrimaryKeyInstructions($table, $primaryKey[0], false)
             );
         }
 
@@ -820,11 +820,16 @@ PCRE_PATTERN;
      *
      * @param \Phinx\Db\Util\AlterInstructions $instructions The instructions to modify
      * @param string $tableName The table name to copy the data to
+     * @param bool $validateForeignKeys Whether to validate foreign keys after the copy and drop operations. Note that
+     *  enabling this option only has an effect when the `foreign_keys` PRAGMA is set to `ON`!
      * @return \Phinx\Db\Util\AlterInstructions
      */
-    protected function copyAndDropTmpTable(AlterInstructions $instructions, string $tableName): AlterInstructions
-    {
-        $instructions->addPostStep(function ($state) use ($tableName) {
+    protected function copyAndDropTmpTable(
+        AlterInstructions $instructions,
+        string $tableName,
+        bool $validateForeignKeys = true
+    ): AlterInstructions {
+        $instructions->addPostStep(function ($state) use ($tableName, $validateForeignKeys) {
             $this->copyDataToNewTable(
                 $state['tmpTableName'],
                 $tableName,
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -846,7 +846,15 @@ protected function copyAndDropTmpTable(AlterInstructions $instructions, string $
                 )
             );
 
+            $foreignKeysEnabled = (bool)$this->fetchRow('PRAGMA foreign_keys')['foreign_keys'];
+            if ($foreignKeysEnabled) {
+                $this->execute('PRAGMA foreign_keys = OFF');
+            }
             $this->execute(sprintf('DROP TABLE %s', $this->quoteTableName($tableName)));
+            if ($foreignKeysEnabled) {
+                $this->execute('PRAGMA foreign_keys = ON');
```
MasterOdin, line 860: Does toggling back on foreign key pragma also revalidate any existing FKs, or do we need to also run `PRAGMA foreign_key_check;`?

ndm2: AFAICT it doesn't (and apparently neither will renaming the temp table). But is a revalidation at that point really neccessary? Do we expect `copyDataToNewTable()` to possibly silently fail (in whatever way)?

MasterOdin: I'd say it wouldn't hurt to do as an extra precaution though I cannot think of a case off-hand that might cause that to error.

ndm2: I'll give it a try.

ndm2: I gave it a try, and it seems to work OK, but it appears that it would introduce a breaking change for at least `changePrimaryKey()`.

Currently when the primary key is part of a foreign key constraint, and the primary key is being changed using `changePrimaryKey()`, things will just continue to work, the PK changes, and the FK constraint stays as-is. However this can actually leave the FK constraint in an invalid state, as its parent key columns must be either primary keys, or subject to a unique constraint.

Since PKs are already unique constraints in and of themselves, one would usually not put an additional unique index on top, which means that a PK change could then leave the old column without any unique constraint, and then `PRAGMA foreign_key_check()` would cause a "foreign key mismatch" SQL error.

I guess that behavior could actually be desired, as one would usually like to know when making changes that leaves the DB in an invalid state, but would that actually be in accordance with the backwards compatibility policy of this project?

ndm2: Pushed the changes so that you have something to look at.

I'm not overly happy with the need for the option whether to validate foreign keys, but there needs to be some form of control over it when running multi-step operations like `getChangePrimaryKeyInstructions()`, otherwise, in this specific case, a constraint error would be triggered when validation occurs in between removing the old PK and adding the new one.

ndm2: @MasterOdin Do you like what you see, or nah?

@dereuromark In light of [the FK problems](https://github.com/cakephp/phinx/issues/2154) you've brought up, do you possibly have any input on the BC break here? Both of this currently works, and would trigger an error after this change in case there isn't already a unique index on the current PK:

```php
 $table
    ->changePrimaryKey('other');
```
```php
 $table
    ->changePrimaryKey('other')
    ->addIndex('current', ['unique' => true]);
```

Adding the index first will continue to work:

```php
 $table
    ->addIndex('current', ['unique' => true])
    ->changePrimaryKey('other');
```

dereuromark:
So what shall we do with this one now?

src/Phinx/Db/Adapter/SQLiteAdapter.php
dereuromark: Is this still an issue given the feature flags being added now?

ndm2: I guess it depends on whether the foreign key check is something that should be hidden behind a feature flag? And on whether this breaking change is actually considered a problem.

MasterOdin: I feel like this is a bugfix and that the prior behavior was incorrect as it could leave the DB in a non-valid state, and so I'm not sure I'd consider this a "breaking change". As such, I don't think we need to wrap this in a feature flag.

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -846,7 +851,15 @@ protected function copyAndDropTmpTable(AlterInstructions $instructions, string $
...
```
MasterOdin, line 860: On a separate note, do we want to enable the foreign key constraint here, or could we move it down to where we validate the FKs, doing:

```php
            if ($foreignKeysEnabled) {
                $this->execute('PRAGMA foreign_keys = ON');
                if ($validateForeignKeys) {
                    $this->validateForeignKeys($tableName);
                }
            }
```

As I think you've noted the `pragma` has no effect on its own for the various other operations done here, which is covered by doing the `foreign_key_check`.

ndm2: I would probably keep it here for now. For #2134 there will be some changes needed to be done, including moving the foreign key check, it will look like this:

https://github.com/ndm2/phinx/blob/0.x-migrations-cake5-test/src/Phinx/Db/Adapter/SQLiteAdapter.php#L1181

I guess the switching could possibly be done in `endAlterByCopyTable()` then too. Maybe we should revisit that in the other PR once it has been rebased/changed accordingly?

MasterOdin: Sure, that's fine.

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1371,6 +1372,208 @@ public function testAlterTableWithConstraints()
...
+
+    /**
+     * Tests that foreign key constraint violations introduced around the table
+     * alteration process (being it implicitly by the process itself or by the user)
+     * will trigger an error accordingly.
+     */
+    public function testAlterTableDoesViolateForeignKeyConstraintOnTargetTableChange()
+    {
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('Integrity constraint violation: FOREIGN KEY constraint on `comments` failed.');
```
MasterOdin: Can we put this right above the line that'll trigger it? That way, we won't risk some other line unexpectedly throwing this error, and miss the regression.

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1371,6 +1372,208 @@ public function testAlterTableWithConstraints()
...
+
+    /**
+     * Tests that foreign key constraint violations introduced around the table
+     * alteration process (being it implicitly by the process itself or by the user)
+     * will trigger an error accordingly.
+     */
+    public function testAlterTableDoesViolateForeignKeyConstraintOnSourceTableChange()
+    {
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('Integrity constraint violation: FOREIGN KEY constraint on `comments` failed.');
```
MasterOdin: Same here.

MasterOdin (changes requested):

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
ndm2: Sure thing, its done.

dereuromark:
@MasterOdin Feel free to merge if it is all right now.

MasterOdin (approved):

PR #2133, opened 2022-10-31
Fix deprecated ${} string interpolation

rigrig:
`"${var}"` is deprecated in PHP 8.2, and should be replaced with `"{$var}"`:
https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation

I searched for `${`, as far as I can tell this line was the only place where this occurred.


MasterOdin (approved):

dereuromark:
https://github.com/cakephp/phinx/actions/runs/3314237268/jobs/5473259638
There might be more invisible issues

Issue #2129, opened 2022-10-25
Renaming/Removing columns clashes with index recreation


ndm2:
As hinted in #2126, (explicitly) renaming/removing columns clashes with the index recreation fix for SQLite added in #2040, as it will try to recreate indices on columns that do not exist anymore, and trigger errors accordingly.

The behavior of the other supported DBMS isn't consistent:

### Renaming a column:

* **MySQL**: Renames the column in the index
* **Postgres**: Renames the column in the index
* **SQL Server**: Renames the column in the index (and triggers a warning)

### Removing a column:

* **MySQL**: Removes the complete index
* **Postgres**: Removes the complete index
* **SQL Server**: Triggers an error

### Renaming a column:

* **MySQL**: Renames the column in the index
* **Postgres**: Renames the column in the index
* **SQL Server**: Renames the column in the index (and triggers a warning)

### Removing a column:

* **MySQL**: Removes the column from the index
* **Postgres**: Removes the complete index
* **SQL Server**: Triggers an error

### Renaming a column:

* **MySQL**: Triggers an error
* **Postgres**: Renames the column in the index

### Removing a column:

* **MySQL**: Triggers an error
* **Postgres**: Removes the complete index

---

# How should the SQLite adapter behave?

I would suggest that indices with expressions should trigger an error, IMHO it's not feasible to go full blown SQL parsing in order to support them. And since they can only be detected as belonging to a specific column when that column is not in an expression, I would say that any index with an expression should simply be ignored alltogether, so that they will cause an SQL error when trying to recreate them. People that are using such custom indices will simply have to deal with removing/adding them themselves, just like they'd currently have to do it with MySQL.

Renaming columns should simply rename the columns in the index, that should be easy enough. As noted in #2126, it seems that the rule for indices is that almost everything after `CREATE [UNIQUE] INDEX` is basically (syntactically correct) garbage in, garbage out, so a little regex foo might already do the job. For example:

```sql
CREate INdex test_idx On test (foo asc , aBs(bar)   DEsc )
```

is stored as-is, except for `CREate INdex`, meaning `sqlite_master` will contain/return:

```sql
CREATE INDEX test_idx On test (foo asc , aBs(bar)   DEsc )
```

And removing columns, I'm not sure, personally I'd be fine with either the behavior of MySQL or Postgres, ie either remove individual columns, or always remove the complete index, even if only one of multiple columns is being removed.

PR #2134, opened 2022-11-01
Fix recreating indices for renamed/removed columns.

ndm2:
A possible implementation to solve #2129, it ignores indices with expressions (which results in SQL errors accordingly when columns are missing on index recreation), and behaves like Postgres on column removal, ie it always removes the index, even composite ones.

Short of actually parsing the SQL, that regex for renaming is the best I could come up with so far. It can handle quite some variations, but I'm certainly not claiming that there aren't possibly cases where it would fail.

### DIFF
--- a/src/Phinx/Db/Adapter/SQLiteAdapter.php
+++ b/src/Phinx/Db/Adapter/SQLiteAdapter.php
@@ -449,7 +449,7 @@ class SQLiteAdapter extends PdoAdapter
         if (!empty($primaryKey)) {
             $instructions->merge(
                 // FIXME: array access is a hack to make this incomplete implementation work with a correct getPrimaryKey implementation
-                $this->getDropPrimaryKeyInstructions($table, $primaryKey[0], false)
+                $this->getDropPrimaryKeyInstructions($table, $primaryKey[0])
             );
         }
 
@@ -734,7 +734,7 @@ PCRE_PATTERN;
             return $newState + $state;
         });
 
-        return $this->copyAndDropTmpTable($instructions, $tableName);
+        return $this->endAlterByCopyTable($instructions, $tableName);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -969,7 +1127,12 @@ protected function getRenameColumnInstructions(string $tableName, string $column
             return $newState + $state;
         });
 
-        return $this->copyAndDropTmpTable($instructions, $tableName);
+        $instructions = $this->bufferIndicesAndTriggers($instructions, $tableName);
+        $instructions = $this->updateIndicesForRenamedColumn($instructions, $columnName, $newColumnName);
+        $instructions = $this->copyAndDropTmpTable($instructions, $tableName);
```
MasterOdin: Thoughts on modifying `copyAndDropTmpTable` to accept an optional third and fourth arguments `$columnName` and `newColumnName`, and then `copyAndDropTmpTable` is responsible for calling `bufferIndicesAndTriggers`, `updateIndicesForRenamedColumn`, and `recreateIndicesAndTriggers`?

This would cut down on the boilerplate around using these as three of them are always used in conjunction.

MasterOdin (changes requested):

src/Phinx/Db/Adapter/SQLiteAdapter.php
ndm2: I'm very much open to some refactoring here, I don't really like all that repitition either. I actually had it like this at first, with the new methods added around the copy/delete post step:

```
copyAndDropTmpTable(AlterInstructions $instructions, string $tableName, string $deletedOrRenamedColumnName, string $newColumnName)
```

It was kinda bloated, and I thought that it wouldn't be overly appreciated.

How about a new method that combines all the post step addition methods, something like `endAlterByCopyTable` to complement `beginAlterByCopyTable`?

MasterOdin: Sure, I'd be open to having a complimentary `endAlterByCopyTable` that wraps these 4 function calls as-is, makes sense as we have a `beginAlterByCopyTable`.

dereuromark:
@MasterOdin Can we move forward here? You still had it on request change

ndm2:
As a heads up, ideally #2131 is merged first, it will make rebasing less awkward, as there needs to be some structural changes here in case the other PRs foreign key check behavior is being kept.

dereuromark:
ping @MasterOdin

ndm2:
Just realized I've linked the wrong PR, I didn't mean to say that this one should to be merged first, but the table drop one, sorry 😬

PR #2140, opened 2022-11-16
Fix enum

### LINKED ISSUE #2139: Setting a value for a column fails when we try to pass a string as values
PHP 7.3
Invalid type cast. When the value is a string.
Example:
```
...
->addColumn('orderExample', 'enum', ['values' => 'first, second', 'default' => 'first'])
...
```
`$values = 'first, second';`
https://github.com/cakephp/phinx/blob/50585ced02cd53a7fc326a728d00c38c5cee43f5/src/Phinx/Db/Table/Column.php#L632
The result is the string `'Array'`.

Thanks!

dereuromark:
Resolves https://github.com/cakephp/phinx/issues/2139

### DIFF
--- a/src/Phinx/Db/Table/Column.php
+++ b/src/Phinx/Db/Table/Column.php
@@ -629,7 +629,7 @@ class Column
     public function setValues($values)
     {
         if (!is_array($values)) {
-            $values = (string)preg_split('/,\s*/', $values);
+            $values = preg_split('/,\s*/', $values) ?: [];
         }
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):
Think it's worth adding some simple unit tests around this function?

dereuromark:
True, can definitly be a follow up
Anyone up for it?

AlekseiMaryasov:
@dereuromark  It solved the problem, nice work

Issue #2145, opened 2022-11-23
Custom Seed Template from Config


SunriseCoder:
Hello!

We are using Phinx for database migration and default template fits not so good in our corporate standards.
I'm aware of the following issue: [seed template #1507](https://github.com/cakephp/phinx/issues/1507), but the way of passing custom seed template as console argument is not so convenient. I would even call it redundant, because we have a single template for the whole project and so on.

I would like to try to add possibility to change the Seed Template File through Config.

I suggest to make the following priority of using the custom template:
1. Passed as a console option
2. If console option is not set, then attempt to take it from the Config. If set - then using console option
3. If nothing from Config either - then using Hardcoded Default Seed Template

One potential problem could be if I change [src/Phinx/Config/ConfigInterface.php](https://github.com/cakephp/phinx/blob/0.x/src/Phinx/Config/ConfigInterface.php) by adding a new method and change one class in the repository which implements it, project should not be broken, but the customers, if they for some reasons implemented that interface, their software could be broken until they implement it. To be honest, I don't see any reasons to implement this interface itself instead of Config-class.

The project is not designed to use custom Config-classes, so I see the only way to add this feature is to change ConfigInterface and Config, so just for circumstances I started this issue before I started the working on it, so please, give me short feedback about it while I'm implementing in order to increase the speed of accepting my Pull Request, which I will add as soon as I finish the changes.

Thank You very much for this nice project.

SunriseCoder:
Implementation is done, just opened the pull request: https://github.com/cakephp/phinx/pull/2146

SunriseCoder:
Ok, I don't understand how the review mechanism is working, so I decided to reopen this issue, maybe it will help somehow...
This is the pull-request: https://github.com/cakephp/phinx/pull/2146, which has been reviewed by @dereuromark, I made all changes as he suggested, but now I don't know what to do, I asked him to review again, but my re-review request is still pending...

Maybe someone else could review and accept it, please?

PR #2146, opened 2022-11-23
Added config option to specify custom seeder template path to use

SunriseCoder:
Hello!
Just finished work for the following issue: [Custom Seed Template from Config #2145](https://github.com/cakephp/phinx/issues/2145)

Could anyone please make code review and accept or give me feedback to fix it?
Thank You very very much :)

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -151,7 +151,7 @@ setting ``seed_base_class`` in your config:
     seed_base_class: MyMagicalSeeder
 
 Custom Migration Template
-----------------------
+-------------------------
 
 Custom template for Migrations could be used either by defining template file path
 in configuration file:
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -150,6 +150,30 @@ setting ``seed_base_class`` in your config:
 
     seed_base_class: MyMagicalSeeder
 
+Custom Migration Template
+----------------------
+
+Custom template for Migrations could be used either by defining template file path
+in configuration file:
+
+.. code-block:: yaml
+
+    templates:
+        file: src/templates/customMigrationTemplate.php.dist
+
+
+Custom Seeder Template
+----------------------
+
+Custom Seeder template could be used either by defining template file path
+in configuration file:
+
+.. code-block:: yaml
+
+    templates:
+        seedTemplateFile: src/templates/customSeederTemplate.php.dist
+
+
 Environments
 ------------
 
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -159,7 +159,7 @@ in configuration file:
 .. code-block:: yaml
 
     templates:
-        file: src/templates/customMigrationTemplate.php.dist
+        file: src/templates/customMigrationTemplate.php
 
 
 Custom Seeder Template
@@ -171,7 +171,7 @@ in configuration file:
 .. code-block:: yaml
 
     templates:
-        seedTemplateFile: src/templates/customSeederTemplate.php.dist
+        seedTemplateFile: src/templates/customSeederTemplate.php
 
 
 Environments
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -171,7 +171,7 @@ in configuration file:
 .. code-block:: yaml
 
     templates:
-        seedTemplateFile: src/templates/customSeederTemplate.php
+        seedFile: src/templates/customSeederTemplate.php
 
 
 Environments
--- a/src/Phinx/Config/Config.php
+++ b/src/Phinx/Config/Config.php
@@ -542,8 +542,8 @@ class Config implements ConfigInterface, NamespaceAwareInterface
     /**
      * @inheritdoc
      */
-    public function getSeederTemplateFile(): ?string
+    public function getSeedTemplateFile(): ?string
     {
-        return $this->values['templates']['seedTemplateFile'] ?? null;
+        return $this->values['templates']['seedFile'] ?? null;
     }
 }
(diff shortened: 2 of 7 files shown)


docs/en/configuration.rst
```diff
@@ -150,6 +150,30 @@ setting ``seed_base_class`` in your config:
 
     seed_base_class: MyMagicalSeeder
 
+Custom Migration Template
+----------------------
```
dereuromark: Length must be equal to header text

src/Phinx/Config/Config.php
```diff
@@ -538,4 +538,12 @@ public function offsetUnset($id): void
     {
         unset($this->values[$id]);
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function getSeederTemplateFile(): ?string
+    {
+        return empty($this->values['templates']) ? null : $this->values['templates']['seedTemplateFile'] ?? null;
```
dereuromark:
```suggestion
        return $this->values['templates']['seedTemplateFile'] ?? null;
```
should suffice, no?

src/Phinx/Console/Command/SeedCreate.php
```diff
@@ -178,11 +178,22 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             ));
         }
 
+        // Command-line option must have higher priority than value from Config
+        $config = $this->getConfig();
+        if (is_null($altTemplate)) {
+            $altTemplate = $config->getSeederTemplateFile();
+            if (!is_null($altTemplate) && !is_file($altTemplate)) {
+                throw new InvalidArgumentException(sprintf(
+                    'The template file "%s" from config does not exist',
```
dereuromark: ` is used for code fencing, not "

docs/en/configuration.rst
```diff
@@ -150,6 +150,30 @@ setting ``seed_base_class`` in your config:
...
+Custom Migration Template
+-------------------------
+
+Custom template for Migrations could be used either by defining template file path
+in configuration file:
+
+.. code-block:: yaml
+
+    templates:
+        file: src/templates/customMigrationTemplate.php.dist
```
dereuromark: why do we need to call them .dist?
usually thats only needed if those end up in main repo as such and have to be copied manually to get activated.

src/Phinx/Config/Config.php
SunriseCoder: > Tried, got problem during implementation if "templates" array is empty, therefore must to use this ugly way... but now I restarted tests, everything works fine, so thank You very much :) Now I know that this works with 2D arrays even :)

src/Phinx/Console/Command/SeedCreate.php
SunriseCoder: > ` is used for code fencing, not "

Oh, sorry, I copy-pasted it from line 176 by @Firtzberg from 2021-01-13, and I don't even know what does that fencing means... I will read about it for sure... but for this code should I replace double quotes with ` in the whole method or just in my code?

@dereuromark Could You answer, please? Just the last problem to fix...

docs/en/configuration.rst
SunriseCoder: > why do we need to call them .dist? usually thats only needed if those end up in main repo as such and have to be copied manually to get activated.

Yes, You're right, just found this file in the code repository and decided to name it this way. I thought that this is some kind of template file extension, because I'm not so much experienced with PHP yet... I will fix it in a moment...

src/Phinx/Console/Command/SeedCreate.php
SunriseCoder: Ok, replaced " with ` here and in the test

SunriseCoder:
@dereuromark Hello Mark! How are You? Could You please accept the pull-request? :)

dereuromark:
I already looked at it, would be nice if someone else also looks at it

src/Phinx/Config/Config.php
```diff
@@ -538,4 +538,12 @@ public function offsetUnset($id): void
     {
         unset($this->values[$id]);
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function getSeederTemplateFile(): ?string
+    {
+        return $this->values['templates']['seedTemplateFile'] ?? null;
```
othercorey: This config name should probably match the getter function name.

SunriseCoder: > This config name should probably match the getter function name.

@othercorey Good that You raised this question - I was not sure, is there any kind of naming conventions in the project, how to name this in the config and how to name the getter...

Do You have any suggestions on the path of this option in the config or it's fine and I should just rename the getter? "getSeedTemplateFile()" is ok, or should I add "templates" to the getter name somehow?

othercorey: I suggest making the config `['templates']['seed']` or `['templates']['seedFile']` and the getter `getSeedTemplate()` or `getSeedTemplateFile()`.

Depends on whether other templates have the `File` suffix.

SunriseCoder: > I suggest making the config `['templates']['seed']` or `['templates']['seedFile']` and the getter `getSeedTemplate()` or `getSeedTemplateFile()`.
> 
> Depends on whether other templates have the `File` suffix.

@othercorey Ok, I changed to `['templates']['seedFile']` and `getSeedTemplateFile()`

SunriseCoder:
By the way, I'm not sure how github and the setup of this project works, but I made a lot of commits like "fix review problems", so is it possible to sqaush it in a single commit to get rid of this junk? Will it be done automatically my accepting the pull-request or should I do something?

dereuromark:
No worries, squash merge is fixing it

othercorey (approved):

SunriseCoder:
@dereuromark @othercorey 
Thank You very much, that was my first open-source contribution, maybe not so big, but thank You very much for Your help :)

PR #2158, opened 2022-12-20
add fix for mariadb virtual column creation

### LINKED ISSUE #2157: Using Literal for creating virtual columns throws error on MariaDB
When using Literal class to create a virtual column in MariaDB like so:
``` 
$this->table['imaginary_table']->addColumn('imaginary_column', Literal::from("varchar(30) COLLATE utf8mb4_general_ci GENERATED ALWAYS AS (JSON_UNQUOTE(JSON_EXTRACT(`json_column`, '$.json_value'))) VIRTUAL"));
```

it throws MySQL error 
```
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'NOT NULL
```
as `NOT NULL` is not allowed after creation of virtual column. `NOT NULL` is by default added on https://github.com/cakephp/phinx/blob/343645c64bcba1327f706dd6a3de5bca8bdb2ae6/src/Phinx/Db/Adapter/MysqlAdapter.php#L1365
Virtual column is set as `NULL` in MariaDB as default and cannot be changed so also if create statement is rewritten to:
``` 
$this->table['imaginary_table']->addColumn('imaginary_column', Literal::from("varchar(30) COLLATE utf8mb4_general_ci GENERATED ALWAYS AS (JSON_UNQUOTE(JSON_EXTRACT(`json_column`, '$.json_value'))) VIRTUAL", [
    'null' => true
]));
```
note `['null' => true]` as the option, it returns an error of
```
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'NULL
```

I suggest rewriting: https://github.com/cakephp/phinx/blob/343645c64bcba1327f706dd6a3de5bca8bdb2ae6/src/Phinx/Db/Adapter/MysqlAdapter.php#L1365
to:
```
if (!($column->getType() instanceof Literal) || strpos($this->getConnection()->getAttribute(PDO::ATTR_SERVER_VERSION), "MariaDB") === false) {
    $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
}
```

Tested on MariaDB and MySQL and works as expected.

Any thoughts?

Regards, Oskar

oskarkregar:
When using virtual column definition with `Literal` class, migration fails with SQL syntax error on MariaDB as `NOT NULL` or `NULL` are not allowed at the end of column definition.  Below is the link to an issue.

Fixes https://github.com/cakephp/phinx/issues/2157

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1363,7 +1363,7 @@ class MysqlAdapter extends PdoAdapter
         $def .= $column->getCollation() ? ' COLLATE ' . $column->getCollation() : '';
         $def .= !$column->isSigned() && isset($this->signedColumnTypes[$column->getType()]) ? ' unsigned' : '';
 
-        if (!($column->getType() instanceof Literal) || strpos($this->getConnection()->getAttribute(PDO::ATTR_SERVER_VERSION), "MariaDB") === false) {
+        if (!($column->getType() instanceof Literal) || strpos($this->getConnection()->getAttribute(PDO::ATTR_SERVER_VERSION), 'MariaDB') === false) {
             $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
         }
 
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1362,7 +1362,10 @@ class MysqlAdapter extends PdoAdapter
         $def .= $column->getEncoding() ? ' CHARACTER SET ' . $column->getEncoding() : '';
         $def .= $column->getCollation() ? ' COLLATE ' . $column->getCollation() : '';
         $def .= !$column->isSigned() && isset($this->signedColumnTypes[$column->getType()]) ? ' unsigned' : '';
-        $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
+
+        if (!($column->getType() instanceof Literal) || strpos($this->getConnection()->getAttribute(PDO::ATTR_SERVER_VERSION), "MariaDB") === false) {
+            $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
+        }
 
         if (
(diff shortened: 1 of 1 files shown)


oskarkregar:
Fixes #2157

ndm2:
I'm not sure if it's overly reasonable to completely break the `null` option for all literal type usages with MariaDB, just because of [this one quirks](https://jira.mariadb.org/browse/MDEV-10964).

Maybe the check should rather look for SQL that is incompatible with `[NOT] NULL`, or maybe it should be possible to explicitly suppress generating `[NOT] NULL`?

In any case, this should have tests. Since existing tests didn't break, there seems to already be missing coverage.

MasterOdin:
We don't currently test against mariadb, so we'd need to wire that up before we could have a good test for this. MySQL does happily allow `null` and `not null` for virtual columns so I wouldn't expect our existing tests to be able to cover this.

oskarkregar:
It could then be fixed with option to explicitly suppress `null` and `not null` as @ndm2 suggested. It is not needed on most MySQL and MariaDB column creation (except on `TIMESTAMP` type).

kyojin007:
Is there an argument to say that adding options to a Literal column type is unnecessary? By definition we are manually specifying the column and allowing options just adds complexity.

Then we might do;

```
        if (!$column->getType() instanceof Literal) {
            $def .= $column->getEncoding() ? ' CHARACTER SET ' . $column->getEncoding() : '';
            $def .= $column->getCollation() ? ' COLLATE ' . $column->getCollation() : '';
            $def .= !$column->isSigned() && isset($this->signedColumnTypes[$column->getType()]) ? ' unsigned' : '';
            $def .= $column->isNull() ? ' NULL' : ' NOT NULL';
        }
```

dereuromark:
@ndm2 So you think we shouldn't merge this?

ndm2:
@dereuromark Not in this state, no. IMHO there should either be an option to exclude `[not] null` generation, or there should be a clear way to define the whole column creation SQL via a `Literal` object. I would also second @MasterOdin that this should have an actual test on MariaDB.

Issue #2154, opened 2022-12-12
BC break mitigation strategy


dereuromark:
https://github.com/cakephp/phinx/pull/1899
Since Phinx 0.13, foreign key columns have to be specified with `"signed" => false`

Also refs https://github.com/cakephp/migrations/issues/599

That causes now BC breaks in all kind of applications due to plugins not containing this

> PDOException: SQLSTATE[HY000]: General error: 1005 Can't create table `release_local`.`state_machine_item_states` (errno: 150 "Foreign key constraint is incorrectly formed") in /shared/httpd/release/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php:192

due to https://github.com/spryker/cakephp-statemachine/blob/81b6ca3c9fddf5dbacc8d863f4096f77a9d6c4c5/config/Migrations/20161029230233_StateMachineInit.php

Why was this a requirement again? Why couldnt we continue to treat also unsigned ones as valid foreign key columns and just use signed on by default then?
I would only expect the generation of new migration files to be "more specific", while keeping support for every file so far.

https://github.com/spryker/cakephp-statemachine/commit/953489cd55fec731dcccda59567ecfcd1fc2f16b seems to be needed, instead, maybe we could offer some migration to provide this where needed instead of just failing hard?
Not sure what could be done here.

markstory:
I also ran into my migrations being 'broken' by this change. Tracking down where each foreign key was added to the schema and updating those migrations was pretty tedious.

othercorey:
Seems that question was asked in the PR but was merged anyway.

https://github.com/cakephp/phinx/pull/1899

dereuromark:
Yeah, there was no answer given at the time. I guess we need to revisit this, or provide a mitigation strategy as outlined above.

othercorey:
Do you want to revert this change for now? It doesn't affect me, but does seem like a change we should make now if we're going to.

dereuromark:
It would be good to discuss mitigation strategies first, as this mainly affects us in terms of cake migration users right now.
But if there is no other Option we can revert for sure

MasterOdin:
I don't think this should be reverted. In my opinion, Phinx (anad other DB migration libraries) should ideally encourage the best practice usage of a database, and having an unsigned primary key int is that, and is the recommendation in the official docs about auto-increment columns.

[...]

PR #2159, opened 2022-12-22
Add feature flags to disable some breaking changes

MasterOdin:
PR to address comments in #2154 

A new `FeatureFlags` class has been introduced that we can use to wrap larger breaking changes, such that end-users may optionally disable them.

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -518,11 +518,12 @@ Feature Flags
 
 For some breaking changes, Phinx offers a way to opt-out of new behavior. The following flags are available:
 
-* ``unsigned_pks``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
-* ``column_null``: Should Phinx create columns as null by default? (default: ``true``)
+* ``unsigned_primary_keys``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
+* ``column_null_default``: Should Phinx create columns as null by default? (default: ``true``)
 
-These values can also be set by modifying class fields on the ```Phinx\Config\FeatureFlags``` class, for example:
+These values can also be set by modifying class fields on the ```Phinx\Config\FeatureFlags``` class, converting
+the flag name to ``camelCase``, for example:
 
 .. code-block:: php
 
-    Phinx\Config\FeatureFlags::$unsignedPks = false;
+    Phinx\Config\FeatureFlags::$unsignedPrimaryKeys = false;
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -512,3 +512,17 @@ Within the bootstrap script, the following variables will be available:
      * @var \Symfony\Component\Console\Output\OutputInterface $output The executing command's output object
      * @var \Phinx\Console\Command\AbstractCommand $context the executing command object
      */
+
+Feature Flags
+-------------
+
+For some breaking changes, Phinx offers a way to opt-out of new behavior. The following flags are available:
+
+* ``unsigned_pks``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
+* ``column_null``: Should Phinx create columns as null by default? (default: ``true``)
+
+These values can also be set by modifying class fields on the ```Phinx\Config\FeatureFlags``` class, for example:
+
+.. code-block:: php
+
+    Phinx\Config\FeatureFlags::$unsignedPks = false;
--- a/src/Phinx/Config/FeatureFlags.php
+++ b/src/Phinx/Config/FeatureFlags.php
@@ -22,7 +22,13 @@ class FeatureFlags {
      */
     public static $columnNull = true;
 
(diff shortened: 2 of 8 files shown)


MasterOdin:
I still need to write some tests for this, but wanted to get the ball rolling with seeing what people thought of this approach.

docs/en/configuration.rst
```diff
@@ -512,3 +512,17 @@ Within the bootstrap script, the following variables will be available:
...
      * @var \Phinx\Console\Command\AbstractCommand $context the executing command object
      */
+
+Feature Flags
+-------------
+
+For some breaking changes, Phinx offers a way to opt-out of new behavior. The following flags are available:
+
+* ``unsigned_pks``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
+* ``column_null``: Should Phinx create columns as null by default? (default: ``true``)
```
dereuromark: Are we usually using `snake_case` or `camelBacked` style for config options?

dereuromark:
So the plan would be to release this as 0.13.4
Then make Migrations current master release a patch that has this version as minimum with the feature flags disabled for now.

We will then add docs on how you can already switch those on if you want before they become default in Cake5 Migrations plugin major.

:+1: Should work

docs/en/configuration.rst
MasterOdin: `snake_case`, e.g.:

* `version_order`
* `default_environment`
* database config options (for example `attr_case` or `mysql_attr_ignore_space`)

etc.

LordSimal:
LGTM but let me add ndm as a reviewer as well since he has been very deep in the whole migrations topic lately.

LordSimal (approved):

docs/en/configuration.rst
```diff
@@ -512,3 +512,17 @@ Within the bootstrap script, the following variables will be available:
      * @var \Symfony\Component\Console\Output\OutputInterface $output The executing command's output object
      * @var \Phinx\Console\Command\AbstractCommand $context the executing command object
      */
+
+Feature Flags
+-------------
+
+For some breaking changes, Phinx offers a way to opt-out of new behavior. The following flags are available:
+
+* ``unsigned_pks``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
```
othercorey, line 521: Are underscores used in other options? Why not just make it camelBack case?

src/Phinx/Config/FeatureFlags.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public static function setFlagsFromConfig(array $config): void
+    {
+        if (isset($config['unsigned_pks'])) {
+            self::$unsignedPks = (bool)$config['unsigned_pks'];
+        }
+        if (isset($config['column_null'])) {
+            self::$columnNull = (bool)$config['column_null'];
+        }
+    }
```
othercorey, line 40: Shouldn't this just be part of the Config class? Wouldn't these options go into the config file?

src/Phinx/Config/FeatureFlags.php
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @param array $config The `feature_flags` section of the config
+     */
+    public static function setFlagsFromConfig(array $config): void
+    {
+        if (isset($config['unsigned_pks'])) {
+            self::$unsignedPks = (bool)$config['unsigned_pks'];
+        }
+        if (isset($config['column_null'])) {
+            self::$columnNull = (bool)$config['column_null'];
```
othercorey: column_null isn't very clear. Can we add default to the name?

src/Phinx/Config/FeatureFlags.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Set the feature flags from the `feature_flags` section of the overall
+     * config.
+     *
+     * @param array $config The `feature_flags` section of the config
+     */
+    public static function setFlagsFromConfig(array $config): void
+    {
+        if (isset($config['unsigned_pks'])) {
+            self::$unsignedPks = (bool)$config['unsigned_pks'];
```
othercorey: I'd also spell this out more: unsigned_primary_keys

docs/en/configuration.rst
othercorey: I also made this mistake thinking we used camelBack. Agree using the same.

src/Phinx/Config/FeatureFlags.php
MasterOdin: Done in ae45d433b71864a2107ac88e860cd7825f7a305d

src/Phinx/Config/FeatureFlags.php
MasterOdin: Added `default` to the end of the name in ae45d433b71864a2107ac88e860cd7825f7a305d. Thought about adding it to the beginning, or trying some other variation, but struggle with what might be the best order of words.

markstory (approved):

PR #2162, opened 2023-01-03
Changed breakpoint column to disallow nulls again

Bilge:
For reasons that remain unclear to me, rather than having a single source of truth, the `breakpoint` column of the migration table may be defined in one of two places, (either [here](https://github.com/cakephp/phinx/blob/2922b17953f4a0910efb4ba5ae7cb528eea5718e/src/Phinx/Db/Adapter/AbstractAdapter.php#L314) or [here](https://github.com/cakephp/phinx/blob/2922b17953f4a0910efb4ba5ae7cb528eea5718e/src/Phinx/Db/Adapter/PdoAdapter.php#L147)). Since 0.13, the breakpoint column may permit null or not, depending on which code path it ends up using. In my case, it suddenly stopped declaring _NOT NULL_ after the upgrade because it used the path which was seemingly missed in #1872. This PR fixes the missed code path.

### DIFF
--- a/src/Phinx/Db/Adapter/AbstractAdapter.php
+++ b/src/Phinx/Db/Adapter/AbstractAdapter.php
@@ -311,7 +311,7 @@ abstract class AbstractAdapter implements AdapterInterface
                 ->addColumn('migration_name', 'string', ['limit' => 100, 'default' => null, 'null' => true])
                 ->addColumn('start_time', 'timestamp', ['default' => null, 'null' => true])
                 ->addColumn('end_time', 'timestamp', ['default' => null, 'null' => true])
-                ->addColumn('breakpoint', 'boolean', ['default' => false])
+                ->addColumn('breakpoint', 'boolean', ['default' => false, 'null' => false])
                 ->save();
         } catch (Exception $exception) {
             throw new InvalidArgumentException(


MasterOdin (approved):

MasterOdin:
Consolidating would be good and just put it all in `createSchemaTable` where the end result is giving you a created schema table, be it a fresh creation or updating a previous installation.

Logic could probably be something like:

```php
    public function createSchemaTable(): void
    {
        try {
            $createTable = !$this->hasTable($this->getSchemaTableName());
            $options = $createTable ? ['id' => false, 'primary_key' => 'version'] : [];
            $columns = $createTable ? [] : array_map(function ($column) {
                return $column->getName();
            }, $table->getColumns());
            if (!in_array('version', $columns, true)) {
                $table->addColumn('version', 'biginteger', ['null' => false]);
            }
            if (!in_array('migration_name', $columns, true)) {
                $table->addColumn('migration_name', 'string', ['limit' => 100, 'default' => null, 'null' => true]);
            }
            if (!in_array('start_time', $columns, true)) {
                $table->addColumn('start_time', 'timestamp', ['default' => null, 'null' => true]);
            }
            if (!in_array('end_time', $columns, true)) {
                $table->addColumn('end_time', 'timestamp', ['default' => null, 'null' => true]);
            }
            if (!in_array('breakpoint', $columns, true)) {
                $table->addColumn('breakpoint', 'boolean', ['default' => false, 'null' => false]);
            }
            $table->save();
        } catch (Exception $exception) {
            throw new InvalidArgumentException(
                'There was a problem creating the schema table: ' . $exception->getMessage(),
                (int)$exception->getCode(),
                $exception
            );
        }
    }
```

where `$table->save();` should be a no-op in the general case where we do not do any `addColumn` calls.

This might break a case where someone was calling `createSchemaTable` manually and wanting the thrown exception when it exists and so the `addColumn` actions fails, but I've no idea if that's a strong enough edge case to not do that.

Thoughts?

Bilge:
You're the principal maintainer. Why are you asking me? 🙃

dereuromark:
You seem to have quite strong opinions from time to time - only fair to consult you once in a while ;)

Bilge:
Having a strong opinion doesn't make you right.

PR #2166, opened 2023-01-06
Test against PHP 8.2

MasterOdin:
PR adds PHP 8.2 to the test matrix, updating the code coverage to run against that version. I also downgraded the version of PHP used for testing sqlsrv to be our lowest version since it's exercising unique code paths compared to the other DB adapters, and I'd like to have confidence we're supporting those paths on the lowest PHP version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1']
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1']
+        php-version: ['7.2', '7.4', '8.0', '8.1', '8.2']
```
dereuromark, line 17: We should probably drop at least 7.2 now (soon).

PR #2167, opened 2023-01-07
Min PHP 7.3

dereuromark:
No need to work with 7.2...8.2 now

https://github.com/cakephp/cakephp/blob/4.x/composer.json#L25 is even 7.4 already

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,31 +13,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1', '8.2']
+        php-version: ['7.3', '7.4', '8.0', '8.2']
         db-type: [sqlite, mysql, pgsql]
         prefer-lowest: ['']
         include:
-          - php-version: '8.2'
+          - php-version: '8.1'
             db-type: mysql
-          - php-version: '7.2'
+          - php-version: '7.3'
             db-type: mysql
             prefer-lowest: prefer-lowest
 
     steps:
     - name: Setup MySQL latest
-      if: matrix.db-type == 'mysql' && matrix.php-version != '7.2'
+      if: matrix.db-type == 'mysql' && matrix.php-version != '7.3'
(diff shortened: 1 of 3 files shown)


MasterOdin (changes requested):
Need to also update the `require.php` value to be `>=7.3` as well so that someone doesn't try to install phinx on 7.2 and it's broken since we're not targeting it anymore.

MasterOdin:
I would consider dropping support for a PHP version a breaking change, and so would necessitate a `0.14.0` release, and so I'd want to hold off doing this until the `0.13.x` release line is more stabilized. Additionally, on upping the min version, I think we should drop both 7.2 and 7.3 and use 7.4 and above to be inline with the CakePHP projects.

dereuromark:
1.x will be 8.1 probably
But dropping a Version is a minor (semantic)

dereuromark:
I dont see a file require.php anywhere.

MasterOdin:
I meant the `php` key/value pair under the `require` object in the `composer.json`. It's currently set to `>=7.2` which should be updated.

MasterOdin:
> But dropping a Version is a minor (semantic)

I guess? Except that since we're still using `0.x` versioning, it means that part of the way through a major release, some people may be locked out of a future release since we cannot go back and do bugfix releases. I also feel like `0.13.x` has ended up not being as stabilized yet as I'd like so I wouldn't want to get people on PHP 7.2 stuck with a slightly buggy version. There's also not really anything in PHP 7.3 that'd be incorporated into phinx beyond bumping dependency/devDependency versions.

All in all, I'd like to hold off doing this until we see `0.13.x` stabilized, and are looking forward to a `0.14.0` release, and then drop minimally PHP 7.2 and 7.3, and support PHP 7.4 purely for compatibility reasons with `cakephp/migrations`.

dereuromark:
Anyone still using < 7.4 is seriously in trouble either way.
I doubt setting 7.3 as minimum has any impact on any normal user of the tool.

As for the rest: Yes, it was rather a rollercoaster, partially due to everyone only being able to commit very small to this project.

But raising the minimum is a super simple task that composer is fully in control, I dont see an issue here.

dereuromark:
Given EOL in 2020 I removed support here now, we will make this clear in the release notes.
But it should simply not be used anymore in favor of 7.3+ on all systems out there.
Composer will give a clear feedback here if violated (and not download updates), therefore no BC issue per se.

Issue #2111, opened 2022-10-07
No More migration possible : error creating AbstractMigration because of type int for the second parameter


zeearth:
Hello,

Since upgrade to 0.13.0, we can't anymore run migration scripte cause of the following error : 
`TypeError: Argument 2 passed to Phinx\Migration\AbstractMigration::__construct() must be of the type int, string given,`

Error throwed by line 824 into Phinx\Migration\Manager.php
`// instantiate it
                    $migration = new $class($environment, $version, $this->getInput(), $this->getOutput());`

$version is initialised at line 779 by the method Util::getVersionFromFileName which return a string value

src/Phinx/Migration/AbstractMigration.php
![image](https://user-images.githubusercontent.com/1954929/194500906-5760cff9-8a14-4b9f-ae89-5e43b800c629.png)

dereuromark:
It seems those are from this commit
https://github.com/cakephp/phinx/commit/88bc125088bfea25ab19c8c5b6f50078dcf26e11#diff-72fcd2be7d86c10f591624e69dcb3bdc646980a23ecf987159a13bb399e7d6ebR73

We might want to then use string instead here, as per other recent changes.

MasterOdin:
The two paths would be:
1. Make it so that `$version` can be a string everywhere
2. Make it so that `$version` is an integer everywhere (what we're currently doing)

For case (2), it would just be modifying `Util::getVersionFromFileName` so that it casts its regex match to an integer and update the type appropriately.

The root cause of the bug is that while we added typing everywhere, we didn't add `declare(strict_types=1);` everywhere as well so there's some amount of implicit conversion happening that need to track down and eliminate (such as here, as well as in casting `$version` in the CLI commands appropriately).

dereuromark:
Should we proceed here?
This also blocks https://github.com/cakephp/migrations/pull/517 now, as it is unclear where to go

As for string change, this would probably be another major now.

zeearth:
Still not working when upgrade to phinx 0.13.3...
I have got error message from duplicate version.
Trying to pass the --date= option to my call, but stil have a cast of string to int in the function getVersionFromFileName in file vendor/robmorgan/phinx/src/Phinx/Util/Util.php
@dereuromark

MasterOdin:
@zeearth Can you provide a link to a repo that reproduces the bug?

dereuromark:
We did not do any release yet
If you want to test it, you need to use `0.x-dev` as constraint for now.

[...]

PR #2169, opened 2023-01-07
Require php-64bit to install phinx

MasterOdin:
PR modifies the `composer.json` such that now installing phinx will only work on php-64bit, which should hopefully avoid errors as shown in #2111 going forward. I think that the number of people using 32-bit PHP is probably low enough to not worry much about having them be stuck on an old version of phinx not realizing it'll break for them, where it's just people using PHP on windows and wherever they're getting PHP is pulling the modern x64 builds.

An alternative approach would be to rewrite phinx such that `$version` could be a string throughout and that we use an alternative data structure than an associative array for sorting migrations keyed by the version (due to how PHP implicitly cast integer looking keys to integers), but I'd say that's a lot of work to benefit a very small number of people.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php-64bit": ">=7.2",
         "cakephp/database": "^4.0",
         "psr/container": "^1.0 || ^2.0",
         "symfony/console": "^3.4|^4.0|^5.0|^6.0",
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -3,7 +3,12 @@
 Phinx Documentation
 ===================
 
-Phinx makes it ridiculously easy to manage the database migrations for your PHP app. In less than 5 minutes, you can install Phinx using Composer and create your first database migration. Phinx is just about migrations without all the bloat of a database ORM system or application framework.
+Phinx makes it ridiculously easy to manage the database migrations for your PHP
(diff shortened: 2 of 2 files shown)


docs/en/index.rst
```diff
@@ -3,7 +3,12 @@
 Phinx Documentation
 ===================
 
-Phinx makes it ridiculously easy to manage the database migrations for your PHP app. In less than 5 minutes, you can install Phinx using Composer and create your first database migration. Phinx is just about migrations without all the bloat of a database ORM system or application framework.
+Phinx makes it ridiculously easy to manage the database migrations for your PHP
+app. In less than 5 minutes, you can install Phinx using Composer and create
+your first database migration. Phinx is just about migrations without all the
+bloat of a database ORM system or application framework.
+
+Phinx requires a 64-bit version of PHP to function.
```
MasterOdin, line 11: The only real change here is the addition of this line. I only added line breaks to the above paragraph.

composer.json
```diff
@@ -36,7 +36,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php-64bit": ">=7.2",
```
dereuromark, line 39: Didnt know this exists

MasterOdin: Neither did I until working on this ticket, but https://getcomposer.org/doc/01-basic-usage.md#platform-packages makes it sound like it'll work how we'd want.

PR #2170, opened 2023-01-07
Update license copyright to CakePHP Software Foundation

MasterOdin:
PR updates the `LICENSE` file to indicate that the CakePHP Software Foundation has held the copyright for Phinx since 2017 when @robmorgan [transferred it cakephp](https://robmorgan.id.au/posts/a-new-home-for-phinx/). It's probably a touch misleading since it was in June of 2017, but adding the months felt ugly, and I'm not sure who might care.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,9 +1,23 @@
 (The MIT license)
 
-Copyright (c) 2012-present Rob Morgan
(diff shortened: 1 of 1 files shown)


dereuromark:
Sounds reasonable

robmorgan:
sounds good to me!

PR #2173, opened 2023-01-23
adding support for column type medium text

sagar-edmingle:


ndm2:
This is already supported via the `limit` option.

```
$table->addColumn('content', 'text', [
    'limit' => \Phinx\Db\Adapter\MysqlAdapter::TEXT_MEDIUM,
]);
```

**https://book.cakephp.org/phinx/0/en/migrations.html#limit-option-and-mysql**

PR #2174, opened 2023-02-06
Add existing comment to column object for mysql

alex-jonas:
When using the `Table::getColumn($name)` method you would expect that afterwards the comment is available in the Column object. 

At least in Mysql this is not the case... therefore my change.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -455,7 +455,7 @@ class MysqlAdapter extends PdoAdapter
     public function getColumns(string $tableName): array
     {
         $columns = [];
-        $rows = $this->fetchAll(sprintf('SHOW COLUMNS FROM %s', $this->quoteTableName($tableName)));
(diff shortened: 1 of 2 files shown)


dereuromark:
Looks good to me

othercorey:
Looks ok to me. Do we have a test harness for this information we can add comments to?

MasterOdin (changes requested):
Please add a row for testing the comment field to this test:

https://github.com/cakephp/phinx/blob/cea47f64fcf9dbbd93b4bb3c80e62730d8594736/tests/Phinx/Db/Adapter/MysqlAdapterTest.php#L1311

You are right we don't set the comment for any of the adapters I think. Would you want to take a stab at adding that in for them?

alex-jonas:
> Would you want to take a stab at adding that in for them?

Sorry, have not much idea about the other adapters

Sinepel (approved):

dereuromark (approved):

MasterOdin (approved):

PR #2175, opened 2023-02-09
Fix blank line above declare statement, see PSR-12

chrisblech:
Fix PSR-12 compliance of generated migration scripts

According to PSR-12 (https://www.php-fig.org/psr/psr-12/), there MUST be a blank line above the `declare` statement (/block)


dereuromark:
Thank you!

PR #2185, opened 2023-03-18
Update configuration.rst

EOM:
Add one new example for custom PDO with SSL_CA for Azure Flex Server

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -48,6 +48,28 @@ This means that:
         ]
     ];
 
+Another example of custom PDO with SSL_CA
+---------------
+
+.. code-block:: php
+
+    // Example of Connection to Azure Flex Server with SSL CA
+    $options = array(
+       PDO::MYSQL_ATTR_SSL_CA => '/var/www/html/MY-PATH-EXAMPLE/Config/DigiCertGlobalRootCA.crt.pem'
+    );
+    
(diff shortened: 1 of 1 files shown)


ndm2:
I don't think that the Phinx docs need to show the PHP basics of creating PDO instances. Besides that, why would it use Azure with a custom SSL CA on a MySQL connection, that's an oddly specific example?

docs/en/configuration.rst
```diff
@@ -48,6 +48,28 @@ This means that:
         ]
     ];
 
+Another example of custom PDO with SSL_CA
+---------------
+
+.. code-block:: php
+
+    // Example of Connection to Azure Flex Server with SSL CA
+    $options = array(
```
dereuromark: Old array syntax?

EOM: @dereuromark Upppsss.. one minute add change..

MasterOdin:
I agree that I think providing an example for using SSL with PDO is outside the scope of the phinx docs. There's a number of different ways you might be looking to use or configure PDO, and that's better covered by docs / tutorials that are specific to PDO.

As an aside, I think you could also set the `PDO::MYSQL_ATTR_SSL_CA` value as a connection option and avoid using PDO altogether.

MasterOdin:
I appreciate you taking your time to submit this PR, but I'm going to close this. To help us understand why the docs were falling short, I'd appreciate if you could provide a run down on what happened when using phinx and the pain points you had with the current docs.

PR #2186, opened 2023-03-29
Fix missing quoting on constraints name for sqlserver

algorhythm:
Test with a constraint that contains `-` sign. It will break with following error: `SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Incorrect syntax near '-'.`


MasterOdin (approved):

MasterOdin:
Thanks!

Issue #2196, opened 2023-05-18
32 bit not supported


pabloelcolombiano:
Hi,

we are having some issues with customers using 32 bit version of Raspbian/Debian.

1. What was the motivation to move from string to int in the first place? Since CakePHP highly recommends to use migrations, the 64 bit requirement indirectly also restrains Cake to it. I am aware that this has been already discussed [here](https://github.com/cakephp/phinx/issues/2111#issuecomment-1279205078) and [there](https://github.com/cakephp/phinx/issues/2192#issuecomment-1514874009), but I would like to understand why this has been done, and if there is not a chance to be less restrictive here.

2. Would you be interested in a PR to get back to the time where version was a string, so 32 bit systems get supported again?

PR #2197, opened 2023-05-22
Cast versions to string (#2196)

pabloelcolombiano:
This PR enables the package to work on 32 bit environement, by keeping the migration version typed as strings.

A PR is ready for the cakephp/migrations package based on this PR.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -389,7 +389,7 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
         }
 
         foreach ($rows as $version) {
-            $result[(int)$version['version']] = $version;
+            $result[$version['version']] = $version;
         }
 
         return $result;
--- a/src/Phinx/Migration/Manager.php
+++ b/src/Phinx/Migration/Manager.php
@@ -304,11 +304,11 @@ class Manager
      * Migrate an environment to the specified version.
      *
      * @param string $environment Environment
-     * @param int|null $version version to migrate to
+     * @param string|null $version version to migrate to
      * @param bool $fake flag that if true, we just record running the migration, but not actually do the migration
      * @return void
      */
-    public function migrate(string $environment, ?int $version = null, bool $fake = false): void
+    public function migrate(string $environment, ?string $version = null, bool $fake = false): void
     {
         $migrations = $this->getMigrations($environment);
         $env = $this->getEnvironment($environment);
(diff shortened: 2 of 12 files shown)


src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -389,7 +389,7 @@ public function getVersionLog(): array
         }
 
         foreach ($rows as $version) {
-            $result[(int)$version['version']] = $version;
+            $result[$version['version']] = $version;
```
MasterOdin, line 392: I'm not totally convinced this won't introduce errors where PHP will automatically cast integer keys to integers, according to `PHP_INT_MAX`, e.g. on PHP-32bit:

```
php > print(PHP_INT_MAX);
2147483647
php > $a = [];
php > $a["2147483647"] = 1;
php > var_dump($a);
array(1) {
  [2147483647]=>
  int(1)
}
php > $a["21474836479999"] = 1;
php > var_dump($a);
array(2) {
  [2147483647]=>
  int(1)
  ["21474836479999"]=>
  int(1)
}
```

and that I'm worried we end up with potential issues around sorting when we use the key.

If we wanted to support PHP 32-bit, my preference would be to make a breaking change and remove the usage of the version as an array index everywhere, and then when we need to do any type of sorting, we do any sorts over the values (be it the version string or a migration object we pull the version off of).

MasterOdin (changes requested):

src/Phinx/Migration/Manager.php
```diff
@@ -847,6 +847,24 @@ function ($phpFile) {
...
     }
 
+    /**
+     * Returns a list of the versions of the migration file
+     * array_keys() of $this->migrations is not reliable on 32 bit systems
+     *
+     * @param string $environment Environment
+     * @return string[]
+     */
+    public function getVersions(string $environment): array
```
MasterOdin, line 857: To clarify, this new function isn't necessary for the rest of the PR changes, just something you'd like downstream?

As a note to myself, on merging this PR, will want to manually do it such that we have two commits, one that does all the other changes and then a second that has just this new function.

pabloelcolombiano: This will be required by cakephp/migrations. We cannot rely on the array_keys of the migtations.
See here: https://github.com/passbolt/migrations/commit/46a3e7bf6f26e71b7c4287497b6d2e47eded1ae2#diff-e60533c90c4ac309c99359e07ec316f9e774c60cacf1ad2d7dfdd615a17a96ebR129

So this new method is needed, and IMO part of the casting to string.

src/Phinx/Db/Adapter/PdoAdapter.php
pabloelcolombiano: I agree, this would be better. Let me investigate in that direction.

src/Phinx/Migration/Manager.php
pabloelcolombiano: Considering your comment above, this might be handled differently though.

dereuromark:
The bc changes here do not get outweighed by the pros
I think we should just keep the documented limit of 64bit
No real System should be using anything else these das anyway

pabloelcolombiano:
It surprised me as well that 32 bits OS' would be an issue in PHP, but yeah we have some people of our community using them.
[See here](https://community.passbolt.com/t/testing-update-to-v4-crashes-on-raspberrypi4/7409/8)

We've forked the present repo and the migration one to support 32 bits. We'll maintain it for a while.

So feel free to close the PR and related issue, I fully understand your point.

PR #2202, opened 2023-06-14
Fix docs: remove new line inside list

albertcansado:


docs/en/migrations.rst
```diff
@@ -902,8 +902,7 @@ Limit Option and MySQL
 
 When using the MySQL adapter, there are a couple things to consider when working with limits:
 
-- When using a ``string`` primary key or index on MySQL 5.7 or below and the default charset of ``utf8mb4_unicode_ci``, you
-must specify a limit less than or equal to 191, or use a different charset.
+- When using a ``string`` primary key or index on MySQL 5.7 or below and the default charset of ``utf8mb4_unicode_ci``, you must specify a limit less than or equal to 191, or use a different charset.
```
MasterOdin, line 905: We'll want to remove the newline on the following list as well then? Do we have other lists with newlines?

MasterOdin (changes requested):
Thanks for the change! Left one comment, if you don't want to deal with it, then we can merge this as is and can address it in a follow-up PR easily enough by someone else.

MasterOdin:
This change was incorporated into #2204, so closing this. Thanks again for your PR.

PR #2204, opened 2023-06-27
Add note about MyISAM length limitation

naidim:
Any version of MySQL tables with the MyISAM storage engine will suffer the same 191 character limitation.


MasterOdin (approved):

MasterOdin:
Thanks!

PR #2218, opened 2023-09-15
Add/improve type declarations for PHP 8.1

othercorey:
cakephp-codesniffer 5 is meant to be paired with cakephp 5 code. One of the key changes is requiring native type declarations where possible.

### DIFF
--- a/src/Phinx/Db/Action/AddColumn.php
+++ b/src/Phinx/Db/Action/AddColumn.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 /**
  * MIT License
@@ -9,6 +10,7 @@ namespace Phinx\Db\Action;
 
 use Phinx\Db\Table\Column;
 use Phinx\Db\Table\Table;
+use Phinx\Util\Literal;
 
 class AddColumn extends Action
 {
@@ -17,7 +19,7 @@ class AddColumn extends Action
      *
      * @var \Phinx\Db\Table\Column
      */
-    protected $column;
+    protected Column $column;
 
     /**
      * Constructor
@@ -40,7 +42,7 @@ class AddColumn extends Action
      * @param array<string, mixed> $options The column options
      * @return static
      */
-    public static function build(Table $table, string $columnName, $type = null, array $options = [])
+    public static function build(Table $table, string $columnName, string|Literal $type, array $options = []): static
     {
         $column = new Column();
         $column->setName($columnName);
--- a/src/Phinx/Db/Action/ChangeColumn.php
+++ b/src/Phinx/Db/Action/ChangeColumn.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 /**
  * MIT License
@@ -9,6 +10,7 @@ namespace Phinx\Db\Action;
 
 use Phinx\Db\Table\Column;
 use Phinx\Db\Table\Table;
+use Phinx\Util\Literal;
 
 class ChangeColumn extends Action
 {
@@ -17,14 +19,14 @@ class ChangeColumn extends Action
      *
      * @var \Phinx\Db\Table\Column
      */
-    protected $column;
+    protected Column $column;
 
     /**
      * The name of the column to be changed
      *
      * @var string
      */
-    protected $columnName;
+    protected string $columnName;
 
     /**
      * Constructor
@@ -55,7 +57,7 @@ class ChangeColumn extends Action
      * @param array<string, mixed> $options Additional options for the column
      * @return static
      */
-    public static function build(Table $table, string $columnName, $type = null, array $options = [])
+    public static function build(Table $table, string $columnName, string|Column|Literal $type, array $options = []): static
     {
         $column = new Column();
         $column->setName($columnName);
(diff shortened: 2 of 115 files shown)


tests/Phinx/Migration/ManagerTest.php
```diff
@@ -6034,7 +6036,7 @@ public function testMigrationWithDropColumnAndForeignKeyAndIndex()
         $adapter->disconnect();
 
         $this->manager->setConfig($config);
-        $this->manager->migrate('production', '20190928205056');
+        $this->manager->migrate('production', 20190928205056);
```
othercorey, line 6039: Some tests were passing  string as a version instead of an int. This appears to only affect tests as the migration loading code converts to int.

src/Phinx/Db/Action/AddColumn.php
```diff
@@ -40,7 +42,7 @@ public function __construct(Table $table, Column $column)
      * @param array<string, mixed> $options The column options
      * @return static
      */
-    public static function build(Table $table, string $columnName, $type = null, array $options = [])
+    public static function build(Table $table, string $columnName, string|Literal $type, array $options = []): static
```
othercorey, line 45: This appears to default to null only for a few tests which were omitting a type. This doesn't seem to be valid as the phinx code checks for either Literal or string. Updated tests to specify a type.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -97,7 +103,7 @@ public function testConnectionWithSocketConnection()
         }
 
         $options = ['unix_socket' => getenv('MYSQL_UNIX_SOCKET')] + MYSQL_DB_CONFIG;
-        $adapter = new MysqlAdapter(MYSQL_DB_CONFIG, new ArrayInput([]), new NullOutput());
+        $adapter = new MysqlAdapter($options, new ArrayInput([]), new NullOutput());
```
othercorey, line 106: Wasn't passing the $config with socket configured so not sure if anything is running this test.

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -723,11 +723,11 @@ protected function getDefaultConstraint(string $tableName, string $columnName)
...
 
     /**
-     * @param int $tableId Table ID
-     * @param int $indexId Index ID
+     * @param string $tableId Table ID
+     * @param string $indexId Index ID
      * @return array
      */
-    protected function getIndexColums(int $tableId, int $indexId): array
+    protected function getIndexColums(string $tableId, string $indexId): array
```
othercorey, line 730: This is passed strings not ints and is internal to this adapter so changed the types.

src/Phinx/Db/Action/ChangeColumn.php
```diff
@@ -55,7 +57,7 @@ public function __construct(Table $table, string $columnName, Column $column)
      * @param array<string, mixed> $options Additional options for the column
      * @return static
      */
-    public static function build(Table $table, string $columnName, $type = null, array $options = [])
+    public static function build(Table $table, string $columnName, string|Column|Literal $type, array $options = []): static
```
MasterOdin: Looks like the docstring on this function was wrong, and similar to above, we should only accept `string|Literal` here as that's what `setType` accepts.

src/Phinx/Db/Adapter/SqlServerAdapter.php
MasterOdin: An unfortunate "quirk" of PDO in that it returns numerical columns as strings.

MasterOdin (approved):

src/Phinx/Db/Table.php
```diff
@@ -291,12 +293,12 @@ public function reset(): void
...
      *
      * @param string|\Phinx\Db\Table\Column $columnName Column Name
-     * @param string|\Phinx\Util\Literal|null $type Column Type
+     * @param string|\Phinx\Util\Literal $type Column Type
      * @param array<string, mixed> $options Column Options
      * @throws \InvalidArgumentException
      * @return $this
      */
-    public function addColumn($columnName, $type = null, array $options = [])
+    public function addColumn(string|Column $columnName, string|Literal $type, array $options = [])
```
ndm2, line 301: This looks like it was possibly an accident? When supplying a `Column` instance in the first argument, it should already hold the type, right?

> Error: src/Table.php:67:47: PossiblyNullArgument: Argument 2 of Phinx\Db\Table::addColumn cannot be null, possibly null value provided (see https://psalm.dev/078)


https://github.com/cakephp/migrations/actions/runs/6269091941/job/17025092704?pr=628

othercorey: Migrations support for phinx 0.15 is here https://github.com/cakephp/migrations/pull/638

othercorey:
@MasterOdin You can override the header settings for phinx from the cakephpcs defaults now if you want. I don't think there will be any large-scale changes after this.

You can also move away from cakephp-cs. It was useful for validating the type declarations and type-hints, but you could pull out the specific rules you want into a custom config.

PR #2220, opened 2023-09-19
Add declare strict_types to seed template

MasterOdin:
PR adds a `declare(strict_types=1);` to the seed template, which brings it inline with how our migrations template file works, where declare was added in https://github.com/cakephp/phinx/pull/1819. We should be highly forcing people to have correct types usage when interacting with their database to prevent any sort of unexpected data loss/corruption, such as talked about in the linked PR.

### DIFF
--- a/src/Phinx/Console/Command/SeedCreate.php
+++ b/src/Phinx/Console/Command/SeedCreate.php
@@ -198,7 +198,7 @@ class SeedCreate extends AbstractCommand
 
         $namespace = $config instanceof NamespaceAwareInterface ? $config->getSeedNamespaceByPath($path) : null;
         $classes = [
-            '$namespaceDefinition' => $namespace !== null ? ('namespace ' . $namespace . ';') : '',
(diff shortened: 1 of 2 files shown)


src/Phinx/Console/Command/SeedCreate.php
```diff
@@ -198,7 +198,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $namespace = $config instanceof NamespaceAwareInterface ? $config->getSeedNamespaceByPath($path) : null;
         $classes = [
-            '$namespaceDefinition' => $namespace !== null ? ('namespace ' . $namespace . ';') : '',
+            '$namespaceDefinition' => $namespace !== null ? (PHP_EOL . 'namespace ' . $namespace . ';' . PHP_EOL) : '',
```
MasterOdin, line 201: Updating this line so that it looks and works the same as the similar line we have for migrations:

https://github.com/cakephp/phinx/blob/09d8f17f782bfb4463e0c5282fd6ef4953a6e13b/src/Phinx/Console/Command/Create.php#L292

PR #2222, opened 2023-09-19
Ensure PostgresAdapter::useIdentity is always initialized

othercorey:
Wrapper package would call setConection() without connect() causing useIdentity to not be set.


othercorey:
This was the wrong change.

PR #2223, opened 2023-09-19
Ensure PostgresAdapter::useIdentity is always initialized

othercorey:
This is causing issues with packages that run their own Manager and use setConnection() directly.


MasterOdin (approved):
Fine not having a test as it's not super easily tested change since we don't directly expose this property, and would have to get at it as part of introspection of constructed SQL which is brittle.

PR #2224, opened 2023-09-22
Allow Table::addColumn() to use type from Column object again

othercorey:
refs https://github.com/cakephp/phinx/pull/2218#discussion_r1333812072

I think this was removed after fixing a test that passed an invalid null, but we should allow the Column to define the type.


ndm2:
Should the exception in that method maybe use `$action->getColumn()->getType()` instead of `$type`?

othercorey:
> Should the exception in that method maybe use `$action->getColumn()->getType()` instead of `$type`?

That seems to make sense. The type can be invalid when defined by either Column or string.

MasterOdin (approved):

PR #2232, opened 2023-10-20
📝 Update docs to reflect current PHP requirements.

coderanger:
Spent a while trying to get things working on 7.4 before I realized this was just out of date. Looks like 8.1 is required as of 0.15 when things were moved to Cake 5?


MasterOdin (approved):
Thanks for updating the docs!

PR #2233, opened 2023-10-29
PostgreSQL - possibility to create "character varying" column without limit

Mapiiik:
This change in the PostgreSQL adapter allows you to create a "character varying" column without a limit. Otherwise, even if we set the limit to "null" it still takes this default.


src/Phinx/Db/Adapter/PostgresAdapter.php
```diff
@@ -1067,7 +1067,7 @@ public function getSqlType(Literal|string $type, ?int $limit = null): array
             case static::PHINX_TYPE_DOUBLE:
                 return ['name' => 'double precision'];
             case static::PHINX_TYPE_STRING:
-                return ['name' => 'character varying', 'limit' => 255];
+                return ['name' => 'character varying'];
```
othercorey, line 1065: Would it make more sense to generate a `text` data type for postgres for this?

Mapiiik: The text type works fine for PostgreSQL, but we are used to using it for multi-line inputs and character variing for single-line inputs (this is also the default for CakePHP for forms).

PostgreSQL directly recommends not to restrict this for no reason, it supposedly has no performance impact.
See: https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_varchar.28n.29_by_default

dereuromark:
Are we OK to merge here?

ajibarra:
@dereuromark @othercorey  Changes looks good. Can we merge it?

othercorey:
@MasterOdin Can you check this?

MasterOdin:
Thanks for opening the PR. I agree that there should be a way to get a column without a limit, and I think explicitly setting `null` should be how that's done, not getting rid of the default altogether, and not also adjusting only one type for one adapter.

Closing this PR in favor of the approach in #2393.

PR #2235, opened 2023-11-03
Fix using test command with manual set config

### LINKED ISSUE #2234: Support for testing "outer" config
I'm using a Phinx as a part of the Nette framework.
There is a Dependency Ijection contaier, where I can setup the Phinx configuration directly without the 'phinx.yml' file, using the `setConfig` method.

Unfortunately the `Test` command force to use the file.

https://github.com/cakephp/phinx/blob/26753bdb66827383781eb0033b34cb467da1d12c/src/Phinx/Console/Command/Test.php#L62

It will be nice to have a possibility to check the connection with this configuration.
Everything that I need is to encapsulate this line like this:

```php
if($this->hasConfig() === false) {
    $this->loadConfig($input, $output);
}
```

MasterOdin:
Fixes #2234 

PR fixes a bug where it was not possible to use the test command if you had manually done `->setConfig` on it, vs having an existing config file. This gets this command to operate similar to all the other commands, where they have the same check on running `->loadConfig` or not (via the `->bootstrap` method).

### DIFF
--- a/src/Phinx/Console/Command/Test.php
+++ b/src/Phinx/Console/Command/Test.php
@@ -59,7 +59,10 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $this->loadConfig($input, $output);
+        if (!$this->hasConfig()) {
+            $this->loadConfig($input, $output);
+        }
+
         $this->loadManager($input, $output);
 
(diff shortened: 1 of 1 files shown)


zeleznypa:
@MasterOdin It works like a charm. Thank you very much.

PR #2236, opened 2023-11-04
:recycle: default to null for type of column in addColumn

gaydamakha:
As accepting `null` for `$type` parameter was rollbacked in #2224, I think it is more consistent to default to null value after its removal in #2218. Passing `Column` object in first argument makes the `$type` parameter useless. So, the PR simplifies a bit the adding of a new column (modified in the unit test).

### DIFF
--- a/src/Phinx/Db/Table.php
+++ b/src/Phinx/Db/Table.php
@@ -298,7 +298,7 @@ class Table
      * @throws \InvalidArgumentException
      * @return $this
      */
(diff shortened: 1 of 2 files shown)


othercorey:
Leave this up to @ndm2

ndm2:
Makes sense to me, no need to force passing a value that isn't going to be used.

ndm2 (approved):

PR #2237, opened 2023-11-09
Loosen psr/container version constraint

### LINKED ISSUE #2227: Revise psr/container compatiblity
With https://github.com/cakephp/phinx/pull/2106 the support for `psr/container` 1.x was dropped, now supporting version 2.x only.

Now I wonder if this change is really necessary. Currently this means that using Phinx with some Symfony packages 5.x (namely `symfony/dependency-injection`) is not possible anymore since these require `psr/container` 1.x; support for `psr/container` 2.x was added with Symfony 6.x.

AFAIS the `ContainerInterface` / `getContainer()` usage in Phinx itself is very slim:

- https://github.com/search?q=repo%3Acakephp%2Fphinx%20psr%5Ccontainer&type=code
- https://github.com/search?q=repo%3Acakephp%2Fphinx+getContainer&type=code

Nothing here requires API from `psr/container` 2.x unless I'm missing something.

So maybe consider supporting `psr/container` 1.x again?

MasterOdin:
Closes #2227 

cakephp/cakephp@5.0.2 loosened their version constraint of `psr/container` when then allows us to loosen it as well, allowing both `^1.1` or `^2.0` to be used based on other things you might install alongside phinx.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     "require": {
         "php-64bit": ">=8.1",
         "cakephp/database": "^5.0",
-        "psr/container": "^2.0",
+        "psr/container": "^1.1|^2.0",
         "symfony/console": "^6.0",
         "symfony/config": "^3.4|^4.0|^5.0|^6.0"
     },


MasterOdin:
@dereuromark This should not have been merged yet since one of the CI checks is failing from this change. I'm also not sure how best to resolve this, where `composer update --prefer-lowest` won't install `psr/container@^1.1` as it doesn't install the lowest version of transitive dependencies, and one of the transitive dependencies ends up with `psr/container@^2.0`. This causes `dereuromark/composer-prefer-lowest` tool to fail.

I suppose a "fix" would be to install the lowest version of a transitive dependency as a `devDependency` within phinx (at least in the CI check), but that's certainly hacky.

dereuromark:
composer is sometimes a bit stupid
https://github.com/cakephp/phinx/commit/62fda40e1870f2896a91130f6b0c9658f18a332d should make it green again

PR #2243, opened 2023-11-14
Use UUID_BLOB for binary16

dereuromark:
Follow https://github.com/cakephp/phinx/pull/2239

Also needs https://github.com/cakephp/cakephp/pull/17421


MasterOdin (approved):
LGTM. Merge once the core PR ha also been approved?

dereuromark:
Yes, we need to wait for the core PR to match approval status.
Also, I needed 

    str_contains(strtolower($column), 'blob')
    
since the column usually comes in uppercase

but actually 

    if (($col === 'binary' && $length === 16) || strtolower($column) === 'uuid_blob')

should suffice

markstory (approved):

PR #2246, opened 2023-11-30
Add Symfony 7 support

Slamdunk:


composer.json
```diff
@@ -40,15 +40,15 @@
         "cakephp/database": "^5.0.2",
         "psr/container": "^1.1|^2.0",
         "symfony/console": "^6.0",
-        "symfony/config": "^3.4|^4.0|^5.0|^6.0"
+        "symfony/config": "^3.4|^4.0|^5.0|^6.0|^7.0"
```
dereuromark, line 43: It might at some point be time to drop old v3 support.
I wonder if the full range is even still possible, with more types being added.

Also, we are using symfony/console v6 only, but then this range for config and yaml?
Seems a bit strange.

dereuromark: But this can be a follow up task at some point.

composer.json
```diff
@@ -40,15 +40,15 @@
         "cakephp/database": "^5.0.2",
         "psr/container": "^1.1|^2.0",
         "symfony/console": "^6.0",
```
dereuromark: What about console?

Slamdunk: Done, thanks

PR #2249, opened 2023-12-04
Validate datetime for version.

### LINKED ISSUE #2248: Validate version int range
If people by accident add a

    config/Migrations/202311128071500_Migration.php

instead of

    config/Migrations/20231112807150_Migration.php

it actually works but is a date for in the future (year 20231 or sth^^).
This messes up the "order".

Fair to say we would directly spot this if we would also display the actual date, not just the timstamp, in the migration output.

I propose we validate this range a bit more carefully.
E.g.
-10 years, +20years
?

Refs https://github.com/dereuromark/cakephp-queue/pull/392#issuecomment-1837282338

dereuromark:
Resolves https://github.com/cakephp/phinx/issues/2248

For legacy reasons and BC on the test harnesses I allowed the 0 to pass for now as well.
I think thats fair since most mistakes would happen with actual datetime values.

### DIFF
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -383,7 +383,7 @@ abstract class AbstractMigration implements MigrationInterface
     protected function validateVersion(int $version): void
     {
         $length = strlen((string)$version);
-        if (!$version || $length === 14) {
+        if ($length === 14) {
             return;
         }
 
--- a/src/Phinx/Migration/AbstractMigration.php
+++ b/src/Phinx/Migration/AbstractMigration.php
@@ -76,6 +76,8 @@ abstract class AbstractMigration implements MigrationInterface
      */
     final public function __construct(string $environment, int $version, ?InputInterface $input = null, ?OutputInterface $output = null)
     {
+        $this->validateVersion($version);
+
         $this->environment = $environment;
         $this->version = $version;
 
@@ -369,5 +371,22 @@ abstract class AbstractMigration implements MigrationInterface
     public function shouldExecute(): bool
     {
         return true;
+    }
+
+    /**
+     * Makes sure the version int is within range for valid datetime.
+     * This is required to have a meaningful order in the overview.
+     *
+     * @param int $version Version
+     * @return void
+     */
+    protected function validateVersion(int $version): void
+    {
+        $length = strlen((string)$version);
+        if (!$version || $length === 14) {
+            return;
+        }
+
+        throw new RuntimeException('Invalid version `' . $version . '`, should be in format `YYYYMMDDHHMMSS` (length of 14).');
     }
 }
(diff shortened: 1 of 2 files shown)


MasterOdin:
I've debated on if the appropriate place to put this is as a runtime check within initialization the migration, vs within the [`getMigrations`](https://github.com/cakephp/phinx/blob/dd8746fcb08d71ae334a0fbf67a497accc6868cc/src/Phinx/Migration/Manager.php#L753) method in which the migrations are initialized (and which all other methods use to get the list of migrations). I'm somewhat leaning towards the latter with the concept that if some alternative versioning scheme was desired, it'd probably be easier to support via the `getMigrations` method where we have access to the config object to determine said alternative versioning strategy.

Thoughts?

> For legacy reasons and BC on the test harnesses I allowed the 0 to pass for now as well.

I'm curious what legacy reason beyond test harnesses exist to support `0` as a version. I feel like having support for this _just_ for tests seems wrong, where the tests should be updated as it's maybe a bug for production to allow this?

-----

I do like the approach here of only validating that the version integer is a valid date string, and not putting limits on what the allowed range of dates is, as I think that has the potential to be limiting for less value.

src/Phinx/Migration/AbstractMigration.php
```diff
@@ -370,4 +372,21 @@ public function shouldExecute(): bool
...
+     *
+     * @param int $version Version
+     * @return void
+     */
+    protected function validateVersion(int $version): void
+    {
+        $length = strlen((string)$version);
+        if (!$version || $length === 14) {
+            return;
+        }
```
MasterOdin: A follow-up here would be to validating that the various pieces of the string are all valid, e.g. that don't have `00` or `13` for `MM`, potentially including also complexity in not allowing invalid dates (e.g. no `30` for `MM == 02`).

dereuromark: Sure, but then it wouldnt fit in a "minor" release probably as people could have existing invalid ones that are valid in range (sorting).
So for now I would like to keep it as is, but in the next major we could try to limit it to a more sane value in general.

dereuromark:
> where the tests should be updated as it's maybe a bug for production to allow this?

could be a follow up if you want to further narrow the possible valid values.

dereuromark:
I pushed the update, was a quick write-up.
Hopefully we can get this in now asap to avoid people running into this like me and others did.

MasterOdin (approved):

MasterOdin:
Easy enough to move this check to somewhere else if we decide we need to down the road without breaking BC I think.

I do think this check does break BC though in that it's possible to bypass the date version naming scheme that phinx uses, and just use any number you wanted (e.g. having a counter), though you'd just have to manually create each migration. Not sure if anyone is doing that, so probably fine to just merge it into the next `0.15.x` release, and easy to revert if there's enough pushback, or minimally just have them override the base `AbstractMigration` class.

PR #2263, opened 2024-01-30
Fix docs

dereuromark:


MasterOdin:
This change is wrong though? When you provide a class name, then we hit this line to get the filename:

https://github.com/cakephp/phinx/blob/5c28445848a1daba6883dd94f77c0feef0db3d28/src/Phinx/Console/Command/Create.php#L183

which per the comment:

https://github.com/cakephp/phinx/blob/5c28445848a1daba6883dd94f77c0feef0db3d28/src/Phinx/Util/Util.php#L103-L105

Issue #1635, opened 2019-10-25
Can I override addTimestamps?


ltdev22:
Can I override addTimestamps() method? I don't like some of the default settings it using like the `timestamp` type for both created_at and updated_at and that only the created_at gets the current timestamp.

I'd like to set both  created_at and updated_at as `datetime` type and default value on both to be NOW(), is that possible some how?

dereuromark:
Does someone want to address this, or make a PR?

PR #2273, opened 2024-05-02
Option for addTimeStamps() to use datetime

joshbmarshall:
Adds a feature flag to use datetime on MySQL for addTimestamps() function instead of timestamp.

Implements #1635 

timestamp columns do not support dates past 2038-01-19 on MySQL

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -521,6 +521,11 @@ For some breaking changes, Phinx offers a way to opt-out of new behavior. The fo
 * ``unsigned_primary_keys``: Should Phinx create primary keys as unsigned integers? (default: ``true``)
 * ``column_null_default``: Should Phinx create columns as null by default? (default: ``true``)
 
+Since MySQL ``TIMESTAMP`` fields do not support dates past 2038-01-19, you have the option to use ``DATETIME`` field
+types for fields created by the ``addTimestamps()`` function:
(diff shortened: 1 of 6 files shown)


dereuromark:
@cakephp/phinx Any reviews?

markstory (commented):
Looks good to me.

dereuromark:
Will this be ported to migrations? I wonder if we should make a list of prs to migrate etc

markstory:
> Will this be ported to migrations? I wonder if we should make a list of prs to migrate etc

In a way yes, I'm hoping to update the internals of migrations to use cake's db reflection so we'll inherit the fix made in cake. I can also port this change as is over it should apply.

MasterOdin (approved):

MasterOdin (commented):
We could probably enable this by default and it wouldn't have the largest blast radius, given that not sure how often people are relying on these auto-generated columns having a specific type. We could then add also getting timestamps or datetimes as an option to `addTimestamps` to help the transition for people who were relying on the timestamp type.

joshbmarshall:
I had considered requesting to make it the default since timestamp fields are deprecated in MySQL. But didn't want to add delays to getting the option into the code so I can start using it ;)

It should definitely become the default on the next major revision of phinx if it has to wait until then.

Issue #2256, opened 2024-01-19
Native uuid type for mariadb


nicosp:
mariadb has a native uuid type since 10.7. We already have:
- uuid -> char(36)
- binaryuuid -> binary(16)

Can we also add a "nativeuuid" type? In terms of PDO it's string so it can be easily supported in cake as well.

What do you think?

PR #2282, opened 2024-05-27
Add "nativeuuid" type for databases with native uuid support

nicosp:
This is to allow the use of the native uuid for mariadb 10.7+ users. For other database it's simply an alias to the existing uuid type. Sqlite is explicitly not supported because it doesnt have a native type.

Related Issue: https://github.com/cakephp/phinx/issues/2256

Note: To test this you need an to add 'nativeuuid' to Cake\Database\TypeFactory $_types.

Once this is accepted I can also do the change in CakePHP.

### DIFF
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -53,6 +53,7 @@ interface AdapterInterface
     public const PHINX_TYPE_JSON = 'json';
     public const PHINX_TYPE_JSONB = 'jsonb';
     public const PHINX_TYPE_UUID = 'uuid';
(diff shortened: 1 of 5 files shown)


dereuromark:
@nicosp Phinx is now being separated standalone.
To also port this into CakePHP (Migrations), this would also need a 2nd PR to
https://github.com/cakephp/migrations

nicosp:
> @nicosp Phinx is now being separated standalone. To also port this into CakePHP (Migrations), this would also need a 2nd PR to https://github.com/cakephp/migrations

Ok, which branch should I target there?

dereuromark:
current one, so 4.x one is fine.

MasterOdin:
Thanks for the PR!

PR #2286, opened 2024-06-04
Fix quoting DB name when creating/dropping

MasterOdin:
Port of https://github.com/cakephp/migrations/pull/717 to phinx which fixed using values that needed being escaped for mysql, but did not fix it for postgres or sqlserver.

I think there is still issues in some other places the DB name can be referenced in being quoted, but they're used within `WHERE` clauses within a SQL statement, and the fix there is to rework the queries to be parameterized vs using `sprintf`, which can be done in a follow-up PR.

### DIFF
--- a/src/Phinx/Db/Adapter/MysqlAdapter.php
+++ b/src/Phinx/Db/Adapter/MysqlAdapter.php
@@ -1301,13 +1301,13 @@ class MysqlAdapter extends PdoAdapter
 
         if (isset($options['collation'])) {
             $this->execute(sprintf(
-                'CREATE DATABASE `%s` DEFAULT CHARACTER SET `%s` COLLATE `%s`',
-                $name,
+                'CREATE DATABASE %s DEFAULT CHARACTER SET `%s` COLLATE `%s`',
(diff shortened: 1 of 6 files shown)


src/Phinx/Db/Adapter/MysqlAdapter.php
```diff
@@ -1301,13 +1301,13 @@ public function createDatabase(string $name, array $options = []): void
 
         if (isset($options['collation'])) {
             $this->execute(sprintf(
-                'CREATE DATABASE `%s` DEFAULT CHARACTER SET `%s` COLLATE `%s`',
-                $name,
+                'CREATE DATABASE %s DEFAULT CHARACTER SET `%s` COLLATE `%s`',
+                $this->quoteColumnName($name),
```
MasterOdin, line 1305: The upstream patch used `$this->quoteTableName` which does an escape around any `.` characters in addition to calling `quoteColumnName` where it's assuming that it's a separator between schema and table name. Given that database names don't have schema references, better I think to use `quoteColumnName` directly.

PR #2287, opened 2024-06-04
Fix escaping database name for SqlServerAdapter::dropDatabase

MasterOdin:
PR fixes a bug introduced in #2286 that broke dropping databases in sqlserver. I've fixed the syntax so that it actually works for database names that need escaping, discovering as well that the escaping of column names was also broken. Per https://learn.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms176027(v=sql.105)?redirectedfrom=MSDN, to escape an identifier that has brackets, need to double up the brackets, not use `\`.

### DIFF
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -211,7 +211,7 @@ class SqlServerAdapter extends PdoAdapter
      */
     public function quoteColumnName(string $columnName): string
     {
-        return '[' . str_replace(']', '\]', $columnName) . ']';
+        return '[' . str_replace(']', ']]', $columnName) . ']';
     }
 
     /**
@@ -1176,7 +1176,7 @@ ORDER BY T.[name], I.[index_id];";
     {
         $databaseName = $this->quoteColumnName($name);
         if (isset($options['collation'])) {
-            $this->execute(sprintf('CREATE DATABASE %s COLLATE [%s]', $databaseName, $options['collation']));
+            $this->execute(sprintf('CREATE DATABASE %s COLLATE %s', $databaseName, $options['collation']));
         } else {
             $this->execute(sprintf('CREATE DATABASE %s', $databaseName));
         }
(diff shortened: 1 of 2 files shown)


src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1204,12 +1204,13 @@ public function hasDatabase(string $name): bool
...
-USE master;
-IF EXISTS(select * from sys.databases where name=N'$name')
-ALTER DATABASE [$name] SET SINGLE_USER WITH ROLLBACK IMMEDIATE;
-DROP DATABASE [$name];
-SQL;
+        $sql = sprintf(
+            'USE master;
+            IF EXISTS(select * from sys.databases where name=N\'' . $name . '\')
+            ALTER DATABASE %1$s SET SINGLE_USER WITH ROLLBACK IMMEDIATE;
+            DROP DATABASE %1$s;',
```
MasterOdin, line 1211: By using `%1$s`, I'm able to replace both instances with a single value. Downside is that cannot use `"` around the string like do in other places. Could use a heredoc, and then run the `$sql` through `sprintf` on a separate line, but that didn't feel in line with other places `sprintf` is used.

src/Phinx/Db/Adapter/SqlServerAdapter.php
```diff
@@ -1176,7 +1176,7 @@ public function createDatabase(string $name, array $options = []): void
     {
         $databaseName = $this->quoteColumnName($name);
         if (isset($options['collation'])) {
-            $this->execute(sprintf('CREATE DATABASE %s COLLATE [%s]', $databaseName, $options['collation']));
+            $this->execute(sprintf('CREATE DATABASE %s COLLATE %s', $databaseName, $options['collation']));
```
MasterOdin, line 1179: Not sure why the collate argument had brackets around it as it's not an identifier and shouldn't be created as such. I also tried `'` and `"` and it didn't work either.

This has probably been broken for a long time, and I guess no one is creating databases with specific collations.

PR #2292, opened 2024-06-17
Fix Google Analytics

robmorgan:
Phinx.org used an older version of Google Analytics (Universal), so this PR migrates it to the latest version - GA4.


MasterOdin:
Does anyone check these google analytics and do they provide any meaningful value to the homepage?

robmorgan:
@MasterOdin are you asking in regards to maintenance or regulation? I'm still selling a screencast so I check them probably twice a year for insights.

PR #2295, opened 2024-07-17
Mark password as a SensitiveParameter

edudobay:
Mark the `$password` parameter as sensitive, so its value is redacted if present in a stack trace.  This already happens for PDO but not for this method:

```
#0 /src/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PdoAdapter.php(96):
  PDO->__construct('pgsql:dbname=aa...', 'my-username', Object(SensitiveParameterValue), Array)
#1 /src/vendor/robmorgan/phinx/src/Phinx/Db/Adapter/PostgresAdapter.php(115):
  Phinx\Db\Adapter\PdoAdapter->createPdoConnection('pgsql:dbname=aa...', 'my-username', 'my password leaked...', Array)
```

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -41,6 +41,7 @@ use Phinx\Migration\MigrationInterface;
 use Phinx\Util\Literal;
 use ReflectionProperty;
 use RuntimeException;
+use SensitiveParameter;
 use Symfony\Component\Console\Output\OutputInterface;
 use UnexpectedValueException;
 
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):

MasterOdin:
`SensitiveParameter` was added in PHP 8.2, is there any issue with using this in PHP 8.1 which we still support? I'm not super well versed in attributes, and from test suite I guess it works and the attribute is ignored if it doesn't exist?

e: playing around with a simple snippet, does seem like older versions of PHP will just ignore the attribute and continue to work the same, so this won't break anything, but is also only applicable to people running PHP 8.2+.

edudobay:
Yes, that's exactly what I expected, it only benefits people running PHP 8.2+, but should not impact older versions in any way.

Issue #2297, opened 2024-08-03
Unable to seed rows using CURRENT_TIMESTAMP


phrfpeixoto:
Hello, I'm trying to write a simple seeder:

```php
class MySeeder extends AbstractSeed
{
    public function run(): void
    {
        $data = [
            [
                'user_name' => 'Administrator',
                'user_creation' => 'CURRENT_TIMESTAMP',
            ]
        ];
        $table = $this->table('users');
        $table->insert($data)->saveData();
    }
}
```

Executing this seeder fails:

```
[PDOException (22007)]
  SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: 'CURRENT_TIMESTAMP' for column 'user_creation' at row 1
```

I explored the code a bit, and tried using Literal and Expression, but got the same result.
Is there a way to achieve this?

MasterOdin:
Yeah, this is currently not possible, and is a bug in this block of code:

https://github.com/cakephp/phinx/blob/5f9ee844fdf3a34af7aac211f9463758c35bc18b/src/Phinx/Db/Adapter/PdoAdapter.php#L378-L415

The root here is that when inserting, we create a prepared statement where all values are then binded. Stuff like `CURRENT_TIMESTAMP` (or any function really) cannot be a binded parameter, but should be in the SQL directly. As such, the fix would be that when we go to construct the SQL statement, instead of just saying all values are going to be binded, we need to step through each, and either insert the literal (if using `Literal`) or a `?` to be binded.

MasterOdin:
I think the bug also exists in the `PdoAdapter::insert` method as well, so will want to fix it in both places.

MasterOdin:
Using literals is now supported as of `0.16.3` released today.

PR #2195, opened 2023-05-13
Fix 2193 Postgres GENERATED ALWAYS

captain-redbeard:
This pull request implements the discussed changes in [2193](https://github.com/cakephp/phinx/issues/2193).

## Changes

- Set's the default generated clause to GENERATED BY DEFAULT to allow database seeding with the OOTB setup.
- Updates the PostgresAdapter insert statements to include OVERRIDING SYSTEM VALUE to allow inserting into columns that were created with the GENERATED ALWAYS clause. This allows database seeding.

## How to set GENERATED ALWAYS

In your migration file add the option `generated`.

```php
...

public function up()
{
    $this->table('my_table')
        ->addColumn('id', 'integer', [
            'autoIncrement' => true,
            'limit' => 11,
            'generated' => \Phinx\Db\Adapter\PostgresAdapter::GENERATED_ALWAYS,
          ])
          ...
}
```

PR #2300, opened 2024-08-04
Consolidate insert/bulkInsert implementations

MasterOdin:
PR consolidates the `insert` and `bulkInsert` implementations between `PdoAdapter` and `PostgresAdapter` so that we once again only have the functions implemented in `PdoAdapter`, while retaining the override behavior added in #2195. Doing this as a precursor to #2297 so that it's easier to fix the bug where only need to change the functions in `PdoAdapter` and not in other places as well.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -324,7 +324,7 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
             $this->quoteTableName($table->getName())
         );
         $columns = array_keys($row);
-        $sql .= '(' . implode(', ', array_map([$this, 'quoteColumnName'], $columns)) . ')';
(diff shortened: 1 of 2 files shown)


phrfpeixoto:
Docs? What changed for bulk insert?

MasterOdin:
The change was that previously there was:

* `PdoAdapter::insert`
* `PdoAdapter::bulkInsert`
* `PostgresAdapter::insert`
* `PostgresAdapter::bulkInsert`

where the code for the `PostgresAdapter` variants of the two functions was wholesale copied from the `PdoAdapter` with a very slight tweak inserted (done in #2195). The PR here gets us to have only:

* `PdoAdapter::insert`
* `PdoAdapter::bulkInsert`

again, while still having the functionality added in #2195 as done via a new protected method that the `PostgresAdapter` overrides. This made it easier for me to do #2301 where I only needed to make the change in 2 places (`PdoAdapter::insert` and `PdoAdapter::bulkInsert`) as opposed to 4 places (aforementioned and `PostgresAdapter::insert` and `PostgresAdapter::bulkInsert`), as well as any other potential future changes made to these two functions.

For end users, nothing practical has changed in how the methods would be used in the adapters.

PR #2301, opened 2024-08-04
Support Literal values in inserts

### LINKED ISSUE #2297: Unable to seed rows using CURRENT_TIMESTAMP

MasterOdin:
Closes #2297 

PR adds support for using `\Phinx\Util\Literal` in insert operations, so that can do stuff like insert `CURRENT_TIMESTAMP`. Previously, we were always passing all values in the insert data as binded variables to the escaped SQL statement. Now, the literal values will be inserted directly into the escaped SQL statement, while non literals values will continue to be binded.

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -336,9 +336,22 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
             $sql .= '(' . implode(', ', array_map([$this, 'quoteValue'], $row)) . ');';
             $this->output->writeln($sql);
         } else {
-            $sql .= '(' . implode(', ', array_fill(0, count($columns), '?')) . ')';
+            $sql .= '(';
+            $vals = [];
+            $values = [];
+            foreach ($row as $value) {
+                $values[] = $value instanceof Literal ? (string)$value : '?';
+                if (!($value instanceof Literal)) {
+                    if (is_bool($value)) {
+                        $vals[] = $this->castToBool($value);
+                    } else {
+                        $vals[] = $value;
+                    }
+                }
+            }
+            $sql .= implode(', ', $values) . ')';
             $stmt = $this->getConnection()->prepare($sql);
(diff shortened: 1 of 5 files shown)


dereuromark:
@markstory Do we need to port this to migrations as well?

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -358,6 +371,10 @@ protected function quoteValue(mixed $value): mixed
             return 'null';
         }
 
+        if ($value instanceof Literal) {
+            return (string)$value;
+        }
```
MasterOdin, line 376: `quoteValue` is used only in the insert functions for dry-run output as well as two queries in the sqlite adapter where it always operates on a string, so not much risk this breaks something.

tests/Phinx/Db/Adapter/MysqlAdapterTest.php
```diff
@@ -2215,9 +2247,46 @@ public function testInsertData()
...
+        $this->assertEquals('test', $rows[1]['column3']);
         $this->assertEquals('foo', $rows[2]['column3']);
     }
 
+    public function testInsertLiteral()
+    {
+        $data = [
+            [
+                'column1' => 'value1',
+                'column3' => Literal::from('CURRENT_TIMESTAMP'),
```
phrfpeixoto, line 2259: 👍

PR #2307, opened 2024-09-06
Append sqlite suffix to name on info output

### LINKED ISSUE #2298: Include suffix in "using database" for sqlite
For sqlite, doing a migration/seed gives an output similar to following:

```
using adapter sqlite
using database ./data
```

where the we only print out the database name, but omit the suffix. We should show the database name _and_ the suffix, e.g. `./data.sqlite3` for above.

MasterOdin:
Closes #2298 

PR makes it so that the database name informational output will contain the suffix for sqlite connections, e.g.:

```
 $ bin/phinx migrate
Phinx by CakePHP - https://phinx.org. 0.x-dev

using config file phinx.yaml
using config parser yml
using migration paths
 - /home/mpeveler/github/phinx/db/migrations
using seed paths
 - /home/mpeveler/github/phinx/db/seeds
warning no environment specified, defaulting to: testing
using adapter sqlite
using database phinx_test.sqlite3
ordering by creation time

 == 20240906173954 Foo: migrating
 == 20240906173954 Foo: migrated 0.0158s

All Done. Took 0.0184s
```

This should make the info more informational on what's happening. Working on this, I am definitely leaning increasingly towards deprecating/removing the `suffix` option altogether and require that users only use the `name` option instead, which is similar to how most ORMs handle it.

### DIFF
--- a/src/Phinx/Console/Command/AbstractCommand.php
+++ b/src/Phinx/Console/Command/AbstractCommand.php
@@ -12,6 +12,7 @@ use InvalidArgumentException;
 use Phinx\Config\Config;
 use Phinx\Config\ConfigInterface;
 use Phinx\Db\Adapter\AdapterInterface;
+use Phinx\Db\Adapter\SQLiteAdapter;
 use Phinx\Migration\Manager;
 use Phinx\Util\Util;
 use RuntimeException;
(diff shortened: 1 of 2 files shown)


zeleznypa:
Unfortunately it cause the https://github.com/cakephp/phinx/issues/2311

InvisibleSmiley:
FTR one used to be able to create a SqliteAdapter with empty options (e.g. for tests), but since this change at least the "name" option needs to be provided. I guess in the context of normal invocation that should always be set (cf. Phinx config environment section).

MasterOdin:
@InvisibleSmiley Just from reading the code, I think the previous behavior was that if no options was provided, then it'd create a file named `.sqlite3`, right? I'm wondering if there's a use for this outside of tests, and then for that, would it be fine to just have it default to `:memory:` if no options were set?

Though how hard is it to just mandate people use either `name` or `memory` option, and I feel like it's better to be explicit with how you want the connection to be setup vs relying on defaults.

InvisibleSmiley:
I'm not saying anything needs to be changed, given that likely in real-world setups no-one actually instantiates SQLiteAdapter manually. If you consider that class to be internal, there is no need to even document the change in behavior/requirements.

Personally I would have had the code handle that case more gracefully, e.g. like this:
```
$name = $options['name'] ?? null;
if ($name === self::MEMORY) {
  return '';
}
```

PR #2314, opened 2024-09-27
Add docs about namespaces for paths

matheusab:
Hello, I didn't find any references to namespaces in the docs when creating migrations, so, after debugging the code, I found out how it could be done.
This PR adds instructions for specifying namespaces in migration paths.


MasterOdin (approved):
Thanks for your contribution!

Need to figure out what to do with https://github.com/cakephp/phinx/blob/0.x/docs/en/namespaces.rst which has similar info, but is not linked anywhere, though I think this is a good start to at least have this be documented minimally.

PR #2317, opened 2024-10-16
Write CLI errors to stderr

### LINKED ISSUE #2270: (At least) Non-interactive commands should write errors to STDERR
Right now commands like `migrate` write errors to STDOUT. This is wrong. Errors, especially from commands that are commonly used non-interactively, need to be written to STDERR so that scripts executing Phinx can easily distinguish between information (STDOUT) and errors (STDERR), possibly redirecting either of them to different locations.

Here's how:
https://symfony.com/doc/current/console/style.html#writing-to-the-error-output

InvisibleSmiley:
Fixes #2270.

In the end I chose to copy/adapt OutputStyle::getErrorOutput as a new protected method on the abstract command rather than using SymfonyStyle because:

1. creating SymfonyStyle requires both $input and $output, but some existing methods only receive $output
2. SymfonyStyle behaves differently than the plain OutputStyle, which might introduce new issues

I also fixed the exit code for the case where no aliases exist, applying the rule "error output -> exit code 1".

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -4,6 +4,12 @@
 
     <rule ref="CakePHP"/>
 
+    <rule ref="SlevomatCodingStandard.Namespaces.UnusedUses">
+        <properties>
+            <property name="searchAnnotations" value="true"/>
+        </properties>
+    </rule>
+
     <arg value="nps"/>
 
     <file>src/</file>
--- a/src/Phinx/Console/Command/ListAliases.php
+++ b/src/Phinx/Console/Command/ListAliases.php
@@ -72,13 +72,11 @@ class ListAliases extends AbstractCommand
             return self::CODE_SUCCESS;
         }
 
-        self::getErrorOutput($output)->writeln(
-            sprintf(
-                '<error>No aliases defined in %s</error>',
-                Util::relativePath($this->config->getConfigFilePath())
-            )
+        $output->writeln(
+            '<comment>warning</comment> no aliases defined in ' . Util::relativePath($this->config->getConfigFilePath()),
+            $this->verbosityLevel
         );
 
-        return self::CODE_ERROR;
+        return self::CODE_SUCCESS;
     }
 }
--- a/src/Phinx/Console/Command/ListAliases.php
+++ b/src/Phinx/Console/Command/ListAliases.php
@@ -68,14 +68,14 @@ class ListAliases extends AbstractCommand
                 ),
                 $this->verbosityLevel
             );
-
-            return self::CODE_SUCCESS;
+        } else {
+            $output->writeln(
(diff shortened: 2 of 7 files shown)


src/Phinx/Console/Command/ListAliases.php
```diff
@@ -68,15 +68,17 @@ function ($alias, $class) use ($maxAliasLength, $maxClassLength) {
...
 
-        return self::CODE_SUCCESS;
+        self::getErrorOutput($output)->writeln(
+            sprintf(
+                '<error>No aliases defined in %s</error>',
+                Util::relativePath($this->config->getConfigFilePath())
+            )
+        );
+
+        return self::CODE_ERROR;
```
MasterOdin: I'm not sure not having any aliases would constitute the command failing per se, should just mean that you've not set any up. I'd leave this as always returning `self::CODE_SUCCESS`.

InvisibleSmiley: I'd be OK with that if the message would be downgraded to a warning or info; otherwise it'll be inconsistent (and as I wrote, my goal was to achieve consistency).

MasterOdin: Symfony does not ship with a warning style out of the box for the console, so I'm fine downgrading it to use `<info>` instead, or I guess could do similar to this:

https://github.com/cakephp/phinx/blob/efd7cf8b8fe3ed3ce0a6bb77fd079bcb5b732026/src/Phinx/Console/Command/AbstractCommand.php#L436

where we wrap a "warning" text in `<comment>` so it gives a bit of visual separation to the user.

InvisibleSmiley: Implemented as suggested

src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -204,7 +204,7 @@ public function connect(): void
      */
     public static function getSuffix(array $options): string
     {
-        if ($options['name'] === self::MEMORY) {
+        if (($options['name'] ?? '') === self::MEMORY) {
```
InvisibleSmiley, line 207: This produced a notice when running tests (with `error_reporting = E_ALL` in php.ini)

phpcs.xml
```diff
@@ -4,6 +4,12 @@
 
     <rule ref="CakePHP"/>
 
+    <rule ref="SlevomatCodingStandard.Namespaces.UnusedUses">
```
InvisibleSmiley, line 7: The tests do not contain all necessary imports (uses) right now because without this PHPCS will complain.
I added the missing imports to MigrateTest and will leave the rest for someone else to clean up.

MasterOdin (approved):

MasterOdin:
Thanks for your PR!

PR #2318, opened 2024-10-17
Document specific query builder functions

MarwanSalim:
Enhance the documentation by including the available method options, as these methods have now been added to the source code.

Reference:
https://github.com/cakephp/phinx/blob/b0f4397dd57eb91bd5369c2770a4f2656a3f993c/src/Phinx/Db/Adapter/PdoAdapter.php#L236C1-L282C6


MasterOdin (approved):
Thanks for the addition!

Issue #1665, opened 2019-12-18
Deferrable constraints


JulienChavee:
Hello,

On some DBMS (like PostgreSQL) it is possible to have deferrable constraints.
This is useful and we use this feature for almost all of our Foreign Key constraints.

It seems that it is not possible to have this feature when using the "addForeignKey" function.

It there something that I missed or it is not implemented in Phinx?
If it is not implemented, is there any chance to have this feature supported in Phinx?

dereuromark:
PR welcome

PR #2322, opened 2024-11-25
Add deferrable constraints management for PSQL

JulienChavee:
Hello,

With this PR, I am trying to implement the deferrable constraints management for PSQL, related to issue #1665.
This is my first PR here, let me know what I can improve, happy to contribute into it.

### DIFF
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -1388,6 +1388,10 @@ class PostgresAdapter extends PdoAdapter
             $def .= " ON UPDATE {$foreignKey->getOnUpdate()}";
         }
 
+        if ($foreignKey->getDeferrableMode()) {
+            $def .= " {$foreignKey->getDeferrableMode()}";
+        }
+
         return $def;
     }
 
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -960,14 +960,17 @@ class PostgresAdapter extends PdoAdapter
      * @param string $tableName Table name
      * @return array
      */
-    protected function getForeignKeys(string $tableName): array
+    public function getForeignKeys(string $tableName): array
     {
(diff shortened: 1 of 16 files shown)


src/Phinx/Db/Adapter/PostgresAdapter.php
MasterOdin, line 1: Can you modify https://github.com/JulienChavee/phinx/blob/809ae6b71053c29d17b605cbc30f9bb3bbc1ce62/src/Phinx/Db/Adapter/PostgresAdapter.php#L963 and have it set the appropriate `deferrable` value? Looking at https://www.postgresql.org/docs/current/infoschema-table-constraints.html, I think that this can be done by selecting `is_deferrable` and `initially_deferred` from the `information_schema.table_constraints` table.

MasterOdin (approved):

MasterOdin (changes requested):
Whoops, hit the wrong button. See above for requested change. Thanks for making this PR to improve phinx!

src/Phinx/Db/Adapter/PostgresAdapter.php
JulienChavee: Sure I can look to add that.
Any reason to make this function protected and not public? I could use it in the PostgresAdapterTest to verify the deferrable status of the FK created.

Also, would you prefer to have both values like
```
is_deferrable -> true/false
initially_deferred -> true/false
```

Or having as you proposed a `deferrable` with either DEFERRED/IMMEDIATE/NOT_DEFERRED?

MasterOdin: Sorry, that was a bad suggestion. I had thought that `getForeignKeys` was public and also returned `ForeignKey` object, and so would just add to that, but since it's not public and doesn't return that, then don't want the change for now.

Making it public is something we could do, but I'd want to then modify it to be like `getColumns` where it returns an array of objects, vs an array of associative arrays.

MasterOdin (approved):

MasterOdin:
Thanks for your contribution!

PR #2331, opened 2024-12-03
Fix SQLiteAdapter::getColumns with Literal column

### LINKED ISSUE #2330: Cannot get columns for sqlite with Literal column type
> That said, it's not clean to just assume that `getPhinxType` returns a string for key "name" and then pass that to `parseDefaultValue` in `getColumns`. This leads to a type error when "name" is a Literal. Instead, a meaningful exception should be thrown in that case that mentions the table name, column name, and given type (string). But that's another issue. 

 _Originally posted by @InvisibleSmiley in [#2319](https://github.com/cakephp/phinx/issues/2319#issuecomment-2478521804)_

MasterOdin:
Fixes #2330

PR fixes a bug where `SQLiteAdapter::getColumns` would throw an exception if the table had a column with a `Literal` type. This was due to a type error when passing the type of `SQLiteAdapter::parseDefaultValue` which requires the type be a string. While I could have modified `parseDefaultValue` to accept a `string|Literal`, as the function is defined as `protected`, this would have been a BC break, and so I went with converting the `$type['name']` to a string in the caller (`getColumns`) instead.

### DIFF
--- a/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
@@ -2438,7 +2438,7 @@ INPUT;
         $this->assertCount(1, $columns);
         $column = array_pop($columns);
         $this->assertSame('real_col', $column->getName());
-        $this->assertEquals('Phinx\Util\Literal', get_class($column->getType()));
+        $this->assertInstanceOf(Literal::class, $column->getType());
         $this->assertEquals(Literal::from('FOO'), $column->getType());
     }
 
--- a/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
@@ -2430,13 +2430,16 @@ OUTPUT;
     public function testLiteralSupport()
     {
         $createQuery = <<<'INPUT'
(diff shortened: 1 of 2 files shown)


tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -2430,13 +2430,16 @@ public function testAlterTableForeignKeyConstraintValidationNotRunningWithDisabl
...
     {
         $createQuery = <<<'INPUT'
-CREATE TABLE `test` (`real_col` DECIMAL)
+CREATE TABLE `test` (`real_col` FOO)
 INPUT;
         $this->adapter->execute($createQuery);
         $table = new Table('test', [], $this->adapter);
         $columns = $table->getColumns();
         $this->assertCount(1, $columns);
-        $this->assertEquals(Literal::from('decimal'), array_pop($columns)->getType());
```
MasterOdin, line 2439: `DECIMAL` is an accepted type, so `->getType()` was returning `"decimal"`, but since we were doing `assertEquals`, which equates to `Literal::from('decimal') == 'decimal'`, the test was passing (as `Literal` was being converted to a string).

This is also why I added an assertion on the actual type of `getType()`.

PR #2334, opened 2024-12-09
Docs: Set PDO attributes after variable has been assigned

piotr-cz:
As in PR title


dereuromark:
why?

MasterOdin (approved):

MasterOdin:
Thanks for the contribution @piotr-cz.

@dereuromark Before this change, if using that example, then the line doing `$this->pdo->setAttribute` would result in a fatal error as `$this->pdo` had not yet been assigned (it was happening after). Now it'll work as expected. 😄

dereuromark:
Sry the diff looked differently for me so I saw it being accessed before Set.
Makes sense as in this pr 👍

PR #2338, opened 2025-01-07
Add config file path to docs example

piotr-cz:
When initializing config manually, one must provide path to phinx config file, so variables such as `%%PHINX_CONFIG_DIR%%` are resolved properly.

See stack overflow > https://stackoverflow.com/a/50693483/1012616


MasterOdin (approved):
Thanks for the PR

PR #2344, opened 2025-03-01
Trim trailing semicolon before executing query

### LINKED ISSUE #2340: Remove ';' in generated query
Hi, 

=== Context
Phinx :) for this work.

I'm writing an Adapter for PDO-OCI.

I'm getting blocked by some things that should be fixed in 

=== Observed behaviour
In the PDO adapter, migrated generates a query with a semi-colon at the end. ';'  is a query separator, and is not allowed by PDO-OCI (ORA-00911 - Invalid character)

=== Expected behaviour
I would ask to remove this ';', given that it won't change anything for other databases.

I have patched it manually (by overriding it in the OCIAdapter I have created) for now, I will push a PR for that when I'll have some time.

Yopai:
Closes #2340 

(Currently writing an OCI Adapter)
OCI doesn't allow a trailing semicolon.

Still output the semicolon, so that output can be easily copy/pasted

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
             n=$((n+1))
             sleep 2
           done
-          mysql -h 127.0.0.1 -u root -proot -D phinx -e 'SELECT 1;'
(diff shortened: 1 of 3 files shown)


MasterOdin (approved):
Thanks for the PR!

PR #2348, opened 2025-04-11
Support Cake\I18n\Date and DateTime for inserts

dereuromark:
Refs https://github.com/cakephp/migrations/pull/838

### DIFF
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -329,22 +329,22 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
      */
     protected function getInsertParameters(array $row): array
     {
-        $values = [];
+        $params = [];
         foreach ($row as $value) {
             if ($value instanceof Literal) {
                 continue;
             } elseif ($value instanceof DateTime) {
-                $values[] = $value->toDateTimeString();
+                $params[] = $value->toDateTimeString();
             } elseif ($value instanceof Date) {
-                $values[] = $value->toDateString();
+                $params[] = $value->toDateString();
             } elseif (is_bool($value)) {
-                $values[] = $this->castToBool($value);
+                $params[] = $this->castToBool($value);
             } else {
-                $values[] = $value;
+                $params[] = $value;
             }
         }
 
-        return $values;
+        return $params;
     }
 
     /**
--- a/src/Phinx/Db/Adapter/PdoAdapter.php
+++ b/src/Phinx/Db/Adapter/PdoAdapter.php
@@ -322,6 +322,32 @@ abstract class PdoAdapter extends AbstractAdapter implements DirectActionInterfa
     }
 
     /**
+     * Get the parameters array for prepared insert statement
+     *
+     * @param array $row Row to be inserted into DB
+     * @return array
+     */
+    protected function getInsertParameters(array $row): array
+    {
+        $values = [];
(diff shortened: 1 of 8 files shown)


dereuromark:
@MasterOdin Can we tag a new patch release after this, please? So we are good for Core migrations plugin to also be green?

src/Phinx/Db/Adapter/PdoAdapter.php
```diff
@@ -430,6 +432,10 @@ public function bulkinsert(Table $table, array $rows): void
                 foreach ($row as $v) {
                     if ($v instanceof Literal) {
                         continue;
+                    } elseif ($v instanceof DateTime) {
+                        $vals[] = $v->toDateTimeString();
```
MasterOdin: Would doing `str($v);` have a similar effect as doing `$v->toDateTimeString()`?

dereuromark: No
Because thats causing it to transform to the string representation, being usually localized (mm/dd/yy) etc

dereuromark:
Feel free to add a test if you have some time.
Once this is merged, the migrations PR should also be green I guess.

MasterOdin:
TIL that `instanceof` will not error if the right-hand operator points at a class that doesn't exist in the current environment.

So `cakephp/i18n` just needs to be a `require-dev` entry for the linter and analyzer to pass, but doesn't need to be there for downstream consumers (unless they want to use it for bulk inserts).

dereuromark:
I double checked: https://github.com/cakephp/migrations/pull/843 green :)

MasterOdin (approved):

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
```diff
@@ -1923,7 +1923,7 @@ public function testBulkInsertDates()
         $rows = $this->adapter->fetchAll('SELECT * FROM table1');
         $this->assertEquals('foo', $rows[0]['name']);
         $this->assertEquals('bar', $rows[1]['name']);
-        $this->assertMatchesRegularExpression('/[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/', $rows[0]['created']);
+        $this->assertMatchesRegularExpression('/[0-9]{4}-[0-9]{2}-[0-9]{2}/', $rows[0]['created']);
```
dereuromark: Any idea whats that about?

dereuromark:
Given the different ms before/after the regex was probably a good idea.
But if it works out now like this, fair enough :)

tests/Phinx/Db/Adapter/SQLiteAdapterTest.php
MasterOdin: Sqlite doesn't have types in the way other databases do, where they're more "suggestions" on what the data in the column is vs what it actually is. So while all other databases will coerce a date string (`2025-04-17`) into a proper datetime for the column type, sqlite leaves it as a date string.

dereuromark:
Should we squash merge this?

dereuromark:
Do you think we could make a new release?

MasterOdin:
> Should we squash merge this?

My preference is always to squash on merge to `main` (or `0.x` here) so that the history for `main` remains clean, while still have the PRs to see the full intermediary history of a given commit on `main`.

> Do you think we could make a new release?

Done: https://github.com/cakephp/phinx/releases/tag/0.16.7

PR #2350, opened 2025-04-15
Remove cakephp/cakephp dev dependency

MasterOdin:
The `cakephp/cakephp` dev dependency was added in 37307f8189b1ed307c4bd00235190b1a984c49e2, but to my knowledge we don't directly depend on it, so removing it as it's unnecessary bloat. I think it's also potentially allowing #2348 to not throw an exception as we don't directly rely on `cakephp/i15n`, but it's included from `cakephp/cakephp`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,6 @@
     "require-dev": {
         "ext-json": "*",
         "ext-pdo": "*",
-        "cakephp/cakephp": "^5.0.2",
         "cakephp/cakephp-codesniffer": "^5.0",
         "phpunit/phpunit": "^9.5.19",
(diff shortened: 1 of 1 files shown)


dereuromark:
Then i18n should be kept for the tests to work out imo

MasterOdin:
None of the current tests or code rely on `cakephp/i18n` though?
 https://github.com/cakephp/phinx/pull/2348 adds such a dependency, but it needs to be a regular dependency, or better an optional dependency.

dereuromark:
Require dev is optional. It only exists when u test the repo

Issue #258, opened 2014-06-03
Allow a count for rollback/migrate.


rquadling:
Hi.

A feature request.

I'm doing some incremental development, but I need to be able to test things backwards and forwards through the migrations.

Currently, I have to track which timestamp to use as my target, but I can see it is always 3 migrations (or 4, 5, etc. as I develop).

A new option of `--count (-o) Number of migrations to process` would be useful for this.

Regards,

Richard Quadling.

ipoz:
:+1:

rquadling:
Now I'm more familiar with Phinx, I'll create a feature branch for a PR later this (or early next) week.

steffkes:
Having used MyBatis Migrations before, they're doing the following:

> The **up** command runs the **do** section of all pending migrations in order, one after the other.
> The **down** command runs the **undo** section of the *last applied migration* only.
http://mybatis.github.io/migrations/updown.html

while specifying the number of migrations to run is obviously fine as well, i've seen this behavior used by multiple migration libraries .. so we should probably go for that?

rquadling:
I'm no longer sure this is a useful feature for me as I developed the breakpoint feature https://github.com/rquadling/phinx/tree/feature/BreakpointSupport / https://github.com/robmorgan/phinx/pull/497 which is what I really needed.

Waiting on some feedback from Phinx management on this.

robmorgan:
I see this as pretty low priority.

dereuromark:
The current timestamps without autocomplete are sure not nice to work with.
A different way, like this count sure could be more user friendly.

othercorey:
Closing due to lack of activity or out of scope. If there is new interest, please open a PR.

MasterOdin:
@othercorey could this be re-opened?

dereuromark:
@MasterOdin Do we need to keep this open?
It is useful, but it seems there is no interest in adding it. Once a PR here is done in the future, that would still be fine.
But tickets like these seem to be just dead ones.

MasterOdin:
Count command has been added to migrate with `-k` / `--count` option flag. Need to add it to rollback still.

dereuromark:
https://github.com/cakephp/migrations/pull/874
can be ported back on the rollback part - then this ticket would be fully resolved - 11 years after the fact^^

PR #2360, opened 2025-06-09
Add count option to migrate

MasterOdin:
Half does #258

PR adds a new `-k` / `--count` option to the migrate command which allows for running arbitrary number of migrations, without relying on target or date. Decided to use the `-k` shorthand as `-c`,  `-n`, and `-t` were already used, and it can stand for "kount" which I feel like other tools have done (vs `-o` as suggested in the issue, which feels more arbitrary).

I decided to deal with rollback in a follow-up PR as there's a workaround for this functionality there, which is just running the rollback command however many times you need.

### DIFF
--- a/src/Phinx/Console/Command/Migrate.php
+++ b/src/Phinx/Console/Command/Migrate.php
@@ -38,15 +38,17 @@ class Migrate extends AbstractCommand
         $this->setDescription('Migrate the database')
             ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
             ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')
+            ->addOption('--count', '-k', InputOption::VALUE_REQUIRED, 'The number of migrations to run')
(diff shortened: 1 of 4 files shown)


src/Phinx/Console/Command/Migrate.php
```diff
@@ -38,15 +38,17 @@ protected function configure(): void
         $this->setDescription('Migrate the database')
             ->addOption('--target', '-t', InputOption::VALUE_REQUIRED, 'The version number to migrate to')
             ->addOption('--date', '-d', InputOption::VALUE_REQUIRED, 'The date to migrate to')
+            ->addOption('--count', '-k', InputOption::VALUE_REQUIRED, 'The number of migrations to run')
```
dereuromark, line 41: wouldnt -n be better here than k?

MasterOdin: `-n` is taken up by builtin option that comes from symfony console:

```
 -n, --no-interaction  Do not ask any interactive question
```

dereuromark:
@MasterOdin 
I added rollback idea here:
https://github.com/cakephp/migrations/pull/874

Maybe you want to enable this for phinx asap and we can do one full release with all the API consistent.

PR #2361, opened 2025-06-09
Unpin version of slevomat/coding-standard

### LINKED ISSUE #1760: False positive for SlevomatCodingStandard.Classes.PropertyDeclaration.MultipleSpacesBetweenTypeHintAndProperty
This file from my project (https://github.com/cakephp/phinx/blob/be536ffb2a5e97fd36d4935964370216531f3637/tests/Phinx/Config/AbstractConfigTest.php) is suddenly failing with 4.19.0 with the following:

```
FILE: /home/mpeveler/github/phinx/tests/Phinx/Config/AbstractConfigTest.php
------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------
 14 | ERROR | [x] There must be exactly one space between type hint and property $migrationPath.
    |       |     (SlevomatCodingStandard.Classes.PropertyDeclaration.MultipleSpacesBetweenTypeHintAndProperty)
------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------

Time: 70ms; Memory: 10MB
```

If I remove the two properties or make the first one `protected string $migrationPath`, then the error goes away. Trying to run `vendor/bin/phpcbf` and it fails to fix whatever error it's complaining about. Have absolutely no idea why this singular class within my project has issues as it's structured like all my other test classes.

The file validated fine with 4.18.1 and earlier.

MasterOdin:
slevomat/coding-standard has fixed https://github.com/slevomat/coding-standard/issues/1760 with 8.19.1, so we should be able to unpin the version we were testing against.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
         "cs-check": "phpcs",
         "cs-fix": "phpcbf",
         "stan": "phpstan analyse",
-        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.9.0 && composer require --dev slevomat/coding-standard:8.18.1 && mv composer.backup composer.json",
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.9.0 && mv composer.backup composer.json",
         "lowest": "validate-prefer-lowest",
         "lowest-setup": "composer update --prefer-lowest --prefer-stable --prefer-dist --no-interaction && cp composer.json composer.backup && composer require --dev dereuromark/composer-prefer-lowest && mv composer.backup composer.json",
         "test": "phpunit --colors=always"


composer.json
```diff
@@ -74,7 +74,7 @@
         "cs-check": "phpcs",
         "cs-fix": "phpcbf",
         "stan": "phpstan analyse",
-        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.9.0 && composer require --dev slevomat/coding-standard:8.18.1 && mv composer.backup composer.json",
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.9.0 && mv composer.backup composer.json",
```
dereuromark, line 77: phpstan should also be 2.1+ at some point.

PR #2362, opened 2025-06-10
Update phpstan to 2.1.x

MasterOdin:
PR updates phpstan from 1.9.x to 2.1.x, which is the latest major/minor release at this time.

PR #2363, opened 2025-06-10
Bump phpstan to 1.12.x

MasterOdin:
PR updates our usage of phpstan from 1.9.x to 1.12.x, which is the latest minor release of the 1.x branch. While preferable to move to phpstan 2.1, per #2362, there's a lot of errors that get thrown out that we'd need to fix before doing so, as I don't want to add them to the baseline.


src/Phinx/Db/Adapter/SQLiteAdapter.php
```diff
@@ -1168,8 +1168,8 @@ protected function calculateNewTableColumns(string $tableName, string|false $col
             $writeColumns[] = $writeName;
         }
 
-        $selectColumns = array_filter($selectColumns, 'strlen');
-        $writeColumns = array_filter($writeColumns, 'strlen');
+        $selectColumns = array_filter($selectColumns, fn($v) => $v && strlen($v) > 0);
+        $writeColumns = array_filter($writeColumns, fn($v) => $v && strlen($v) > 0);
```
MasterOdin, line 1172: Per https://github.com/phpstan/phpstan/issues/2266, the maintainer of phpstan made a conscious decision to encode this where the callable passed to `array_filter` must return a boolean value (I think), which strlen doesn't do.

Passing `'strlen'` would also implicitly filter out both emtpy strings and `false` values, whereas doing it explicitly throws a type error passing a boolean to `strlen`, so this is certainly more type safe in general I guess.

PR #2366, opened 2025-06-26
Fix errors with custom autoload

0zd0:
When we prefix this library with https://github.com/BrianHenryIE/strauss it creates its own composer autoloader and it is impossible to use cli


MasterOdin (approved):

MasterOdin:
Thanks for the contribution. This is available in [0.16.10](https://github.com/cakephp/phinx/releases/tag/0.16.10).

PR #2370, opened 2025-07-22
fixed typo in format warning message

Bogghi:
This pull request updates the naming convention for migration class names in the `Create` command to use PascalCase instead of CamelCase. The changes ensure consistency in the format and provide clear guidance to users.

### Updates to naming convention:

* [`src/Phinx/Console/Command/Create.php`](diffhunk://#diff-97d6d3bf0fdbfbabdbc99e17ca8b51971dceff15c96afb1e32c0d3431a6101adL58-R58): Updated the argument description in the `configure` method to specify that the migration class name should be in PascalCase.
* [`src/Phinx/Console/Command/Create.php`](diffhunk://#diff-97d6d3bf0fdbfbabdbc99e17ca8b51971dceff15c96afb1e32c0d3431a6101adL197-R197): Updated the error message in the `execute` method to indicate that the migration class name must use PascalCase format.

### DIFF
--- a/src/Phinx/Console/Command/Create.php
+++ b/src/Phinx/Console/Command/Create.php
@@ -55,7 +55,7 @@ class Create extends AbstractCommand
         parent::configure();
 
         $this->setDescription('Create a new migration')
-            ->addArgument('name', InputArgument::OPTIONAL, 'Class name of the migration (in CamelCase)')
+            ->addArgument('name', InputArgument::OPTIONAL, 'Class name of the migration (in PascalCase)')
             ->setHelp(sprintf(
                 '%sCreates a new database migration%s',
                 PHP_EOL,
@@ -194,7 +194,7 @@ class Create extends AbstractCommand
         } else {
             if (!Util::isValidPhinxClassName($className)) {
                 throw new InvalidArgumentException(sprintf(
-                    'The migration class name "%s" is invalid. Please use CamelCase format.',
+                    'The migration class name "%s" is invalid. Please use PascalCase format.',
                     $className,
                 ));
             }


src/Phinx/Console/Command/Create.php
```diff
@@ -55,7 +55,7 @@ protected function configure(): void
         parent::configure();
 
         $this->setDescription('Create a new migration')
-            ->addArgument('name', InputArgument::OPTIONAL, 'Class name of the migration (in CamelCase)')
+            ->addArgument('name', InputArgument::OPTIONAL, 'Class name of the migration (in PascalCase)')
```
dereuromark, line 58: Thats not a typo, the naming convention so far dictates:
- CamelCase
- camelBacked

as official wordings
https://book.cakephp.org/5/en/intro/conventions.html

PascalCase is just an alias that is much more uncommon in our circles and not clear to many (camel from the animal is).

dereuromark: I guess strictly speaking it would be
UpperCamelCase
vs
lowerCamelCase

https://en.wikipedia.org/wiki/Camel_case

But we simplified that 20+ years ago and have that convention for 2 decades now using the above naming.

Bogghi: ok, thanks a lot for the clarification and now i understand and share your convention. Wouldn't be useful to specify camedBacked to avoid confusion when reading the warning in the console?

dereuromark: But the warning ist correct.
Or what do you want to change?

Bogghi: My advice would be this: since the format of the name should be `camelBacked` as per convention it would be useful to specify it instead of saying `camelCase`. Today I've lost about 10 minutes trying to understand why it wasn't accepting the camelCase name of the class. 

It's a minor change but it would make it clear what the format of the class should be.

dereuromark: But it is supposed to be CamelCase and thats the message

dereuromark:
"CamelCaseControllerName" is correct format. Closing.

PR #2378, opened 2025-10-25
Add test for mysql zero scale column def fix

MasterOdin:
PR adds a test for the fix done in #2377, validating the case where when `scale` is set to `0`, then we do still get `decimal(15,0)` in the column definition.

For reference, the default value for precision if the arguments are omitted is 10.

### DIFF
--- a/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
+++ b/tests/Phinx/Db/Adapter/MysqlAdapterTest.php
@@ -2732,13 +2732,13 @@ INPUT;
         $this->adapter->connect();
         $table = new Table('test_table', ['id' => false], $this->adapter);
         $table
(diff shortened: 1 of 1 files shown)


dereuromark:
What PR/issue was that related to?

MasterOdin:
@dereuromark #2377, which you merged 4 days ago without any tests.

I suspect there's also a similar bug in the implementation for at least sqlserver as well, though that's a bit more tricky as whereas mysql requires both of precision and scale or neither, sqlserver allows omitting scale per [their docs](https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql).

dereuromark:
True, but the fix was quite obvious from the PR.

MasterOdin:
Regardless of simplicity of a given fix, I would still like to see tests to prevent regressions in the future.

PR #2381, opened 2025-10-26
Update phpunit to ^10.5

MasterOdin:
PR updates the phpunit dev dependency to `^10.5` which is the latest major version that supports 8.1, which phinx supports. At the moment, I don't think there's value in also doing something like was in #2371 where we try to support multiple phpunit versions, unless there's something that comes up that's only possible in phpunit 11+ (which to my knowledge we've not hit).


dereuromark:
It is actually quite easy to support all of them:
https://github.com/cakephp/cakephp/blob/87b366cb714f6872e5609a9749ccbfd529886c1b/composer.json#L65
Just can be done piece by piece, no need to do all at once.

dereuromark (approved):
Good first step

MasterOdin:
Given that PHPUnit seemingly always deprecates/removes something for every major version, and that this is a dev dependency, I have little interest in trying to support multiple versions of phpunit concurrently.

tests/Phinx/Config/AbstractConfigTestCase.php
```diff
@@ -5,13 +5,13 @@
...
 /**
- * Class AbstractConfigTest
+ * Class AbstractConfigTestCase
  *
  * @package Test\Phinx\Config
  * @group config
  * @coversNothing
  */
-abstract class AbstractConfigTest extends TestCase
+abstract class AbstractConfigTestCase extends TestCase
```
MasterOdin, line 14: Renamed this as it wasn't a file that had tests, but rather something the concrete `Config*Test` files extended to make use of the same helper functions. phpunit 10 starting having a warning about this, so renaming to `AbstractConfigTestCase` seemed most in keeping with the naming scheme that phpunit uses, and avoids the warning.

.github/workflows/ci.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         if [[ ${{ matrix.db-type }} == 'pgsql' ]]; then export PGSQL_DSN='pgsql://postgres:postgres@127.0.0.1/phinx'; fi
 
         if [[ ${{ matrix.prefer-lowest != 'prefer-lowest' }} ]]; then
-          export CODECOVERAGE=1 && vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+          export CODECOVERAGE=1 && vendor/bin/phpunit --coverage-clover=coverage.xml
```
MasterOdin, line 105: I think the biggest change would be no longer printing out the skipped tests, but I think that's actually preferable as they were essentially just noise for the most part and something to annoyingly scroll past to see actual test errors.

If we wanted to bring back the behavior, can use the new `--display-skipped` flag or `displayDetailsOnSkippedTests` XML option.

Issue #2323, opened 2024-11-26
Add Native Support for ALGORITHM=INSTANT in Migrations for MYSQL


abhikaddy:
Currently, phinx does not natively support specifying ALGORITHM=INSTANT for MySQL migrations. ALGORITHM=INSTANT allows for schema changes (like adding columns) without requiring a table copy or affecting existing data, offering significant performance and downtime advantages for compatible operations.

This feature would enhance the flexibility and efficiency of database schema migrations, especially for users leveraging MySQL 8.0+ where ALGORITHM=INSTANT is available for certain DDL operations.

Proposed Solution:

Introduce a method or syntax in migrations to allow developers to specify the use of ALGORITHM=INSTANT explicitly for operations that support it.
Example:
$table->addColumn('new_column', 'string', ['algorithm' => 'instant'])->update();
Ensure fallback behavior is clearly defined if ALGORITHM=INSTANT is unsupported by the MySQL version or for the requested operation.
Benefits

Reduces downtime and performance impact during migrations for supported schema changes.
Aligns phinx with modern database practices and MySQL optimizations.
MySQL documentation on [ALGORITHM=INSTANT](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl.html).

Adding this feature will provide a smoother migration experience, particularly for large-scale production environments. It would also make phinx more competitive by leveraging cutting-edge database features.

Please let me know if further clarification or input is required!

dereuromark:
This has been added in https://github.com/cakephp/migrations/pull/955
Feel free to port to phinx as well directly.

MasterOdin:
Added in #2399

PR #2399, opened 2025-12-03
Add algorithm/lock Support for alter columns in mysql

lutdev:
Fix for the issue - https://github.com/cakephp/phinx/issues/2323
Original fix - https://github.com/cakephp/migrations/pull/955/files

Also changed version of PHP for Docker because project requires min 8.1 but there is 7.3 in the container

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,13 +77,16 @@ install [docker-compose](https://docs.docker.com/compose/) for your platform.
     docker-compose run --rm phinx
     ```
 
-1. Install dependencies:
+    If you use Mac with Apple Silicon add `platform: linux/amd64` for `mysql` and `postgres` services first. Otherwise,
+    you might have an error `no matching manifest for linux/arm64/v8 in the manifest list entries`
+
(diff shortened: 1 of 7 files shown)


lutdev:
- [x] Fix tests

MasterOdin:
I'm not super familiar with this feature of mysql, but could you set the algorithm and lock for other operations when altering a table, like adding or removing an index?

lutdev:
> I'm not super familiar with this feature of mysql, but could you set the algorithm and lock for other operations when altering a table, like adding or removing an index?

for most of operations, only "In Place" algorithm takes place) https://dev.mysql.com/doc/refman/8.4/en/innodb-online-ddl-operations.html
So, I think it's not possible to set an algorithm for such kind of operations. For example, I need it only when need to modify column. And without migrations it's a bit painful :D

MasterOdin:
While I don't necessarily love the approach of setting options on `alterColumn` affecting subsequent alter column calls, I don't really have a good way to handle this in the current API, so :shrug:

Thanks for the PR.

PR #2401, opened 2025-12-08
Use timestamp column type for phinxlog if addTimestampsUseDateTime flag disabled

qnnp-me:
- Change start_time column type to datetime
- Change end_time column type to datetime

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -562,7 +562,8 @@ For some breaking changes, Phinx offers a way to opt-out of new behavior. The fo
 * ``column_null_default``: Should Phinx create columns as null by default? (default: ``true``)
 
 Since MySQL ``TIMESTAMP`` fields do not support dates past 2038-01-19, you have the option to use ``DATETIME`` field
-types for fields created by the ``addTimestamps()`` function:
+types for fields created by the ``addTimestamps()`` function. This setting also affects the start_time and end_time 
+columns in the schema table:
 
 * ``add_timestamps_use_datetime``: Should Phinx create created_at and updated_at fields as datetime? (default: ``false``)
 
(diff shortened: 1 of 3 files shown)


lutdev:
Just for my curious, what problem should these changes fix?)

qnnp-me:
> Just for my curious, what problem should these changes fix?)

for "2038-01-19 03:14:07 UTC"

dereuromark:
In general datetime nowadays is sure a better default for normal use.
I wonder though if this is BC in this case.

qnnp-me:
> In general datetime nowadays is sure a better default for normal use. I wonder though if this is BC in this case.

Overall, using DATETIME as the default is indeed a better choice for general use today. Regarding your concern about backward compatibility (BC), this has been addressed in the code. By introducing the `FeatureFlags::$addTimestampsUseDateTime` flag, the system can now dynamically choose between DATETIME and TIMESTAMP column types, ensuring that existing behaviors remain unaffected while allowing the improved default for new implementations.

lutdev:
@qnnp-me  what if it would be possible to change column type automatically?) if feature flag is provided, of course. To update the existing table in the database.

qnnp-me:
@lutdev I agree with your approach. For new projects, setting the preferred column type early via the feature flag is ideal. For existing projects, automatically changing column types based on the flag could be risky and unexpected. Instead, developers can manually create and run a migration script when needed to update their tables. This keeps control with the developer and avoids unintended changes to production databases.

MasterOdin (changes requested):
Can you update the docs for the feature flag to indicate that it also modifies the behavior of creating the `phinxlog` table in https://github.com/cakephp/phinx/blob/0.x/docs/en/configuration.rst#feature-flags as well as its docstring (https://github.com/cakephp/phinx/blob/cea9c12633c90cbb16db97ee50c58be2130cdba7/src/Phinx/Config/FeatureFlags.php#L26).

MasterOdin:
When I put out the next version with this PR, I will include in the notes on updating the phinxlog column types manually for those on older versions. I agree that I don't think it's worth trying to automatically migrate folks.

qnnp-me:
@MasterOdin Done. Ready for final review.

MasterOdin (approved):

Issue #2386, opened 2025-11-04
SqlServerAdapter::getColumns returns associative array


MasterOdin:
The [`SqlServerAdapter::getColumns`](https://github.com/cakephp/phinx/blob/8270c6ec96bb1262c43c5f1ea322b6b4218cac28/src/Phinx/Db/Adapter/SqlServerAdapter.php#L450-L489) function returns an associative array of columns, where the keys are column names. All other adapters return a numeric array which is also expressed by our `AdapterInterface`:

https://github.com/cakephp/phinx/blob/8270c6ec96bb1262c43c5f1ea322b6b4218cac28/src/Phinx/Db/Adapter/AdapterInterface.php#L426-L432

The `SqlServerAdapter::getColumns` method should be changed to return numeric array. As this is public facing functionality, this will require a major release.

dereuromark:
Thats a valid point.
I took a stab at this in https://github.com/cakephp/migrations/pull/953

Maybe the next phinx release can do sth similar.

lutdev:
Prepared PR -> https://github.com/cakephp/phinx/pull/2402

PR #2402, opened 2025-12-10
[2386] SqlServerAdapter::getColumns returns non-associative array

lutdev:
Fix for https://github.com/cakephp/phinx/issues/2386

### DIFF
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -428,7 +428,7 @@ interface AdapterInterface
      * Returns table columns
      *
      * @param string $tableName Table name
-     * @return list<\Phinx\Db\Table\Column>
+     * @return \Phinx\Db\Table\Column[]
      */
     public function getColumns(string $tableName): array;
 
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -428,7 +428,7 @@ interface AdapterInterface
      * Returns table columns
      *
      * @param string $tableName Table name
-     * @return \Phinx\Db\Table\Column[]
+     * @return list<\Phinx\Db\Table\Column>
      */
     public function getColumns(string $tableName): array;
 
--- a/src/Phinx/Db/Adapter/SqlServerAdapter.php
+++ b/src/Phinx/Db/Adapter/SqlServerAdapter.php
@@ -487,7 +487,7 @@ class SqlServerAdapter extends PdoAdapter
                 $column->setPrecision((int)$columnInfo['precision']);
             }
 
-            $columns[$columnInfo['name']] = $column;
+            $columns[] = $column;
         }
 
         return $columns;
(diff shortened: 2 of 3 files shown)


src/Phinx/Db/Adapter/AdapterInterface.php
```diff
@@ -428,7 +428,7 @@ public function truncateTable(string $tableName): void;
      * Returns table columns
      *
      * @param string $tableName Table name
-     * @return \Phinx\Db\Table\Column[]
+     * @return list<\Phinx\Db\Table\Column>
```
dereuromark: The former also means non assoc. But it doesnt have the downsides of list<>.
Usually list is too strict as in it has to be the perfect numerical order, which is often kind of pointless.
As return type it is probably fine.

dereuromark:
This probably has to wait until next major.
A bit too bc breaking for a bugfix

src/Phinx/Db/Adapter/AdapterInterface.php
lutdev: @dereuromark 
> Usually list is too strict as in it has to be the perfect numerical order

yes) that's why I decided to change it. To be sure that method returns strict list.

dereuromark: But why is that important? If it is only about values? And we only iterate over it - as per definition of non assoc?

lutdev: @dereuromark 
> And we only iterate over it

when you only iterate through it and don't care about keys, there is no difference, for sure. But if you want to use keys, then it's important. For example,
```
foreach ($myArray as $key => $arrayValue) {
    if ($key === 2) {
        continue;
    }
}
```

It's not important in our case but in general `list` explicitly shows that keys in array start from 0 and array has a strict order. Just `type[]` doesn't guarantee it.

_Of course, "guarantee" is not about hard check, just on PHPDoc level_

dereuromark: Well, but exactly here is the problem:
You are saying its a non-assoc list (which by definition means you do not care about keys) - and then you care about keys.
Its illogical by design. We moved away from this anti pattern in CakePHP for this very reason. Its completely bongers. It's basically saying: "It's all about the keys now".

If you care about keys, use assoc, if not, ideally list<> would have been used to say "its only about the values", iterate over it, but dont access keys. 
Its now kind of abused to define this "super strict assoc list", or at least doesnt leave another interpretation for true value-only.

I understand what you mean, but in practical terms, its not really useful with current specs.

We use `array<value>` to signal that it is only about the values, and otherwise include the key.

MasterOdin: I agree with @dereuromark, where we should be using `Type[]` to designate a list of elements, where they keys are irrelevant, it's something to `foreach` through. This mirrors the types you'll see in other systems (e.g. typescript).

dereuromark: Using generic is cool though: `array<Type>` is a bit more modern.

lutdev: @dereuromark @MasterOdin thanx for discussion and opinions) I like this. Changed it back.

dereuromark:
@MasterOdin Are you preparing a 1.0.0 stable release this year? Then you could open the 1.x branch for it and merge this in.

PR #2405, opened 2026-02-27
Fix php85 mysql attr constant

schlndh:
[PHP 8.4](https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses) introduced driver-specific PDO subclases (e.g. `\PDO\Mysql`). [PHP 8.5](https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.pdo) deprecated `\PDO::MYSQL_*` (etc) constants.

I'm using phinx with `mysql_attr_multi_statements: 0` and on PHP 8.5 I get

```
PHP Deprecated:  Constant PDO::MYSQL_ATTR_MULTI_STATEMENTS is deprecated since 8.5, use Pdo\Mysql::ATTR_MULTI_STATEMENTS instead in vendor/robmorgan/phinx/src/Phinx/Db/Adapter/MysqlAdapter.php on line 154
```
This PR should fix it.

I checked the other drivers as well, and the only other one with similar functionality is the SqlServerAdapter. However, that one doesn't seem to have the a PDO subclass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['8.1', '8.4']
+        php-version: ['8.1', '8.5']
         db-type: [sqlite, mysql, pgsql]
         prefer-lowest: ['']
         include:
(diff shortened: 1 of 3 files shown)


remluben:
+1 😄 - I recently ran into these as well. Having this merged would be nice. Many thanks for suggesting the changes.

dereuromark (approved):

PR #2412, opened 2026-05-14
Add native enum type support for postgresql

### LINKED ISSUE #2258: Support enums in postgres
Hi there, basically just checking to see if this something you all would consider adding. Postgres supports enums on columns but it's a two-step process where you make an enum type and then you set that type on the create like this:

```
CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');
CREATE TABLE person (
    name text,
    current_mood mood
);
```
More information at https://www.postgresql.org/docs/current/datatype-enum.html

ErfanMomeniii:
Adds native PostgreSQL enum type support to `PostgresAdapter`.                
                                                                                
## Problem                                                                    
                                                                                
PostgreSQL requires a two-step process for enum columns — first `CREATE TYPE ... AS ENUM (...)`, then referencing that type in column definitions. Phinx only supported MySQL's inline `ENUM('a','b')` syntax.      
                                                                                
See: https://github.com/cakephp/phinx/issues/2258                             
                                                                                
## Solution                                                                   
                                                                                
Handle the full enum lifecycle automatically with the naming convention `{table}_{column}` (e.g. `orders_status`) to prevent cross-table collisions.  
                                                                                
| Operation | Behavior |                                                      
|---|---|                                                                     
| `createTable()` | Emits `CREATE TYPE` before `CREATE TABLE` |               
| `addColumn()` | Emits `CREATE TYPE` before `ALTER TABLE` |                  
| `dropColumn()` | Drops the associated enum type after column removal |      
| `dropTable()` | Drops all associated enum types after table removal |       
| `getColumns()` | Round-trips enums back as `PHINX_TYPE_ENUM` with values from `pg_enum` |                                                              
                                                                                
## Usage                                                                      
                                                                                
```php                                                                        
$this->table('orders')
      ->addColumn('status', 'enum', [
          'values'  => ['pending', 'shipped', 'delivered'],
          'default' => 'pending',
      ])
      ->create();
```
Rollbacks work automatically — dropping a column or table cleans up the associated PostgreSQL type.

## Tests

10 new integration tests covering create, add, drop column, drop table, round-trip, cross-table isolation, and schema-qualified tables.

[Fixes #2258 ](https://github.com/cakephp/phinx/issues/2258)

### DIFF
--- a/src/Phinx/Db/Adapter/AdapterInterface.php
+++ b/src/Phinx/Db/Adapter/AdapterInterface.php
@@ -70,11 +70,13 @@ interface AdapterInterface
         self::PHINX_TYPE_POLYGON,
     ];
 
-    // only for mysql so far
+    // MySQL-specific types
     public const PHINX_TYPE_MEDIUM_INTEGER = 'mediuminteger';
-    public const PHINX_TYPE_ENUM = 'enum';
     public const PHINX_TYPE_SET = 'set';
     public const PHINX_TYPE_YEAR = 'year';
+
+    // Supported by MySQL and PostgreSQL
+    public const PHINX_TYPE_ENUM = 'enum';
 
     // only for postgresql so far
     public const PHINX_TYPE_CIDR = 'cidr';
--- a/src/Phinx/Db/Adapter/PostgresAdapter.php
+++ b/src/Phinx/Db/Adapter/PostgresAdapter.php
@@ -42,6 +42,7 @@ class PostgresAdapter extends PdoAdapter
         self::PHINX_TYPE_MACADDR,
         self::PHINX_TYPE_INTERVAL,
         self::PHINX_TYPE_BINARYUUID,
+        self::PHINX_TYPE_ENUM,
     ];
 
     private const GIN_INDEX_TYPE = 'gin';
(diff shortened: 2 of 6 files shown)


ErfanMomeniii:
Hi @dereuromark , could you please review it?

dereuromark:
@MasterOdin took over maintainance.

dereuromark:
@MasterOdin are u ok with this?

MasterOdin (changes requested):
I think there's a bug here where since using the table and column name to look up enum names, if a rename happens for either, then the link breaks on the enum. Probably should add a rename of the enum on renaming the table or column I guess? Alternative would be to do a DB look-up of the enum name when doing a drop column or drop table.

ErfanMomeniii:
Thanks, @MasterOdin. I fixed it.

`getRenameTableInstructions` and `getRenameColumnInstructions` now emit `ALTER TYPE ... RENAME TO`, so the `{table}_{column}` enum naming convention stays invariant after table or column renames. That way, `dropColumn` and `dropTable` can still locate the correct enum type.

The rename only applies to convention-based enum types, so custom enum types are left untouched.

I decided to skip the DB lookup approach for now to keep the change small, but we can always revisit it if supporting externally renamed enum types becomes necessary.

I also added tests covering table and column renames, as well as dropping columns/tables after a rename.

MasterOdin (approved):
thanks for your contribution.

PR #2418, opened 2026-06-29
Fix SQL injection in hasDatabase() by quoting database name in adapters

markstory:
Update query generation for reflection methods in adapter layers to mitigate potential sql injection. While phinx should never have user controlled data provided to it, security researchers don't know that.

This will stay as a draft until I have more time to go through the rest of the sprintf() queries in each adapter.


dereuromark (approved):

MasterOdin:
A better solution might be to add an optional parameters argument to `fetchRow` and `fetchAll` that get routed through to `query`. A breaking change on the adapter interface, but allows for stronger usage without having to deal with quoting ourselves.

